PR #5, opened 2015-10-05
Update EVG.md

benwilson-digicert:


guidolotito:
Ben, just a nit: instead of "the CA/Browser Forum EV policy identifier", shouldn't we call it the EV for SSL or TLS policy identifier, to distinguish it from the EV CS policy identifier?

PR #7, opened 2015-10-22
Ballot 151

benwilson-digicert:
Merging based on passage of Ballot 151 on Sept. 28 2015


benwilson-digicert:
Looking at this I don't see some of the  changes.  I'll start another branch for errata.

PR #17, opened 2016-04-29
Documentation and generation improvements.

jsha:
Add detail and links to README.
Set max-width on HTML output for better readability on wide screens.
Add a requirements.txt file for Python.

PR #19, opened 2016-04-30
Fix section 6.1.5

kroeckx:
Github's markdown requires 3 dashes, kramdown works with any

This also adds the header to each tables, as they are in the pdf file.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1018,7 +1018,7 @@ Certificates MUST meet the following requirements for algorithm type and key siz
 (1) Root CA Certificates
 
 ||Validity period beginning on or before 31 Dec 2010|Validity period beginning after 31 Dec 2010|
-|--|--|--|
+|---|---|---|
 |Digest algorithm|MD5 (NOT RECOMMENDED), SHA-1, SHA-256, SHA-384 or SHA-512|SHA-1\*, SHA-256, SHA-384 or SHA-512|
 |Minimum RSA modulus size (bits)|2048\*\*|2048|
 |ECC curve|NIST P-256, P-384, or P-521|NIST P-256, P-384, or P-521|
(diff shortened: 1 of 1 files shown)


sleevi:
I'm sorry, this change both doesn't work and contains material changes that would need to be balloted, such as updating the text and references.

pzb:
It looks like the build is broken for an unrelated reason.  Additionally the changes Kurt made match the current BRs.  Both https://cabforum.org/wp-content/uploads/BRv1.2.5.pdf and https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.3.4.pdf have the reference to FIPS 186-4.  The markdown version does not match the official version here.

sleevi:
Ah, yeah, I broke the .travis.yml build for PIP/WeasyPrint when merging #17. Separate from that, through, changing the footnote in an issue about fixing tables seems... Not right.

pzb:
I just changed this topic to fixing 6.1.5.  Once the build works again, we can compare the current official 6.1.5 to both the old Appendix A and to the HTML/PDF build from markdown and see if this PR has any net changes.

kroeckx:
After fixing one thing, I saw other things that didn't look right, so I think the combination of those patches should make it match with the official BR docs.

docs/BR.md
```diff
@@ -1018,31 +1018,35 @@ Certificates MUST meet the following requirements for algorithm type and key siz
...
+|Digest algorithm|SHA-1, SHA-256, SHA-384 or SHA-512|SHA-1*, SHA-256, SHA-384 or SHA-512|
+|Minimum RSA modulus size (bits)|1024|2048|
+|ECC curve|NIST P-256, P-384, or P-521|NIST P-256, P-384, or P-521|
+|Minimum DSA modulus and divisor size (bits)\*\*\*|L= 2048, N= 224 or  L= 2048, N= 256|L= 2048 N= 224 or L= 2048 N= 256
 
 (3) Subscriber Certificates
 
+||Validity period *ending* on or before 31 Dec 2013|Validity period *ending* after 31 Dec 2013|
+|---|---|---|
 |Digest algorithm|SHA-1*, SHA-256, SHA-384 or SHA-512|
```
pzb: This row is still wrong.  The last column is blank but should be `SHA1*,	SHA‐256,	SHA‐384	or	SHA‐
512	`

pzb:
From my review, the revised version matches the current official version of the BRs with the noted exception.  In my opinion, this should be good to merge if the missing cell is fixed.

docs/BR.md
kroeckx: That's fixed in the next patch?

Maybe I should squash them?

kroeckx: Oh, I see now ...

kroeckx:
I think everything should be fixed now.

pzb:
Looks good to me now.  Can someone else double check and merge?

PR #24, opened 2016-06-22
Ballot 164: Add serial number randomness requirement

jsha:
Statement of intent:

As demonstrated in
https://events.ccc.de/congress/2008/Fahrplan/attachments/1251_md5-collisions-1.0.pdf,
hash collisions can allow an attacker to forge a signature on the certificate
of their choosing. The birthday paradox means that, in the absence of random
bits, the security level of a hash function is half what it should be. Adding
random bits to issued certificates mitigates collision attacks and means
that an attacker must be capable of a much harder preimage attack. For a
long time the BRs have encouraged adding random bits to the serial number
of a certificate, and it is now common practice. This ballot makes that best
practice required, which will make the Web PKI much more robust against all
future weaknesses in hash functions. Additionally, it replaces "entropy" with
"CSPRNG" to make the requirement clearer and easier to audit, and clarifies
that serial number must be positive.

In Section 1.6.1 of the Baseline Requirements,

ADD

CSPRNG: A random number generator intended for use in cryptographic system.

In Section 7.1 of the Baseline Requirements,

REPLACE

"CAs SHOULD generate non-sequential Certificate serial numbers that exhibit at
least 20 bits of entropy."

WITH

"Effective September 30, 2016, CAs SHALL generate Certificate serial numbers
greater than zero (0) containing at least 64 bits of output from a CSPRNG."

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -214,7 +214,7 @@ No stipulation.
 
 **Cross Certificate**: A certificate that is used to establish a trust relationship between two Root CAs.
 
-**CSPRNG**: A random number generator intended for use in cryptographic system.
+**CSPRNG**: A random number generator intended for use in a cryptographic system.
 
 **Delegated Third Party**: A natural person or Legal Entity that is not the CA but is authorized by the CA to assist in the Certificate Management Process by performing or fulfilling one or more of the CA requirements found herein.
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -214,6 +214,8 @@ No stipulation.
 
 **Cross Certificate**: A certificate that is used to establish a trust relationship between two Root CAs.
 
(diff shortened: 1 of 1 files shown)


sleevi:
@jsha Appreciate you putting this together. However, the changes need to be balloted first; it's unclear if you're submitting this as a ballot on behalf of Let's Encrypt (and looking for two co-sponsors) or if this was meant as an individual submission - in which case, a member will need to submit it as a ballot as appropriate.

jsha:
This was intended to be accompanied by an email, which I appear to have sent from the wrong address. Re-sending from the correct address shortly.

jsha:
This ballot has passed. Ping on merging this change?

jsha:
Superseded by https://github.com/cabforum/documents/pull/31

PR #48, opened 2017-03-21
Improve CSS to make weasyprint version look more like current CAB Forum ver

gerv:
Plus a formatting fix here and there, and a blockquote padding fix.


pzb:
LGTM

PR #78, opened 2017-12-28
Removal of Domain Verification Method 1 and 5

JeremyRowley:
This removes method 1 and 5 from 3.2.2.4. The changes will be effective in March.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -532,19 +532,13 @@ This section defines the permitted processes and procedures for validating the A
 
 The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qualified Domain Name (FQDN) listed in the Certificate using at least one of the methods listed below.
 
-Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as Section 4.2.1 of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
+Completed validations of Applicant's ownership or control of the domain under this section may be valid for the issuance of multiple Certificates over time, except that any completed validation of Applicant's ownership or control using a modified or deleted subsections may not be reused for future issuance of Certificates as of the effective date of the deletion or modification. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as Section 4.2.1 of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate. 
 
 CAs SHALL maintain a record of which domain validation method, including relevant BR version number, they used to validate every domain.
 
 Note: FQDNs may be listed in Subscriber Certificates using dNSNames in the subjectAltName extension or in Subordinate CA Certificates via dNSNames in permittedSubtrees within the Name Constraints extension.
 
-##### 3.2.2.4.1 Validating the Applicant as a Domain Contact
-
-Confirming the Applicant's control over the FQDN by validating the Applicant is the Domain Contact directly with the Domain Name Registrar. This method may only be used if:
-1.	The CA authenticates the Applicant's identity under BR Section 3.2.2.1 and the authority of the Applicant Representative under BR Section 3.2.5, OR
-2.	The CA authenticates the Applicant's identity under EV Guidelines Section 11.2 and the agency of the Certificate Approver under EV Guidelines Section 11.8; OR
-3.	The CA is also the Domain Name Registrar, or an Affiliate of the Registrar, of the Base Domain Name.
-Note: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN.  This method is suitable for validating Wildcard Domain Names.
+##### 3.2.2.4.1 Reserved
 
 ##### 3.2.2.4.2 Email, Fax, SMS, or Postal Mail to Domain Contact
 
@@ -582,9 +576,7 @@ The Random Value SHALL remain valid for use in a confirming response for no more
 
 Note: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN.  This method is suitable for validating Wildcard Domain Names. 
 
-##### 3.2.2.4.5 Domain Authorization Document
-
-Confirming the Applicant's control over the FQDN by relying upon the attestation to the authority of the Applicant to request a Certificate contained in a Domain Authorization Document. The Domain Authorization Document MUST substantiate that the communication came from the Domain Contact. The CA MUST verify that the Domain Authorization Document was either (i) dated on or after the date of the domain validation request or (ii) that the WHOIS data has not materially changed since a previously provided Domain Authorization Document for the Domain Name Space.
+##### 3.2.2.4.5 Reserved
 
 ##### 3.2.2.4.6 Agreed-Upon Change to Website
 


docs/BR.md
```diff
@@ -532,19 +532,13 @@ This section defines the permitted processes and procedures for validating the A
...
 Note: FQDNs may be listed in Subscriber Certificates using dNSNames in the subjectAltName extension or in Subordinate CA Certificates via dNSNames in permittedSubtrees within the Name Constraints extension.
 
-##### 3.2.2.4.1 Validating the Applicant as a Domain Contact
-
-Confirming the Applicant's control over the FQDN by validating the Applicant is the Domain Contact directly with the Domain Name Registrar. This method may only be used if:
-1.	The CA authenticates the Applicant's identity under BR Section 3.2.2.1 and the authority of the Applicant Representative under BR Section 3.2.5, OR
-2.	The CA authenticates the Applicant's identity under EV Guidelines Section 11.2 and the agency of the Certificate Approver under EV Guidelines Section 11.8; OR
-3.	The CA is also the Domain Name Registrar, or an Affiliate of the Registrar, of the Base Domain Name.
-Note: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN.  This method is suitable for validating Wildcard Domain Names.
+##### 3.2.2.4.1 Reserved
```
sleevi: So the challenge I see is that the document goes into effect immediately upon adoption (thus deleting this section), but your intent is clearly to permit this until March.

An alternative way to approach this is keep the existing text, but then add an:
`For certificates issued on or after March 1, 2018, this method SHALL NOT be used for validation, and completed validations using this method SHALL NOT be used for the issuance of certificates`.

as a new paragraph to both this and 3.2.2.4.5

and then the changes above

docs/BR.md
```diff
@@ -532,19 +532,13 @@ This section defines the permitted processes and procedures for validating the A
...
 
-Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as Section 4.2.1 of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
+Completed validations of Applicant's ownership or control of the domain under this section may be valid for the issuance of multiple Certificates over time, except that any completed validation of Applicant's ownership or control using a modified or deleted subsections may not be reused for future issuance of Certificates as of the effective date of the deletion or modification. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as Section 4.2.1 of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
```
sleevi: First, read the comment on line 541 (to make sense of this)
Second, this is tricky, because arguably this statement is now restating what appears in 4.2.1 - that is, this is effectively descriptive of what 4.2.1 will then modify

So should we be touching 4.2.1 instead?

This is why I was unhappy about the language that was added to 190, because it doesn't provide for how it can be normatively specified in the text.

An alternative way, if we want to restate the individual requirements (to avoid ambiguity), would be to keep the existing text, and add:

For certificates issued on or after March 1, 2018, previously completed validations using the methods 3.2.2.4.1 and 3.2.2.4.5 SHALL NOT be valid for the issuance of new certificates.

docs/BR.md
JeremyRowley: I don't care about keeping it until March (and would be okay with an immediate deprecation date).  Considering there's two weeks of ballots plus an IPR review period, we'll be in March before it's effective anyway.  However, people may not vote for it without that language so I guess I better add it.

docs/BR.md
timfromdigicert: 4.2.1 has some interesting text:

"a CA may continue to reuse validation data or documents collected prior to the change, or the	validation itself, for	the period stated in this BR 4.2.1 ***unless otherwise specifically provided in a ballot***"

which I find a bit odd because I think all the appropriate dates and requirements should be in the BRs themselves (I don't want to have to search ballot histories for deadlines ...).

I think I agree we should fix 4.2.1 to state explicitly that you can't reuse information from validation methods that can't be used any more.

docs/BR.md
sleevi: Drat, that's a fair point with the explicit language - it can be problematic if it's retroactively enforced - e.g. if this didn't become adopted until March 8, what does it mean for those issued March 1 to March 8

My math shows this being (potentially) adopted mid-Feb, but if that's delayed, then I think your approach makes sense.

docs/BR.md
sleevi: Agreed, that's the problematic text I was referring to, and why I was trying to capture with the suggestion a way to textually capture in the BRs the limitation, and then that be the example of 'specifically provided in the ballot'. :)

docs/BR.md
timfromdigicert: The IPR review is a bit entertaining in this case.  I'm going to quick file a patent on not using bad validation methods ...

docs/BR.md
timfromdigicert: How about we be less clever and just fix the problematic text?

I think 4.2.1 should read:

"a CA may continue to reuse validation data or documents collected prior to the change, or the	validation itself, for	the period stated in this BR 4.2.1 unless otherwise specified in these requirements."

It's really weird for a requirements document to refer to ballots involved in its creation, especially as part of a normative requirement.

For a hilarious loophole, it isn't required that the ballot *PASSES*, only that a ballot exists ...

sleevi: Mostly because of the overly long rathole that this section proved to be and the way in which it was passed could potentially derail an otherwise valuable ballot.

I agree that this is all fundamentally problematic, and inherent to how 3.2 <-> 4.2.1 cross-reference eachother that leaves it ambiguous whether they're separate requirements, or if one is restating the other (and which one). The result is that it introduces two states when previously there were one - namely, you're either validating under Section 3.2 or you're validating under Section 4.2.1.

That's also why I'm not sure your proposed change resolves it - is "these requirements" specific to Section 4.2.1, or is it the overall Baseline Requirements? If the former, then we'd still need to touch 4.2.1 to forbid 3.2.2.4.1/.5. If the latter, then does that mean it can be read as extending the validity period as much as shortening it? Namely, does "the relevant requirement" refer to 3.2.2.4.1 (which limits it) or 4.2.1 (which states up to max validity)?

I'm increasingly inclined to think that the very broken language of 190 is best addressed in this by simply being explicit in 4.2.1 that these specific methods (3.2.2.4.1, 3.2.2.4.5) may not be reused. Jeremy's originally proposed language would implicitly nuke the 'any other method' that Kirk was arguing for with 190 - that is, it sets up a conflict between 4.2.1 and 3.2 that I don't think would be germane for this ballot. Changing the requirements in 3.2.2.4.1/3.2.2.4.5, as I originally proposed, would still potentially be ambiguous/loopholey. Fixing 4.2.1, as you propose (which I'm very much in support of, in spirit and ideal), will require more care. So an expedient route may just be the explicit provision in 4.2.1, so that we're clear on the specific requirements for 3.2.2.4.1/3.2.2.4.5, and then separately find a way to resolve 4.2.1 long term.

Does that sound reasonable?

timfromdigicert: So, because of the exact rathole problem you mentioned, one thing I was thinking of doing was first just be very explicit in 3.2.2.4.1/5 about what the rules exactly are.  As you noted, it's technically in the "ballot" so I think it operates adequately with the existing language.

And then fix the problematic language separately so it isn't tied up with the other issues, and doesn't give people an excuse to sidetrack the 1/5 removal.  So the short answer is yes, I think we're both in complete agreement.

timfromdigicert: How does this look:

https://github.com/cabforum/documents/compare/master...timfromdigicert:timfromdigicert-removebaddcv

sleevi: If you create a Pull Request, we can actually discuss there, but the short answer is I think 750 is still 'tricky' and potentially opening up the rathole, because of its use of `e.g.` rather than specific

`Validations completed using methods specified in Section 3.2.2.4.1 or Section 3.2.2.4.5 SHALL NOT be re-used on or after March 1, 2018.`

This resolves a few quirks - namely,
* on/after (in 3.2.2.4.1 & 3.2.2.4.5) vs after (4.2.1)
* e.g. indicates its an open set, this opening it up for interpretation (and ratholing), while the reword poses it as a closed set (the restriction applies to these two methods specifically), hopefully avoiding it

Anyways, if you're taking over for Jeremy, let's create a new PR and carry on there :)

JeremyRowley: Yeah - Tim will take it over. I'll still chime in with useless comments to distract everyone though.

PR #79, opened 2017-12-28
Remove bad DCV methods

timfromdigicert:


docs/BR.md
```diff
@@ -743,6 +747,8 @@ In no case may a prior validation be reused if any data or document used in the
 
 After the change to any validation method specified in the Baseline Requirements or EV Guidelines, a CA may continue to reuse validation data or documents collected prior to the change, or the validation itself, for the period stated in this BR 4.2.1 unless otherwise specifically provided in a ballot.
 
+Information from validations done using methods that are no longer allowed to be used (e.g. 3.2.2.4.1 and 3.2.2.4.5 on or after March 1, 2018) MUST NOT be re-used for validations.
```
sleevi: Copying over from https://github.com/cabforum/documents/pull/78#discussion_r158997450

I think this is still 'tricky' and potentially opening up the rathole, because of its use of e.g. rather than specific

`Validations completed using methods specified in Section 3.2.2.4.1 or Section 3.2.2.4.5 SHALL NOT be re-used on or after March 1, 2018.`

This resolves a few quirks - namely,
* on/after (in 3.2.2.4.1 & 3.2.2.4.5) vs after (4.2.1)
* `e.g.` indicates its an open set, this opening it up for interpretation (and ratholing), while the reword poses it as a closed set (the restriction applies to these two methods specifically), hopefully avoiding it

timfromdigicert:
My general feeling is that there is little that can be done in standards about willful misrepresentations and intentional misreadings of standards language.  It's tough enough dealing with all potential non-malicious interpretations ...

However I would be shocked if there isn't someone out there re-using "Any other method" data ... I haven't gone searching for such loopholes since neither my current nor previous employers were interested in that sort of chicanery, but it's probably worth looking into, and that might be another thing to address in another ballot to avoid encumbering this one.

dzacharo:
@timfromdigicert, shall we close this PR?

dzacharo:
I recommend closing this PR as obsolete. @timfromdigicert any objections?

sleevi:
@timfromdigicert objections to closing?

timfromdigicert:
Feel free to close.

timfromdigicert:
Or I can (and did).

PR #80, opened 2017-12-28
The Baseline Requirements should not reference ballots

timfromdigicert:


docs/BR.md
```diff
@@ -741,7 +741,7 @@ Section 6.3.2 limits the validity period of Subscriber Certificates. The CA MAY
 
 In no case may a prior validation be reused if any data or document used in the prior validation was obtained more than the maximum time permitted for reuse of the data or document prior to issuing the Certificate.
 
-After the change to any validation method specified in the Baseline Requirements or EV Guidelines, a CA may continue to reuse validation data or documents collected prior to the change, or the validation itself, for the period stated in this BR 4.2.1 unless otherwise specifically provided in a ballot.
+After the change to any validation method specified in the Baseline Requirements or EV Guidelines, a CA MAY continue to reuse validation data or documents collected prior to the change, or the validation itself, for the period stated in this BR 4.2.1, unless the description of the validation method in section 3.2.2.4 specifies otherwise.
```
sleevi: Quick thoughts:
* Section 4.2.1 references Section 3.2 in its entirety (Line 736), and not just 3.2.2.4, for purposes of validation - should the references be aligned?
* Is this meant that a section in 3.2.2.4 may extend the lifetime? Happy to leave that off unless/until it comes up explicitly, but an alternative would be 'the lesser of'
* I assume the intent is to align with the 'current' description of the validation method, and not allow it to be misread. For example, if v1.0 of a 3.2.2.4.1 said 90 days, and a v1.1 of 3.2.2.4.1 said MUST NOT be used, then the intent here is MUST NOT, but a CA (perhaps not a member of the Forum) could misinterpret that as valid within the 90 days at time of 1.0 issuance.

This is why I see it being challenging as long as we allow 'all past versions' to be in scope. An alternative way that could resolve this is:

* CAs MAY continue to reuse validation data or documents collected, or validations completed, under prior versions of the Baseline Requirements, for the period stated in this BR 4.2.1, provided that the data and documents and method of collection, or the prior validation, meet the requirements set forth in Section 3.2.

timfromdigicert: 1) On the first one, I used a narrow reference to avoid referencing anything I hadn't reviewed.  We can reference 3.2, but we'd have to make sure there are no unintended consequences first.

2) I originally actually had "the lesser of", but the language was complicated when trying to describe the lesser of two values, one of which may or may not exist.  So I went with the simpler language.

3) I'll have to think about the best way to avoid potential problems there.  I'm not sure the problem of people thinking past versions of the BRs are still valid should be handled here as opposed to more generally.  It's related to the point Gerv made in Taiwan about the fact that the BRs require compliance with the latest version of the BRs, and WebTrust requires compliance with another version of the BRs, and Mozilla policy says what it does, and the whole thing ends up being a complete mess.  But there doesn't seem to be the motivation to untangle all that.

dzacharo:
@timfromdigicert , shall we close this as well? You can convert it as an issue to be addressed in a future update if you still feel strongly about it.

dzacharo:
Shall we close this PR as well? @timfromdigicert @sleevi ?

sleevi:
@timfromdigicert Poking on this to see about closing out?

PR #10, opened 2016-01-28
Benwilson digicert patch 3

benwilson-digicert:
This is ready to merge into  the master branch.

PR #82, opened 2018-01-24
Clarify usage and requirements for random numbers

timfromdigicert:


docs/BR.md
```diff
@@ -267,6 +267,8 @@ No stipulation.
 
 **Expiry Date**: The "Not After" date in a Certificate that defines the end of a Certificate's validity period.
 
+**Freshness Value**: A Random Value that cannot be predicted in advance, and is intended to prevent the re-use of stale data and replay attacks.  It need not remain secret.
```
sleevi: I'm not sure this is meaningfully distinct from a Random Value - specifically, the "predicted in advance" aspect.

I don't think this would be a definitional thing, I think it's something that would be part of the explicit requirements themselves.

docs/BR.md
```diff
@@ -347,6 +349,8 @@ The binding SHALL use a digital signature algorithm or a cryptographic hash algo
 
 **Root Certificate**: The self-signed Certificate issued by the Root CA to identify itself and to facilitate verification of Certificates issued to its Subordinate CAs.
 
+**Secret Value**: A Random Value intended to be communicated to a third party, where the security of the associated process relies upon that value remaining secret.
```
sleevi: Similarly, I don't think this is an improvement. I think it hinders readability, and the goal of what you're trying to add ("where the security of ...") is an expression of intent more than it is a definitional issue.

docs/BR.md
timfromdigicert: More than happy to move requirements out of definitions.  The only real intent of this is to distinguish the two uses of random values.  It has caused confusion in the past and will continue to cause confusion in the future.

sleevi: I'm not sure I fully understand what you see as "two uses"

Your PR touches on two dimensions - secret/non-secret (although that seems to be an expression of intent rather than what the method guarantees?) and fresh/stale (where fresh is defined by the validity period of the random value within the validation method)

timfromdigicert: Well, if you look at the validation requirements, we only use two of the four combinations:

Fresh and must be Secret
Fresh and need not be Secret

where as you note, it's actually pseudo-freshness, which is why Wayne wanted to avoid using the word nonce.

sleevi: I do not believe that's the case.

We also have
Not fresh and need not be secret

3.2.2.4.10 is this

timfromdigicert: Yeah I noticed that and believe it's actually a bug in #10.

sleevi: I mean, it was by design from GlobalSign :)

It's not desirable for security, I think, but it's desirable to at least one CA.

timfromdigicert: GlobalSign is method #9.  Method #10 came from Let's Encrypt through Gerv and suffered from a bad case of inattention to detail and the consequences of playing telephone instead of talking directly.

sleevi: That's not correct. Please see https://groups.google.com/d/msg/mozilla.dev.security.policy/RHsIInIjJA0/SJ-SYQwHAAAJ

dzacharo:
@timfromdigicert @sleevi are you happy to close this PR? If you still see that the current language in the BRs is ambiguous, perhaps we should convert this into an issue and close the PR.

wthayer:
@timfromdigicert @CBonnell discussed this at the 11/2 validation sc meeting and agreed to close it.

PR #84, opened 2018-02-08
CP Review Working Group's GitHub Branch 

benwilsonusa:
All,
The Policy Review Working Group has just completed its review of the Baseline Requirements in an attempt to clarify use of the term “CA”, which sometimes can be ambiguous.  As the CA/B Forum has been briefed several times, the WG chose the term “TSP” to refer to the legal entity that operates the CA, recognizing that the TSP may operate multiple CAs as well as other businesses.  To maintain consistency, we decided to leave the term “CA” in as many as places as possible.  Please review this pull request and submit your comments.
Thanks,
CABF Policy Review Working Group


docs/BR.md
```diff
@@ -355,16 +356,18 @@ The binding SHALL use a digital signature algorithm or a cryptographic hash algo
...
 
-**Terms of Use**: Provisions regarding the safekeeping and acceptable uses of a Certificate issued in accordance with these Requirements when the Applicant/Subscriber is an Affiliate of the CA or is the CA.
+**Terms of Use**: Provisions regarding the safekeeping and acceptable uses of a Certificate issued in accordance with these Requirements when the Applicant/Subscriber is an Affiliate of the TSP or is the TSP.[Comment: Similar to the Subscriber Agreement, Terms of Use are entered with a TSP.]
 
 **Test Certificate**: A Certificate with a maximum validity period of 30 days and which: (i) includes a critical extension with the specified Test Certificate CABF OID  (2.23.140.2.1), or (ii) is issued under a CA where there are no certificate paths/chains to a root certificate subject to these Requirements.
 
+**Trust Service Provider**: An organization providing trust services, through a number of Certification Authorities, to its customers who may act as Subscribers or Relying Parties.  [Comment:  Do we want to  include Relying Parties?]
```
RufusJWB: I would make a reference here to http://www.etsi.org/deliver/etsi_en/319400_319499/31941101/01.02.00_20/en_31941101v010200a.pdf chapter 5.4.1 that very clearly defines the term TSP

docs/BR.md
```diff
@@ -233,7 +234,7 @@ No stipulation.
 
 **Certificate Revocation List**: A regularly updated time-stamped list of revoked Certificates that is created and digitally signed by the CA that issued the Certificates.
 
-**Certification Authority**: An organization that is responsible for the creation, issuance, revocation, and management of Certificates. The term applies equally to both Roots CAs and Subordinate CAs.
+**Certification Authority**: A certificate generation service (people, procedures, systems and technology) that is trusted by one or more entities to create, sign, revoke, and provide status information for public key certificates and is operated by a Trust Service Provider.[Comment: The term "CA" encompasses people, too.]
```
RufusJWB: ETSI EN 319 411-1defines a *CA* as:

> A CA is commonly understood to be a type of Trust Service Provider (TSP), as defined in the Regulation (EU)
> No 910/2014 [i.14], and also a form of certification service provider as defined in the Electronic Signatures Directive
> 1999/93/EC [i.1], which issues public key certificates. However, in the present document the term is used more to
> reference the technical component of the TSP concerned with certificate issuance.
> The authority trusted by the users of the certification services (i.e. subscribers as well as relying parties) to assign
> certificates is the TSP. The TSP has overall responsibility for the provision of the certification services identified in
> clause 4.4. The CA, as well as the TSP, can be identified in the certificate as the issuer and its private key is used to sign certificates.

RufusJWB:
I like this approach to differ between TSP and CA a lot, since it makes things clearer, as different things will have different names.

Maybe it would be reasonable to use this clean-up to get all the definitions of the BRGs in line with the definitions within ETSI TS 319 411-1 and ETSI TS 319 401.

docs/BR.md
timfromdigicert: So there was a long discussion about this.  The CA/B Forum definition and usage of TSP is not the ETSI usage of TSP.  They are similar, but different.  I actually argued that it will be confusing to have the same term used in two different but closely related documents, but I lost that fight.  It turns out ISO also uses TSP, with its own definition.

docs/BR.md
timfromdigicert: Yeah, this is another spot where the BR usage and ETSI usage diverge.

In the BRs, a CA is an organizational unit, not a TSP.  TSPs are legal entities that operate CAs.

docs/BR.md
dzacharo: Exactly. The TSP definition is also included in ISO 21188 which is referenced as a core document in WebTrust for CAs 2.0. So, this seems to be the common ground between the two audit standards (WebTrust and ETSI), at least for the interpretation of the Baseline Requirements.

RufusJWB: @timfromdigicert Do you have a reference where exactly ISO uses TSP in a different definition as ETSI? Maybe one could get in contact with ETSI and ask them to align their definition with the definition of ISO. The situation reminds me a little bit on this: https://xkcd.com/927/

timfromdigicert: Ok, so this is why using TSP was a very, very bad idea.  The purpose of this project was *NOT* and was *NEVER* intended to be about aligning the BRs with ISO or ETSI terminology.  That's a much bigger and longer project.

The purpose of this project was to separate out two existing concepts in the BRs, that both used the term CA: the organizational unit that performs CA operations and issues certificates, and the legal entity responsible for that organizational unit.

If it's going to continue to confuse people, we're going to have to go back to "CA operator" instead of TSP.

RufusJWB: I just finished the preparation of my annual ETSI audit and I would have been highly confused, if there would have been one three-letter abbreviation ('TSA') with a (slightly) different meaning in two documents (ETSI EN 319 411-1 / ETSI EN 319 401 and BRG), which are closely coupled.

dzacharo: @timfromdigicert, TSP *is* the legal entity responsible for the operations of CAs (which are organizational units that manage specific CA Keys, have specific practices, etc). I am very happy with this definition and I think it makes things clearer than the "CA Operator" (which contains the term "CA" :)

benwilsonusa:
Closing - this is stale

PR #95, opened 2018-05-18
Clarify CA definition

dzacharo:
This is the CA/Browser Forum Policy Review WG proposal to **clarify the use of the word "CA"** in the Baseline Requirements.


dzacharo:
We still need to remove some comments but it's ok for the review process.

dzacharo:
Obsolete

PR #99, opened 2018-08-06
Ballot 224

1bbbrrr800:


dzacharo:
Will add ballot 224 changes sequentially.

PR #130, opened 2019-04-08
1.6.4

dzacharo:
Incorporates ballots SC14, SC15 and SC7.


dzacharo:
> 
> 
> The definition from SC14 in section 1.6.1 is missing: DNS TXT Record Phone Contact: The phone number defined in section B.2.2.
> 
> Everything else looks good.

I should have mentioned that this definition part is already in the BRs of the master branch from Doug's previous commit which was not reverted properly. So the outcome is the same. Please review in https://github.com/cabforum/documents/blob/master/docs/BR.md

PR #137, opened 2019-07-04
Ballot sc17   alternative registration numbers for ev certificates

dzacharo:
SC17


timfromdigicert:
These changes look good to me too.

PR #139, opened 2019-08-20
LEI Ballot

timfromdigicert:


docs/EVG.md
```diff
@@ -180,6 +180,10 @@ Capitalized Terms are defined in the Baseline Requirements except where provided
 
 **Extended Validation Certificate:**  See EV Certificate.
 
+**Global Legal Entity Identifier Foundation:** Established by the Financial Stability Board in June 2014, the Global Legal Entity Identifier Foundation (GLEIF) is tasked to support the implementation and use of the Legal Entity Identifier (LEI). The foundation is backed and overseen by the LEI Regulatory Oversight Committee, representing public authorities from around the globe that have come together to jointly drive forward transparency within the global financial markets. GLEIF is a supra-national not-for-profit organization headquartered in Basel, Switzerland.
```
sleevi: None of this marketing copy is necessary. This provides no definitional value, especially for the purpose of the EVGs.

Suggestion: Delete Entirely.

docs/EVG.md
```diff
@@ -180,6 +180,10 @@ Capitalized Terms are defined in the Baseline Requirements except where provided
...
+
+**Global Legal Entity Identifier Index:** The Global Legal Entity Identifier (LEI) Index contains historical and current LEI records including related reference data in one authoritative, central repository. The reference data provides the information on a legal entity identifiable with an LEI.
```
sleevi: This is also marketing copy, and can/should be deleted. It does not disambiguate in any meaningful sense, which is normally what these definitions should be used for.

docs/EVG.md
```diff
@@ -200,6 +204,8 @@ Capitalized Terms are defined in the Baseline Requirements except where provided
 
 **Legal Entity**: A Private Organization, Government Entity, Business Entity, or Non-Commercial Entity.
 
+**Legal Entity Identifier**: The Legal Entity Identifier (LEI) is a 20-character, alpha-numeric code based on the ISO 17442 standard developed by the International Organization for Standardization (ISO). It connects to key reference information that enables clear and unique identification of legal entities participating in financial transactions. Each LEI contains information about an entity’s ownership structure and thus answers the questions of 'who is who’ and ‘who owns whom’.
```
sleevi: Similarly, marketing copy is included here.

This part is wholly unnecessary, because the only salient piece of definitional requirement, namely that it complies with ISO17442, is duplicated in the actual normative requirement.

Suggestions:
1) Delete this text entirely
2) Meaningfully define it. What are the salient bits? Possible answers include:
  * Complies with the format set forth in ISO 17442
  * Assigned by an entity approved entity
    - For example, if the intent is that it complies with the LEI Regulatory Oversight Committee's requirements and uses an LOU recognized by the LEI ROC, then let's try to find out how to say that.
    - Alternatively, it might mean assigned by an LOU accredited by GLEIF

docs/EVG.md
```diff
@@ -1864,3 +1872,20 @@ guidelines:
    jurisdiction as specified in Section 9.2.4.  The stated address of the organisation 
    combined with the organization name SHALL NOT be the only information used to 
    disambiguate the organisation.
+   
+**LEI**: The Legal Entity Identifier (LEI) as specified in the ISO 17442 and registered
+   in the Global LEI Index.
```
sleevi: So it's ambiguous here as to what "registered in the Global LEI Index" means. It implies that an accredited-by-GLEIF LOU can withhold data from the Global LEI System, and thus some additional requirement is required.

If the changes to LEI are made within the definitions (e.g. does it have to be an LEI defined by an LOU accredited by GLEIF/the LEI ROC), then I think this entire first sentence may be unnecessary.

Alternatively, if we want to avoid normative requirements within the definition, such that LEI is merely the number that conforms to ISO 17442, then this is where we'd introduce the requirement that the LEI be issued by an LOU recognized by the Global Legal Entity Identifier Foundation.

No definition for "who" the foundation is is necessary.

docs/EVG.md
```diff
@@ -1864,3 +1872,20 @@ guidelines:
...
    combined with the organization name SHALL NOT be the only information used to 
    disambiguate the organisation.
+   
+**LEI**: The Legal Entity Identifier (LEI) as specified in the ISO 17442 and registered
+   in the Global LEI Index.  
+   
+   1. This information SHALL be validated by matching the organization name and 
+   registration number found in the Global LEI Index against the Subject Organization 
+   Name Field (see 9.2.1) and Subject Registration Number Field (see 9.2.5) within the 
+   context of the subject’s jurisdiction as specified in Section 9.2.4.
```
sleevi: It's completely undefined here what "matching" means, and this is where I wanted to flesh out more comprehensive normative requirements.

> Before including an LEI within a certificate, the CA SHALL
> * Confirm that the ValidationSources field of the associated LEI record, as reported by GLEIF, contains FULLY_CORROBORATED
> * That the RegistrationAuthority of the associated LEI Record is the same entity as the Incorporating Agency or Registration Agency used in validating the information within the Subject Jurisdiction of Incorporation or Registration Field (see Section 9.2.4) and the Subject Registration Number Field (see Section 9.2.5)
> * That the Subject Organization Name Field (see Section 9.2.1) is contained, exactly as entered, within either the LegalName, OtherEntityNames, or TransliteratedOtherEntityNames of the associated LEI Record
> * That the Subject Registration Number (see Section 9.2.5) is equal, exactly as entered, to the RegistrationAuthorityEntityID of the associated LEI Record

That said, I'm concerned about whether or not we need to worry about the Registration Details - in particular, the registration status - e.g. should a lapsed or canceled LEI Registration be used? What about challenges to the data source? Should we worry about the LegalJursidiction code and matching it?

docs/EVG.md
```diff
@@ -1864,3 +1872,20 @@ guidelines:
...
+   
+   1. This information SHALL be validated by matching the organization name and 
+   registration number found in the Global LEI Index against the Subject Organization 
+   Name Field (see 9.2.1) and Subject Registration Number Field (see 9.2.5) within the 
+   context of the subject’s jurisdiction as specified in Section 9.2.4.  
+   
+   2. The address information SHALL be compared in order to detect potential 
+   matching errors or errors in the registration information.  If errors are found, 
+   they SHALL be reported to the GLEIF Foundation and/or relevant registration 
+   authority.
```
sleevi: I'm actually not a fan of this. I think, if anything, any matching errors should PREVENT the CA from issuing such a certificate.

dzacharo:
LEI is now in an ISO standard. See:
- https://www.gleif.org/en/about-lei/iso-17442-the-lei-code-structure
- https://www.iso.org/standard/78829.html

sleevi:
@dzacharo Could you clarify what you mean about “now”? ISO 17442 bas existed before the discussions even began, and that of course has factored into the discussions.

However, it also has zero bearing whatsoever on the discussions here, inasmuch as ISO 17442 is with respect to the numbering format and syntax. Even the recognition of GLEIF with respect to the LEI ROC predates our discussions, and similarly has no bearing. The existence of a naming authority is orthogonal to the core questions at play, which apply to validation, purpose, and the demonstrable negative security impact to relying parties by inclusion within certificates used for an inherently separate trust framework.

I wasn’t sure if there was something you believed had changed, or whether this was just catching up with understanding the details we’ve been discussing from the beginning.

dzacharo:
As I was going over the open pull requests, I just thought it was relevant, that's all :-)

PR #142, opened 2019-09-20
Markdown-friendly formatting and readability cleanups

castillar:
Wound up being kind of a large PR, but this contains all the changes to create output using Pandoc, as well as look-and-feel changes to make it Markdown-clean. If we need to split this up, we can; we can also just ballot this change since it's significant-sized. I *think* the only thing I changed in terms of actual words was the sample shell script provided in 3.2.2.4.6, but I may be remembering wrong.


dzacharo:
I think Ryan and Tim H were preparing a clean-up ballot that would take care of the shell script of 3.2.2.6. If your ballot is focused on fixing formatting issues, then I would recommend removing the change of the shell script so we don't mix the scope of the ballot.

castillar:
Good plan! Looking at it, my change was 95% whitespace-related, but I reverted it so that it's 99% whitespace and formatting. The 1% remaining is to go from this:
```
E.g. <shell script here>
```
To this:
```
E.g.:
<formatted shell script here>
```
Which should hopefully not ruffle any feathers. :)

docs/BR.md
```diff
@@ -760,11 +766,11 @@ Confirming the Applicant's control over the IP Address by performing the procedu
...
+Determination of what is "registry-controlled" versus the registerable portion of a Country Code Top-Level Domain Namespace is not standardized at the time of writing and is not a property of the DNS itself. Current best practice is to consult a "public suffix list" such as <http://publicsuffix.org/> (PSL), and to retrieve a fresh copy regularly. If using the PSL, a CA SHOULD consult the "ICANN DOMAINS" section only, not the "PRIVATE DOMAINS" section. The PSL is updated regularly to contain new gTLDs delegated by ICANN, which are listed in the "ICANN DOMAINS" section. A CA is not prohibited from issuing a Wildcard Certificate to the Registrant of an entire gTLD, provided that control of the entire namespace is demonstrated in an appropriate way.
```
sleevi:
```suggestion
Determination of what is "registry-controlled" versus the registerable portion of a Country Code Top-Level Domain Namespace is not standardized at the time of writing and is not a property of the DNS itself. Current best practice is to consult a "public suffix list" such as the [Public Suffix List (PSL)](<http://publicsuffix.org/>), and to retrieve a fresh copy regularly.

If using the PSL, a CA SHOULD consult the "ICANN DOMAINS" section only, not the "PRIVATE DOMAINS" section. The PSL is updated regularly to contain new gTLDs delegated by ICANN, which are listed in the "ICANN DOMAINS" section. A CA is not prohibited from issuing a Wildcard Certificate to the Registrant of an entire gTLD, provided that control of the entire namespace is demonstrated in an appropriate way.
```

docs/BR.md
```diff
@@ -875,9 +880,9 @@ If a Delegated Third Party fulfills any of the CA's obligations under this secti
...
 
-Within 30 days after ICANN has approved a new gTLD for operation, as evidenced by publication of a contract with the gTLD operator on [www.ICANN.org] each CA MUST (1) compare the new gTLD against the CA's records of valid certificates and (2) cease issuing Certificates containing a Domain Name that includes the new gTLD until after the CA has first verified the Subscriber's control over or exclusive right to use the Domain Name in accordance with Section 3.2.2.4.
+Within 30 days after ICANN has approved a new gTLD for operation, as evidenced by publication of a contract with the gTLD operator on <http://www.ICANN.org> each CA MUST (1) compare the new gTLD against the CA's records of valid certificates and (2) cease issuing Certificates containing a Domain Name that includes the new gTLD until after the CA has first verified the Subscriber's control over or exclusive right to use the Domain Name in accordance with Section 3.2.2.4.
```
sleevi:
```suggestion
Within 30 days after ICANN has approved a new gTLD for operation, as evidenced by publication of a contract with the gTLD operator on <http://www.ICANN.org> each CA MUST:
1. compare the new gTLD against the CA's records of valid certificates; and
2. cease issuing Certificates containing a Domain Name that includes the new gTLD until after the CA has first verified the Subscriber's control over or exclusive right to use the Domain Name in accordance with Section 3.2.2.4.
```

You made an equivalent change on line 867, so this keeps consistency. Although the Spring Cleanup will nuke this.

docs/BR.md
```diff
@@ -1039,19 +1044,19 @@ After reviewing the facts and circumstances, the CA SHALL work with the Subscrib
...
 
 For the status of Subscriber Certificates:
 
-If the CA publishes a CRL, then the CA SHALL update and reissue CRLs at least once every seven days, and the value of the nextUpdate field MUST NOT be more than ten days beyond the value of the thisUpdate field
+If the CA publishes a CRL, then the CA SHALL update and reissue CRLs at least once every seven days, and the value of the nextUpdate field MUST NOT be more than ten days beyond the value of the thisUpdate field.
 
 For the status of Subordinate CA Certificates:
 
-The CA SHALL update and reissue CRLs at least (i) once every twelve months and (ii) within 24 hours after revoking a Subordinate CA Certificate, and the value of the nextUpdate field MUST NOT be more than twelve months beyond the value of the thisUpdate field
+The CA SHALL update and reissue CRLs at least (i) once every twelve months and (ii) within 24 hours after revoking a Subordinate CA Certificate, and the value of the nextUpdate field MUST NOT be more than twelve months beyond the value of the thisUpdate field.
```
sleevi:
```suggestion
The CA SHALL update and reissue CRLs at least:
1. once every twelve months; and
2. within 24 hours after revoking a Subordinate CA Certificate.

 The value of the nextUpdate field MUST NOT be more than twelve months beyond the value of the thisUpdate field.
```

Similar to your changes on line 867.

docs/BR.md
```diff
@@ -1083,7 +1088,7 @@ If the OCSP responder receives a request for status of a certificate that has no
 Effective 1 August 2013, OCSP responders for CAs which are not Technically Constrained in line with Section 7.1.5 MUST NOT respond with a "good" status for such certificates.
 
 ### 4.9.11 Other forms of revocation advertisements available
-If the Subscriber Certificate is for a high-traffic FQDN, the CA MAY rely on stapling, in accordance with [RFC4366], to distribute its OCSP responses. In this case, the CA SHALL ensure that the Subscriber "staples" the OCSP response for the Certificate in its TLS handshake. The CA SHALL enforce this requirement on the Subscriber either contractually, through the Subscriber Agreement or Terms of Use, or by technical review measures implemented by the CA.
+If the Subscriber Certificate is for a high-traffic FQDN, the CA MAY rely on stapling, in accordance with [RFC4366](https://www.ietf.org/rfc/rfc4366.txt), to distribute its OCSP responses. In this case, the CA SHALL ensure that the Subscriber "staples" the OCSP response for the Certificate in its TLS handshake. The CA SHALL enforce this requirement on the Subscriber either contractually, through the Subscriber Agreement or Terms of Use, or by technical review measures implemented by the CA.
```
sleevi:
```suggestion
If the Subscriber Certificate is for a high-traffic FQDN, the CA MAY rely on stapling, in accordance with [RFC4366](https://tools.ietf.org/html/rfc4366), to distribute its OCSP responses. In this case, the CA SHALL ensure that the Subscriber "staples" the OCSP response for the Certificate in its TLS handshake. The CA SHALL enforce this requirement on the Subscriber either contractually, through the Subscriber Agreement or Terms of Use, or by technical review measures implemented by the CA.
```

Linking to the TXT version is cruel. Cruel I say!

The Browser Alignment will clean this up, but this also fits in with my remarks about reference links. It'd be useful to solve in the Markdown cleanup.

docs/BR.md
```diff
@@ -1493,124 +1498,124 @@ Certificates MUST be of type X.509 v3.
 This section specifies the additional requirements for Certificate content and extensions for Certificates generated after the Effective Date.
 
 #### 7.1.2.1 Root CA Certificate
-a. basicConstraints
-This extension MUST appear as a critical extension. The cA field MUST be set true. The pathLenConstraint field SHOULD NOT be present.
+a. `basicConstraints`
```
sleevi: Do we want to revisit how this section is titled in the Markdown ballot? Or are you intentionally deferring the renumbering implications?

That is:
```suggestion
##### 7.1.2.1.1 `basicConstraints`
```

docs/BR.md
```diff
@@ -1493,124 +1498,124 @@ Certificates MUST be of type X.509 v3.
...
-> Other values MAY be present.
->
-> If present, this extension SHOULD be marked non-critical.
->
-> \*\* Generally Extended Key Usage will only appear within end entity certificates (as highlighted in RFC 5280 (4.2.1.12)), however, Subordinate CAs MAY include the extension to further protect relying parties until the use of the extension is consistent between Application Software Suppliers whose software is used by a substantial portion of Relying Parties worldwide.
+f. `nameConstraints` (optional)
 
+   If present, this extension SHOULD be marked critical[^*].
+
+[^*]: Non-critical Name Constraints are an exception to RFC 5280 (4.2.1.10), however, they MAY be used until the Name Constraints extension is supported by Application Software Suppliers whose software is used by a substantial portion of Relying Parties worldwide.
```
sleevi: Unclear what syntax this is trying to use, but you don't seem to have used the same syntax in 6.1.5 (thankfully(?) the browser cleanup proposal will significantly rewrite and simplify that section?)

docs/BR.md
```diff
@@ -1493,124 +1498,124 @@ Certificates MUST be of type X.509 v3.
...
+
+[^**]: Generally Extended Key Usage will only appear within end entity certificates (as highlighted in RFC 5280 (4.2.1.12)), however, Subordinate CAs MAY include the extension to further protect relying parties until the use of the extension is consistent between Application Software Suppliers whose software is used by a substantial portion of Relying Parties worldwide.
 
 #### 7.1.2.3 Subscriber Certificate
-a. certificatePolicies
+a. `certificatePolicies`
+
+   This extension MUST be present and SHOULD NOT be marked critical.
+
+   *  `certificatePolicies:policyIdentifier` (Required)
```
sleevi: You used two spaces after the bullet here, but only one space after the bullet in 7.1.2.2

This threw me off as I was trying to wonder if my eyes were going wonky on me re: the indentation 😅

docs/BR.md
```diff
@@ -1493,124 +1498,124 @@ Certificates MUST be of type X.509 v3.
...
+   Either the value `id-kp-serverAuth` [RFC5280] or `id-kp-clientAuth` [RFC5280] or both values MUST be present. `id-kp-emailProtection` [RFC5280] MAY be present. Other values SHOULD NOT be present.
 
 #### 7.1.2.4 All Certificates
-All other fields and extensions MUST be set in accordance with RFC 5280. The CA SHALL NOT issue a Certificate that contains a keyUsage flag, extendedKeyUsage value, Certificate extension, or other data not specified in section 7.1.2.1, 7.1.2.2, or 7.1.2.3  unless the CA is aware of a reason for including the data in the Certificate.
+All other fields and extensions MUST be set in accordance with RFC 5280. The CA SHALL NOT issue a Certificate that contains a `keyUsage` flag, `extendedKeyUsage` value, Certificate extension, or other data not specified in section 7.1.2.1, 7.1.2.2, or 7.1.2.3  unless the CA is aware of a reason for including the data in the Certificate.
 
 CAs SHALL NOT issue a Certificate with:
 
 a. Extensions that do not apply in the context of the public Internet (such as an extendedKeyUsage value for a service that is only valid in the context of a privately managed network), unless:
```
sleevi: Similar remarks about renumbering and lists. I'm fine with not touching this yet.

docs/BR.md
```diff
@@ -1633,66 +1638,61 @@ By issuing the Certificate, the CA represents that it followed the procedure set
 Subject attributes MUST NOT contain only metadata such as '.', '-', and ' ' (i.e. space) characters, and/or any other indication that the value is absent, incomplete, or not applicable.
 
 #### 7.1.4.2.1 Subject Alternative Name Extension
-Certificate Field: extensions:subjectAltName
-Required/Optional: Required
-Contents: This extension MUST contain at least one entry. Each entry MUST be either a dNSName containing the Fully-Qualified Domain Name or an iPAddress containing the IP address of a server. The CA MUST confirm that the Applicant controls the Fully-Qualified Domain Name or IP address or has been granted the right to use it by the Domain Name Registrant or IP address assignee, as appropriate.
+
+__Certificate Field:__ `extensions:subjectAltName`
```
sleevi: Our other sections-of-emphasis use `**` rather than `__` - I'm curious if that was similarly intentional here?

docs/BR.md
```diff
@@ -1633,66 +1638,61 @@ By issuing the Certificate, the CA represents that it followed the procedure set
...
 
 Wildcard FQDNs are permitted.
 
 As of the Effective Date of these Requirements, prior to the issuance of a Certificate with a subjectAlternativeName extension or Subject commonName field containing a Reserved IP Address or Internal Name, the CA SHALL notify the Applicant that the use of such Certificates has been deprecated by the CA / Browser Forum and that the practice will be eliminated by October 2016. Also as of the Effective Date, the CA SHALL NOT issue a certificate with an Expiry Date later than 1 November 2015 with a subjectAlternativeName extension or Subject commonName field containing a Reserved IP Address or Internal Name. Effective 1 October 2016, CAs SHALL revoke all unexpired Certificates whose subjectAlternativeName extension or Subject commonName field contains a Reserved IP Address or Internal Name.
 
 Prior to April 1, 2019, certificates containing underscore characters (“_”) in domain labels in dNSName entries MAY be issued as follows:
```
sleevi: Should the underscore here be escaped or code-quoted? It certainly throws off the GitHub diff view, and wondering if it'll throw off others.

```suggestion
Prior to April 1, 2019, certificates containing underscore characters (“`_`”) in domain labels in dNSName entries MAY be issued as follows:
```

docs/BR.md
```diff
@@ -1752,21 +1754,21 @@ If the Certificate asserts the policy identifier of 2.23.140.1.2.1, then it MUST
...
+If the Certificate asserts the policy identifier of 2.23.140.1.2.2, then it MUST also include `organizationName`, `localityName` (to the extent such field is required under Section 7.1.4.2.2), `stateOrProvinceName` (to the extent such field is required under Section 7.1.4.2.2), and `countryName` in the Subject field.  If the Certificate asserts the policy identifier of 2.23.140.1.2.3, then it MUST also include (i) either `organizationName` or `givenName` and `surname`, (ii) `localityName` (to the extent such field is required under Section 7.1.4.2.2), (iii) `stateOrProvinceName` (to the extent required under Section 7.1.4.2.2), and (iv) `countryName` in the Subject field.
```
sleevi: Similar remarks about the cleanups from line 867

```suggestion
If the Certificate asserts the policy identifier of 2.23.140.1.2.2, then it MUST also include `organizationName`, `localityName` (to the extent such field is required under Section 7.1.4.2.2), `stateOrProvinceName` (to the extent such field is required under Section 7.1.4.2.2), and `countryName` in the Subject field. 

If the Certificate asserts the policy identifier of 2.23.140.1.2.3, then it MUST also include, in the Subject field:
  * either `organizationName` or both `givenName` and `surname`;
  * `localityName` (to the extent such field is required under Section 7.1.4.2.2);
  * `stateOrProvinceName` (to the extent required under Section 7.1.4.2.2); and
  * `countryName`.
```

Also totally understand if you want to defer that to a post-Markdown cleanup (and we should try to figure how to track these)

docs/BR.md
```diff
@@ -1,15 +1,12 @@
-**CA/Browser Forum**
-
+# Verso
```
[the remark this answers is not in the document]
castillar: Do you mean the "Verso" line itself? No special meaning to it—I just stuck it in there because Pandoc will now generate a title page for us automatically. But since it does that dynamically, not having any sort of title information on the document looked weird, so I converted it into a 'Verso' section that will be the fly-leaf info for the document.

assets/BR.yaml
```diff
@@ -0,0 +1,22 @@
+---
+title:  Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates
+subtitle:  Version 1.6.3
+version: "1.6.3"
```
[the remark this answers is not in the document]
castillar: Whoops. I'll make sure to clean those up while merging in the latest changes!

docs/BR.md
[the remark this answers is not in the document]
sleevi: Is "Verso" a special keyword for Pandoc? I guess I'm not familiar with it - could just be a lack of vocabulary, or if it was supposed to be a 'magic' phrase.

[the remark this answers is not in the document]
castillar: It's nothing significant for Pandoc—just me being compulsive. :) It looked weird not having a title page in the Markdown, but then the PDF looked weird with a title page followed by a reiteration of the title page info. My solution was to keep the "extra" title page but label it 'Verso' (which, after [some reading](https://lts.library.cornell.edu/lts/rt/ref/ptsbk), doesn't seem quite correct--should probably go with "Colophon" or just "Document Information"). However, in looking over the resulting PDF, it looks like it's not doing what I expected, so I eliminated it in 376d176.

docs/BR.md
```diff
@@ -45,18 +41,17 @@ This certificate policy (CP) contains the requirements for the issuance and mana
 
 The following Certificate Policy identifiers are reserved for use by CAs as an optional means of asserting compliance with this document (OID arc 2.23.140.1.2) as follows:
 
-{joint-iso-itu-t(2) international-organizations(23) ca-browser-forum(140) certificate-policies(1) baseline-requirements(2) domain-validated(1)} (2.23.140.1.2.1); and
+`{joint-iso-itu-t(2) international-organizations(23) ca-browser-forum(140) certificate-policies(1) baseline-requirements(2) domain-validated(1)} (2.23.140.1.2.1);` and
```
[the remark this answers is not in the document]
castillar: I like this, but oddly the version I have on my machine already has it. I did catch a similar join-iso-itu-t blob in 7.1.6.1, though, which I updated similarly in 30f5dd3.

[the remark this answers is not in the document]
castillar: Are the "and"s part of the ASN.1? It looks like you're sliding them inside the code-quote block.

docs/BR.md
sleevi: From the call yesterday, the decision was to not expand any of these lists to bullets or to renumber.

docs/BR.md
sleevi: Just capturing from the call - also something post-this.

docs/BR.md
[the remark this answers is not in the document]
sleevi: I'm not sure I follow? The backticks moved to exclude the (2.23.140.1.2.1) - that is, the code quote only code-quotes the ASN.1. The () isn't part of the ASN.1

docs/BR.md
```diff
@@ -45,18 +41,17 @@ This certificate policy (CP) contains the requirements for the issuance and mana
...
-
-{joint-iso-itu-t(2) international-organizations(23) ca-browser-forum(140) certificate-policies(1) baseline-requirements(2) individual-validated(3)} (2.23.140.1.2.3).
+`{joint-iso-itu-t(2) international-organizations(23) ca-browser-forum(140) certificate-policies(1) baseline-requirements(2) individual-validated(3)} (2.23.140.1.2.3)`.
 
 
 ### 1.2.1 Revisions
 
 | **Ver.** | **Ballot** | **Description** | **Adopted** | **Effective\*** |
-| :---: | :---: | --- | :---: | :---: |
+|-|-|-----|--|--|
```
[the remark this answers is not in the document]
castillar: It's a width specifier, with the pipes matching the columns. Per that Pandoc table link (extension "pipe_tables"):
> If a pipe table contains a row whose printable content is wider than the column width (see --columns), then the table will take up the full text width and the cell contents will wrap, with the relative cell widths determined by the number of dashes in the line separating the table header from the table body. (For example ---|- would make the first column 3/4 and the second column 1/4 of the full text width.)

The default is to even-width the columns, which squishes Description into an unreadable mess, so I'm specifying the relative column widths. Annoyingly, if you specify alignment using colons like this—`|:-:|:-:|-----|:--:|:--:|`—then it adds a bunch of column padding and things get very tight in the middle. One option might be to move the 'Description' column all the way to the right so it only needs padding on the left, but that's a non-trivial re-write of that table!

docs/BR.md
```diff
@@ -124,7 +119,7 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
 ### 1.2.2. Relevant Dates
 
 | **Compliance** | **Section(s)** | **Summary Description (See Full Text for Details)** |
-| --- | --- | --- |
+|--|--|----------|
```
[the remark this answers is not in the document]
castillar: As above.

docs/BR.md
castillar: Heh! Oops. Using one works fine, so I fixed it everywhere to use one space for consistency (bb0feb9).

docs/BR.md
castillar: Heh! Wellllllll, okay. :) Fixed in 965acc5.

docs/BR.md
castillar: The `[^*]` blob creates an auto-numbered footnote (https://pandoc.org/MANUAL.html#footnotes). Not actually sure why I used it there and didn't in 6.1.5—it's possible that since 6.1.5 contained multi-asterisk footnotes, I was worried about modifying the doc too much. I'll wait to see how the cleanup ballot modifies those sections and then we can revisit.

docs/BR.md
castillar: Hmmmm. Not intentional, no—not sure what my brain was doing at the time. Pandoc doesn't care, but it's a good idea to be consistent. For visibility with all the asterisks in the document, I might recommend converting everything to double-underscore, but I'm flexible on this one. Thoughts?

docs/BR.md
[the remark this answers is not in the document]
castillar: Aaaaaah, I missed the added back-tick. Got it! Fixed to match this in both sections in 6a362d4.

docs/BR.md
```diff
@@ -1655,59 +1651,61 @@ By issuing the Certificate, the CA represents that it followed the procedure set
...
+__Certificate Field:__ `extensions:subjectAltName`  
+__Required/Optional:__ Required  
+__Contents:__ This extension MUST contain at least one entry. Each entry MUST be either a dNSName containing the Fully-Qualified Domain Name or an iPAddress containing the IP address of a server. The CA MUST confirm that the Applicant controls the Fully-Qualified Domain Name or IP address or has been granted the right to use it by the Domain Name Registrant or IP address assignee, as appropriate.
 
 Wildcard FQDNs are permitted.
 
 CAs SHALL NOT issue certificates with a subjectAlternativeName extension or Subject commonName field containing a Reserved IP Address or Internal Name.
 
-Entries in the dNSName MUST be in the "preferred name syntax", as specified in RFC 5280, and thus MUST NOT contain underscore characters ("_").
+Prior to April 1, 2019, certificates containing underscore characters (“`_`”) in domain labels in dNSName entries MAY be issued as follows:
```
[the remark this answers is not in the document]
castillar: Something is definitely weird, because the version I have in the repo pushed from that update (https://github.com/castillar/documents/commit/f452eaa012fb6bc29530f4d91893f4b39a4b8af6#diff-7f6d14a20e7f3beb696b45e1bf8196f2) definitely includes all of the changes you mention here. I'm not sure what's going on, but perhaps it would be better for me to specifically reference that commit?

sleevi:
Replaced by #165

PR #147, opened 2019-11-13
Ballot SC21

dzacharo:


dzacharo:
@benwilson-digicert please review if the changes are according to SC21.

PR #165, opened 2020-03-11
(Draft) SC26: Pandoc Friendly Markdown

sleevi:


docs/BR.md
```diff
@@ -1699,59 +1698,61 @@ By issuing the Certificate, the CA represents that it followed the procedure set
...
-Entries in the dNSName MUST be in the "preferred name syntax", as specified in RFC 5280, and thus MUST NOT contain underscore characters ("_").
+Prior to April 1, 2019, certificates containing underscore characters (“`_`”) in domain labels in dNSName entries MAY be issued as follows:
+
+* dNSName entries MAY include underscore characters such that replacing all underscore characters with hyphen characters (“-“) would result in a valid domain label, and;
+* Underscore characters MUST NOT be placed in the left most domain label, and;
+* Such certificates MUST NOT be valid for longer than 30 days.
+
+All certificates containing an underscore character in any dNSName entry and having a validity period of more than 30 days MUST be revoked prior to January 15, 2019.
+
+After April 30, 2019, underscore characters (“_”) MUST NOT be present in dNSName entries.
```
sleevi: Compare L1710 in the original with L1710-1718 in new. This is a regression from the cleanup ballot.

docs/BR.md
```diff
@@ -1699,59 +1698,61 @@ By issuing the Certificate, the CA represents that it followed the procedure set
...
+
+i. __Certificate Field:__ `subject:organizationalUnitName` (OID: 2.5.4.11)
+   __Required/Optional:__ Optional.  
+   Contents: The CA SHALL implement a process that prevents an OU attribute from including a name, DBA, tradename, trademark, address, location, or other text that refers to a specific natural person or Legal Entity unless the CA has verified this information in accordance with Section 3.2 and the Certificate also contains `subject:organizationName`, `subject:givenName`, `subject:surname`, `subject:localityName`, and `subject:countryName` attributes, also verified in accordance with Section 3.2.2.1.
```
sleevi:
```suggestion
   __Contents__: The CA SHALL implement a process that prevents an OU attribute from including a name, DBA, tradename, trademark, address, location, or other text that refers to a specific natural person or Legal Entity unless the CA has verified this information in accordance with Section 3.2 and the Certificate also contains `subject:organizationName`, `subject:givenName`, `subject:surname`, `subject:localityName`, and `subject:countryName` attributes, also verified in accordance with Section 3.2.2.1.
```

docs/BR.md
```diff
@@ -1699,59 +1698,61 @@ By issuing the Certificate, the CA represents that it followed the procedure set
...
+
+c. __Certificate Field:__ `subject:givenName` (2.5.4.42) and `subject:surname` (2.5.4.4)  
+   __Required/Optional:__ Optional.  
+   __Contents:__ If present, the `subject:givenName` field and `subject:surname` field MUST contain an natural person Subject’s name as verified under Section 3.2.3. A Certificate containing a `subject:givenName` field or `subject:surname` field MUST contain the (2.23.140.1.2.3) Certificate Policy OID.
+
+d. __Certificate Field:__ Number and street: `subject:streetAddress` (OID: 2.5.4.9)  
+   __Required/Optional:__ Optional if the `subject:organizationName` field, `subject:givenName` field, or `subject:surname` field are present. Prohibited if the `subject:organizationName` field, `subject:givenName`, and `subject:surname` field are absent.
```
sleevi: This seems like a bit of a regression compared with the BRs (https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.8.pdf ), which emphasized **Required** / **Optional** via bold and also had them on separate lines.

docs/BR.md
```diff
@@ -2119,14 +2141,16 @@ These methods allow domain owners to publish contact information in DNS for the
 
 ### B.1.1. CAA contactemail Property
 
-SYNTAX: contactemail <rfc6532emailaddress> 
+SYNTAX: contactemail <rfc6532emailaddress>
```
sleevi: Do you need to escape this syntax with backticks? That is, does it try to make this a link?

docs/BR.md
```diff
@@ -2147,7 +2171,39 @@ The contactphone property MAY be critical if the domain owner does not want CAs
...
 
 ### B.2.2. DNS TXT Record Phone Contact
-The DNS TXT record MUST be placed on the "_validation-contactphone" subdomain of the domain being validated.  The entire RDATA value of this TXT record MUST be a valid Global Number as defined in RFC 3966 section 5.1.4, or it cannot be used.
+The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain of the domain being validated.  The entire RDATA value of this TXT record MUST be a valid Global Number as defined in RFC 3966 section 5.1.4, or it cannot be used.
+
+# APPENDIX C – Issuance of Certificates for .onion Domain Names
+This appendix defines permissible verification procedures for including one or more RFC 7686 ".onion" special-use Domain Names in a Certificate.
+
+1. The Domain Name MUST contain at least two labels, where the right-most label is "onion", and the label immediately preceding the right-most "onion" label is a valid Version 3 Onion Address, as defined in section 6 of the Tor Rendezvous Specification - Version 3 located at https://spec.torproject.org/rend-spec-v3.
```
sleevi:
```suggestion
1. The Domain Name MUST contain at least two labels, where the right-most label is "onion", and the label immediately preceding the right-most "onion" label is a valid Version 3 Onion Address, as defined in section 6 of the Tor Rendezvous Specification - Version 3 located at <https://spec.torproject.org/rend-spec-v3>.
```

docs/BR.md
```diff
@@ -2147,7 +2171,39 @@ The contactphone property MAY be critical if the domain owner does not want CAs
...
+2. The CA MUST verify the Applicant’s control over the .onion Domain Name using at least one of the methods listed below:
+    a. The CA MAY verify the Applicant’s control over the .onion service by using method 3.2.2.4.6, Agreed‐Upon Change to Website. If this method is replaced by a newer version(s) of Agreed-Upon Change to Website, the timelines for use of new and existing version of this method that are defined in section 3.2.2.4 SHALL apply.	
+    b. The CA MAY verify the Applicant's control over the .onion service by having the Applicant provide a Certificate Request signed using the .onion public key if the Attributes section of the certificationRequestInfo contains:
+        (i) A caSigningNonce attribute that contains a Random Value that is generated by the CA; and
+        (ii) An applicantSigningNonce attribute that contains a single value with at least 64-bits of entropy that is generated by the Applicant.
+	The signing nonce attributes have the following format:
+```
```
sleevi: Surely this doesn't render correctly, since you're trying to markdown in an escape sequence?

sleevi:
Jos: See the comments above

docs/BR.md
castillar: Curiously enough, it does! (And don't call me Shirley. :) ) Or at least, I think so--see attached screenshot for how this renders out in PDF form, and the Word doc looks much the same. I definitely had to play around with the spacing of the numbering sequence to get it to come out right.

br_sc26_sc27

sleevi: @castillar That's the /wrong/ output, although you're correct it's what @wthayer drafted in SC27, because that was copied over from EVG.md

You can see the 'expected' output in Appendix F of the current EV Guidelines, https://cabforum.org/wp-content/uploads/CA-Browser-Forum-EV-Guidelines-v1.7.1.pdf

When @dzacharo tried to fix the formatting in https://github.com/cabforum/documents/commit/ab17c2f9e3748cd54c7191fcec1fb8b8b206459d , he code quoted the sections, but this broke the markdown from when @benwilsonusa originally landed it in https://github.com/cabforum/documents/commit/ded42886c0569b181063439f804f9fa02378829d

That is, the correct approach for this section is either:
A) Use code-quoting (\`\`\`) for this section, and use whitespace to indicate correctness
B) Do not use code-quoting, and use tables to get the correct alignment

I think A is the correct answer here.

For a comparison with that is expected, beyond pointing to the EVGs that it's lifted from, [RFC 5912](https://tools.ietf.org/html/rfc5912) is relevant.

So, I know I suggested on the list we might be able to punt on SC27, but I think this matter is severe enough that we should fix it here, and have this replace the SC27 text.

sleevi:
@castillar I used GitHub to directly commit to your branch with the fix. Hopefully that works for you.

docs/BR.md
castillar: It doesn't try to make a link, but it does just ignore the \<\> contents, which isn't good. (See screenshot below) If we surround it with backticks, we'll get that whole line literally--is that the right output?
br_sc26_sc27_b111

sleevi:
```suggestion
SYNTAX: `contactemail <rfc6532emailaddress>`
```

docs/BR.md
castillar: Hunh. Good catch! Fixed in 266c208.

docs/BR.md
castillar: Good! Fixed, and changed the same block in B.1.2 to match (in f4f2597).

docs/BR.md
castillar: Whoof. Yeah, so this is one of those areas that strict Markdown doesn't handle well, partly because we're interleaving lists with non-list stuff, and partly because it's just a very specific layout. Thanks for the reference, though: that's exactly what I needed. I went with option A and then tried to adjust the spacing to make things come out correctly below the code-quote block. Does this look better?
sc26_sc27_appC

sleevi: Yup. That looks right!

docs/BR.md
castillar: I'm definitely wrestling with this one. You're absolutely right, I'm just trying to figure out how to fix it. This is a lettered list, so it's done manually rather than as a real Markdown list—as a result, trying to get indentations underneath items is fiddly, because Markdown doesn't like doing indent/outdent stuff (one place where something like reStructuredText or AsciiDoc might be more powerful). Here's one option that's probably the simplest to implement:
SC26-opt1
I'd want more indents on the Optional/Required thing, but I can't figure out a way to get them without blowing up the Contents line after them. Or we could re-org and put Contents before Required/Optional, so those are hanging off the bottom, _comme ça:_
opt2

sleevi: Oh! I see what you mean (and recording it here for others who may ask):

If this was a numbered list, Markdown would naturally support the aligned indent by aligning the text on the second line to the first text on the opening line. That is:

```
1. Foo
    Bar
    Baz
``` 
which results in:

1. Foo
    Bar
    Baz

However, because we're using a lettered list with a non-monospaced font, the whitespace will be variable (and not aligned). That's because Markdown sees this as "just" text.

Could you capture in a comment what your proposed syntax would be, both code-quoted and without? I'm curious how the GitHub UI is going to interpret things.

castillar: Sure! Here's option 1:
_snip!_
d. __Certificate Field:__ Number and street: `subject:streetAddress` (OID: 2.5.4.9)  
   __Required/Optional:__  
   __Optional__ if the `subject:organizationName` field, `subject:givenName` field, or `subject:surname` field are present.  
   __Prohibited__ if the `subject:organizationName` field, `subject:givenName`, and `subject:surname` field are absent.  
   __Contents:__ If present, the `subject:streetAddress` field MUST contain the Subject's street address information as verified under Section 3.2.2.1.
_/snip!_

```
d. __Certificate Field:__ Number and street: `subject:streetAddress` (OID: 2.5.4.9)  
   __Required/Optional:__  
   __Optional__ if the `subject:organizationName` field, `subject:givenName` field, or `subject:surname` field are present.  
   __Prohibited__ if the `subject:organizationName` field, `subject:givenName`, and `subject:surname` field are absent.  
   __Contents:__ If present, the `subject:streetAddress` field MUST contain the Subject's street address information as verified under Section 3.2.2.1.
```

Here's option 2:

_snip!_
d. __Certificate Field:__ Number and street: `subject:streetAddress` (OID: 2.5.4.9)  
   __Contents:__ If present, the `subject:streetAddress` field MUST contain the Subject's street address information as verified under Section 3.2.2.1.  
   __Required/Optional:__
   - __Optional__ if the `subject:organizationName` field, `subject:givenName` field, or `subject:surname` field are present.
   - __Prohibited__ if the `subject:organizationName` field, `subject:givenName`, and `subject:surname` field are absent.  
_/snip!_
```
d. __Certificate Field:__ Number and street: `subject:streetAddress` (OID: 2.5.4.9)  
   __Contents:__ If present, the `subject:streetAddress` field MUST contain the Subject's street address information as verified under Section 3.2.2.1.  
   __Required/Optional:__
   - __Optional__ if the `subject:organizationName` field, `subject:givenName` field, or `subject:surname` field are present.
   - __Prohibited__ if the `subject:organizationName` field, `subject:givenName`, and `subject:surname` field are absent.  
```

sleevi: Option 1 matches [our existing PDF](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.7.pdf#page=52)

Note that in the PDF/Word doc version, only one field has a "Required/Optional" and that's subject:commonName ... I'm not even sure what the 'required' is supposed to convey here.

Looking at the other fields in 7.1.4.2.2, they're aligned with your Option 1, with the note that they don't use "Required/Optional" but instead specify, each on a new line, Required, Optional, or Prohibited (e.g. 7.1.4.2.2(f) )

castillar: Great! Easy enough. a1a20ca fixes this whole section to Option 1, with an additional correction in 8b70d19, and a fix in 671531c to fix a missing backtick marker on item e. :)

assets/BR.yaml
```diff
@@ -0,0 +1,22 @@
+---
+title:  Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates
+subtitle:  Version 1.6.8
```
castillar, line 3: Looks like we'll need to fix this; I can submit a PR to correct it along with the Makefile for building.

wthayer: That made me realize that I needed to update the version to 1.7.0 and add the to the version table, which I have done.

BenWilson-Mozilla:
Is this the final that can be merged into the master?

dzacharo:
Ballot SC26

> 
> 
> Is this the final that can be merged into the master?

I think this is waiting for some additional Travis CI hooks and changes to pandoc. to be added/reviewed before they get merged, but I'm uncertain at this point. Asking @castillar and @sleevi.

sleevi:
> I think this is waiting for some additional Travis CI hooks and changes to pandoc. to be added/reviewed before they get merged, but I'm uncertain at this point. Asking @castillar and @sleevi.

Er, why does it need to wait on that? Do you use those Travis CI hooks for something? The discussion in the Forum Infrastructure WG was about making it easier for you in the future, but not about blocking anything now. However, if I'm misunderstanding how you currently do things, and that makes resolving that low-priority issue urgent and preventing the Forum from proceeding, it would be good to know.

dzacharo:
Since this particular ballot is not changing any of the requirements but just adds formatting to assist with the pandoc conversion, and since Jos has already provided the tools for that, I thought it was a matter of a couple of days to review and add those to this PR.

I'm fine to merge this and create another PR with the remaining files that don't need to go through a ballot.

Is that ok with everyone?

sleevi:
> Since this particular ballot is not changing any of the requirements but
> just adds formatting to assist with the pandoc conversion, and since Jos
> has already provided the tools for that, I thought it was a matter of a
> couple of days to review and add those to this PR.
>
Ah, no. It’s important to get this merged so that other ballots can deal
with sections that have been modified. That is, until you merge this,
everyone has to do more work to create a redline and can potentially create
redlines that revert the changes for this.

Automating the tooling is important and valuable, but if the tooling is not
an integral part of your workflow for producing docs yet, we shouldn’t wait
on it. That’s something we can fix up, but it’s going to be a little fiddly
to get right in the “doesn’t require a person sitting at their machine to
follow special commands” sense.

> I'm fine to merge this and create another PR with the remaining files that
> don't need to go through a ballot.
>
> Is that ok with everyone?
>
That is definitely preferable here. It would, for example, allow me to
update the cleanup and clarification ballot to create a proper redline and
resolve the conflicts. Which I would like to do (for this and for the other
draft ballots) :)

>

PR #167, opened 2020-03-13
(DRAFT) SMCWG Charter

sleevi:
Draft WIP charter

### DIFF
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -325,14 +325,5 @@ Issuance and Management of Publicly-Trusted S/MIME Certificates".
 
 ### 8. IPR Policy
 
-As with all Forum Working Group activity, the [IPR
-Policy](https://cabforum.org/ipr-policy/), v1.3 or later, SHALL apply to all
-activities and work of the SMCWG. All Participants in the SMCWG SHALL have on
-file with the Forum a valid, signed IPR Policy Agreement (v.1.3).  A previously
-submitted IPR Policy Agreement (v1.3) by an existing Member of the Forum shall
-suffice as meeting the obligation under section 4.5 of the IPR Policy that a
-Participant in the SMCWG commit to CAB Forum License requirements.
-
-In accordance with the Forum’s antitrust policy, an antitrust compliance
-statement SHALL be read at the start of all Working Group Meetings, in
-substantially the form written in Bylaw 1.3.
+The CA/Browser Forum Intellectual Rights Policy, v. 1.3 or later, SHALL apply
+to all Working Group activity.
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -326,4 +326,4 @@ Issuance and Management of Publicly-Trusted S/MIME Certificates".
 ### 8. IPR Policy
 
 The CA/Browser Forum Intellectual Rights Policy, v. 1.3 or later, SHALL apply
-to all Working Group activity.
+to all Working Group activity and Members.
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -127,8 +127,10 @@ The SMCWG SHALL consist of two classes of voting members, Certificate Issuers
 and Certificate Consumers meeting the eligibility criteria below.
 
 1. A Certificate Issuer eligible for voting membership in the SMCWG MUST have
-   a publicly-available audit report or attestation statement in accordance
-   with one of the following schemes:
+   a publicly-available audit report or attestation statement based upon a
+   publicly-available audit criteria or attestation scheme relevant to the
+   issuance of S/MIME certificates. This includes, but is not limited to,
+   the following schemes and criteria:
 
     * WebTrust for CAs v.2.0 or newer; or
     * ETSI EN 319 411-1, which includes normative references to ETSI EN 319 401
@@ -141,7 +143,9 @@ and Certificate Consumers meeting the eligibility criteria below.
     * No more than 27 months have elapsed since the beginning of the
       reported-on period and no more than 15 months since the end of the
       reported-on period; and
-    * The audit report was prepared by a Qualified Auditor.
+    * The audit report was prepared by a natural or legal person that is
+      duly authorized or licensed to conduct audits according to the relevant
+      audit criteria or attestation scheme used.
 
    In addition, the Certificate Issuer MUST actively issue S/MIME certificates
    that are treated as valid by at least one Certificate Consumer that produces
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -7,9 +7,9 @@ assist the interpretation of this Charter.
 
 An S/MIME certificate contains a public key bound to an email address; and may
 also contain the identity of a natural person or legal entity that controls such
-email address. The key pair can then be used to sign, encrypt, and decrypt
-email. An S/MIME certificate can be identified by the existence of an Extended
-Key Usage (EKU) Object Identifier (OID) of `1.3.6.1.5.5.7.3.4` for
+email address. The key pair can then be used to sign, verify, encrypt, and
+decrypt email. An S/MIME certificate can be identified by the existence of an
+Extended Key Usage (EKU) Object Identifier (OID) of `1.3.6.1.5.5.7.3.4` for
 emailProtection.
 
 The objective of an S/MIME certificate is to provide cryptographic security
@@ -77,12 +77,11 @@ capitalized terms in this Charter.
 The authorized scope of the SMCWG SHALL be to discuss, adopt, and maintain
 policies, frameworks, and sets of standards related to the issuance and
 management of S/MIME certificates by third-party CAs under a publicly trusted
-root, limited as follows:
+root.
 
-- Baseline verification of control over email addresses, including those used
-  by a natural person or a legal entity, or used by automated systems such as
-  for mailing lists
-- Identity validation for natural persons and legal entities
+The primary deliverable shall be scoped as follows:
+
+- Verification of control over email addresses
 - Key management and certificate lifecycle (subject to coordination with other
   Forum CWGs to ensure consistency and avoid redundancy)
 - Certificate profiles for S/MIME certificates and Issuing CA certificates
@@ -90,8 +89,13 @@ root, limited as follows:
   be present)
 - CA operational practices, physical/logical security, etc.
 
-The SMCWG SHALL exercise caution to ensure that its work product does not impede
-the issuance of certificates with other EKU types.
+In addition to the primary deliverable, the SMCWG MAY also address:
+- Identity validation for natural persons and legal entities, in the context of
+  S/MIME certificates
+
+Deliverables of the SMCWG SHALL be constrained to those certificates
+that contain the emailProtection (OID: 1.3.6.1.5.5.7.3.4) or that are technically
+capable of such issuance.
 
 ### 2. Out of Scope
 
@@ -162,8 +166,8 @@ forth in the Bylaws.
    * The organization name, as they wish it to appear on the Forum Web site and
      in official Forum documents.
    * URL of the applicant’s main Web site.
-   * Names and email addresses of employees who will participate in the Working
-     Group and Forum as Member representatives.
+   * Names and email addresses of designated representatives who will
+     participate in the Working Group and Forum on behalf of the Member.
    * Emergency contact information for security issues related to certificate
      trust.
 
@@ -196,13 +200,13 @@ forth in the Bylaws.
    Teleconference, or upon the request of any Member of the SMCWG, by a Ballot
    among Members of the SMCWG.
 
-   In order to determine the composition of the initial set of SMCWG Members,
-   at least twenty-four (24) hours prior to the initial meeting of the SMCWG,
-   the SMCWG Chair SHALL publish a list of Members seeking to participate who
-   they determine meet the criteria set forth above. As the first order of
-   business at the first meeting of the SMCWG, those organizations on the
-   Chair’s list of proposed, qualifying Members SHALL vote to determine the
-   initial set of SMCWG Members.
+To confirm the list of initial Members, at least two-thirds of the organizations
+on the Chair's list of proposed, qualifying Members SHALL publicly vote in favor
+of adopting the list of Members. If the initial list fails to be adopted, the
+Chair SHALL consider the feedback and MAY update the initial list of proposed,
+qualifying Members as a result, and a second vote using these same rules shall
+be performed. If an initial list of Members cannot be agreed upon, the SMCWG
+shall be dissolved.
 
 ##### 4.2.3. Suspension and Cessation of Working Group Membership
 
@@ -217,9 +221,7 @@ following become true:
 1. it stops providing support for its membership-qualifying software
    product;
 1. its membership-qualifying software product ceases to consume
-   S/MIME certificates;
-1. it announces an end of life for its membership-qualifying software product
-   and said end of life occurs.
+   S/MIME certificates.
 
 A Certificate Issuer’s membership in the SMCWG may be suspended if any of the
 following become true:
@@ -275,9 +277,9 @@ than 50% of the votes cast by the Certificate Consumers must be in favor of the
 ballot. At least one member of each class must vote in favor of a ballot for it
 to be adopted. Quorum is the average number of Member organizations (cumulative,
 regardless of Class) that have participated in the previous three (3) SMCWG
-Meetings or Teleconferences (not counting subcommittee meetings thereof). For
-transition purposes, if three (3) meetings have not yet occurred, quorum is
-three (3).
+Meetings or Teleconferences (not counting subcommittee meetings thereof). No
+Ballots shall be adopted until at least (3) meetings have occurred and quorum
+determined.
 
 #### 6.2. Other Organizational Matters
 
@@ -287,18 +289,18 @@ three (3).
   to the Chair herein MAY be performed by the Vice Chair. For example, the Vice
   Chair MAY preside at SMCWG Meetings and Teleconferences in the Chair’s
   absence.
-- SMCWG-created Subcommittees MAY be approved either:
-
-  1. by formal ballot as described in [6.1](#6.1-voting-structure), or
-  2. by simple majority vote of those members present at a regularly
-     scheduled SMCWG Meeting or Teleconference provided that the proposal is
-     mentioned in an agenda circulated on the SMCWG Mail List at least
-     forty-eight (48) hours prior to the SMCWG Meeting or Teleconference.
+- The SMCWG may establish subcommittees, by formal ballot, to address any of the
+  SMCWG’s business as specified in the ballot (each, a “SMCWG Subcommittee”).
+  SMCWG Subcommittees are open to all SMCWG Members. A SMCWG Subcommittee may
+  work on and recommend ballots, complete delegated functions, or issue reports
+  to the SMCWG that are within the scope of this charter. SMCWG Subcommittees
+  must post all agendas and minutes on a public mail list.
 
 ### 7. Summary of Major Deliverables
 
-The deliverables of the SMCWG are defined in the Scope section above and shall
-be named the "Baseline Requirements for S/MIME Certificates" proposed standard.
+The primary deliverable of the SMCWG is set forth in the Scope section, and
+shall include a Draft Guideline be named the "Baseline Requirements for S/MIME
+Certificates".
 
 ### 8. Primary Means of Communication
 
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -59,6 +59,15 @@ developing a consistent profile for S/MIME certificates in order to facilitate
 technical interoperability across the Internet are also to be addressed by
 the SMCWG.
 
+The objective of the SMCWG is to provide documents that describe an integrated
+set of technologies, protocols, identity-proofing, lifecycle management, and
+auditing requirements that are necessary (but not sufficient) for the issuance
+and management of Publicly-Trusted S/MIME Certificates; Certificates that are
+trusted by virtue of the fact that their corresponding Root Certificate is
+distributed in widely-available application software. Such requirements are not
+mandatory for Certification Authorities unless and until they become adopted and
+enforced by relying–party Application Software Suppliers. 
+
 ## Chartering of the S/MIME Certificate Working Group
 
 A Chartered Working Group (“SMCWG”) is created to perform the activities as
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -106,19 +106,14 @@ Deliverables of the SMCWG SHALL be constrained to those certificates
 that contain the emailProtection (OID: 1.3.6.1.5.5.7.3.4) or that are technically
 capable of such issuance.
 
-### 2. Out of Scope
-
-Certificates issued under a root certificate that is not publicly trusted SHALL
-be out of scope.
-
-### 3. Charter Expiration
+### 2. Charter Expiration
 
 The SMCWG is chartered indefinitely until it is dissolved as specified in Bylaw
 5.3.2(c).
 
-### 4. Personnel and Participation
+### 3. Personnel and Participation
 
-#### 4.1. Selection of Officers
+#### 3.1. Selection of Officers
 
 Stephen Davidson will act as chair of the SMCWG until the first Working Group
 Teleconference, at which time the group will elect a chair and vice-chair. The
@@ -128,9 +123,9 @@ held for chair and vice chair every two years in coordination with the Forum’s
 election process and in conjunction with its election cycle. Voting SHALL occur
 in accordance with Bylaw 4.1(c).
 
-#### 4.2. Participation
+#### 3.2. Participation
 
-##### 4.2.1. Eligibility to Participate
+##### 3.2.1. Eligibility to Participate
 
 The SMCWG SHALL consist of two classes of voting members, Certificate Issuers
 and Certificate Consumers meeting the eligibility criteria below.
@@ -168,7 +163,7 @@ and Certificate Consumers meeting the eligibility criteria below.
 The SMCWG SHALL allow participation by Interested Parties, as set
 forth in the Bylaws.
 
-##### 4.2.2. Membership Application/Declaration Process
+##### 3.2.2. Membership Application/Declaration Process
 
 1. An Applicant not already a member of the Forum SHALL provide the following
    information:
@@ -221,7 +216,7 @@ qualifying Members as a result, and a second vote using these same rules shall
 be performed. If an initial list of Members cannot be agreed upon, the SMCWG
 shall be dissolved.
 
-##### 4.2.3. Suspension and Cessation of Working Group Membership
+##### 3.2.3. Suspension and Cessation of Working Group Membership
 
 Members may resign from the SMCWG at any time. Resignation, suspension, or
 other termination of membership in the SMCWG does not prevent a Member from
@@ -269,7 +264,7 @@ endorse ballots or take part in any form of voting.
 
 Votes cast before the announcement of a Member’s suspension will stand.
 
-### 5. Membership Declaration
+### 4. Membership Declaration
 
 In accordance with the IPR Policy, Members that choose to participate in the
 SMCWG MUST declare their participation and SHALL do so prior to participating.
@@ -277,9 +272,9 @@ The Chair of the SMCWG SHALL establish a list for declarations of participation
 and manage it in accordance with the Bylaws, the IPR Policy, and the IPR
 Agreement.
 
-### 6. Voting and Other Organizational Matters
+### 5. Voting and Other Organizational Matters
 
-#### 6.1. Voting Structure
+#### 5.1. Voting Structure
 
 The rules described in Bylaw 2.3 and 2.4 SHALL apply to all ballots, including
 Draft Guideline Ballots.
@@ -294,7 +289,7 @@ Meetings or Teleconferences (not counting subcommittee meetings thereof). No
 Ballots shall be adopted until at least (3) meetings have occurred and quorum
 determined.
 
-#### 6.2. Other Organizational Matters
+#### 5.2. Other Organizational Matters
 
 - The Chair MAY delegate any of their duties to the Vice Chair as necessary.
   The Vice Chair has the authority of the Chair in the event of any absence or
@@ -309,13 +304,13 @@ determined.
   to the SMCWG that are within the scope of this charter. SMCWG Subcommittees
   must post all agendas and minutes on a public mail list.
 
-### 7. Summary of Major Deliverables
+### 6. Summary of Major Deliverables
 
 The primary deliverable of the SMCWG is set forth in the Scope section, and
 shall include a Draft Guideline be named the "Baseline Requirements for the
 Issuance and Management of Publicly-Trusted S/MIME Certificates".
 
-### 8. Primary Means of Communication
+### 7. Primary Means of Communication
 
 1. The SMCWG SHALL appoint a webmaster to maintain the SMCWG’s pages on the
    wiki and the Forum’s Public Web Site.
@@ -328,7 +323,7 @@ Issuance and Management of Publicly-Trusted S/MIME Certificates".
    Minutes SHALL be kept, and such minutes SHALL be made public in accordance
    with Bylaw 5.2.
 
-### 9. IPR Policy
+### 8. IPR Policy
 
 As with all Forum Working Group activity, the [IPR
 Policy](https://cabforum.org/ipr-policy/), v1.3 or later, SHALL apply to all
--- a/docs/SMCWG-charter.md
+++ b/docs/SMCWG-charter.md
@@ -62,16 +62,17 @@ the SMCWG.
 ## Chartering of the S/MIME Certificate Working Group
 
 A Chartered Working Group (“SMCWG”) is created to perform the activities as
-specified in this Charter, subject to the terms and conditions of the CA/Browser
-Forum Bylaws and Intellectual Property Rights (IPR) Policy, as such documents
-may change from time to time. This charter for the S/MIME Certificate Working
-Group has been created according to CAB Forum Bylaw 5.3.1. In the event of a
-conflict between this Charter and any provision in either the Bylaws or the IPR
-Policy, the provision in the Bylaws or IPR Policy SHALL take precedence. The
-definitions found in the Forum’s Bylaws SHALL apply to capitalized terms in this
-Charter.
+specified in this Charter, subject to the terms and conditions of the
+[CA/Browser Forum Bylaws](https://cabforum.org/bylaws/) and [Intellectual
+Property Rights (IPR) Policy](https://cabforum.org/ipr-policy/), as such
+documents may change from time to time. This charter for the S/MIME Certificate
+Working Group has been created according to CAB Forum Bylaw 5.3.1. In the event
+of a conflict between this Charter and any provision in either the Bylaws or
+the IPR Policy, the provision in the Bylaws or IPR Policy SHALL take
+precedence. The definitions found in the Forum’s Bylaws SHALL apply to
+capitalized terms in this Charter.
 
-### Scope
+### 1. Scope
 
 The authorized scope of the SMCWG SHALL be to discuss, adopt, and maintain
 policies, frameworks, and sets of standards related to the issuance and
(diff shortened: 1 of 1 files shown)


docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+
+The authorized scope of the SMCWG SHALL be to discuss, adopt, and maintain
+policies, frameworks, and sets of standards related to the issuance and
+management of S/MIME certificates by third-party CAs under a publicly trusted
+root, limited as follows:
+
+- Baseline verification of control over email addresses, including those used
+  by a natural person or a legal entity, or used by automated systems such as
+  for mailing lists
+- Identity validation for natural persons and legal entities
```
sleevi: As noted in https://cabforum.org/pipermail/public/2020-March/014917.html , one of the conclusions from the Cupertino and Thessaloniki F2F was that if identity is included in scope of the charter, that it's gated as a deliverable within the scope of the charter.

To that end, I think that we can remove this bullet, and then add something around line 94 that says:

---
Following the production of guidelines that meet these objectives, the SMCWG MAY further consider:
- Identity validation for natural persons and legal entities
---

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+- Key management and certificate lifecycle (subject to coordination with other
+  Forum CWGs to ensure consistency and avoid redundancy)
+- Certificate profiles for S/MIME certificates and Issuing CA certificates
+  (including the appropriateness of extensions and when those extensions should
+  be present)
+- CA operational practices, physical/logical security, etc.
+
+The SMCWG SHALL exercise caution to ensure that its work product does not impede
+the issuance of certificates with other EKU types.
```
sleevi: I'm worried about this statement, because a profile for intermediate certificates that precludes the use of `id-kp-serverAuth` may be interpreted as impeding the issuance of certificate with other EKU types.

---

Certificate profiles for the SMCWG SHALL be constrained to those certificates technically capable of issuing certificates with the emailProtection (OID: `1.3.6.1.5.5.7.3.4`) Extended Key Usage (EKU).

---

This still leaves 'technically capable' somewhat ambiguous, but at least then the answer for any impedance, if any, is separate roots, rather than suggesting that such requirements cannot be adopted at all.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+and Certificate Consumers meeting the eligibility criteria below.
+
+1. A Certificate Issuer eligible for voting membership in the SMCWG MUST have
+   a publicly-available audit report or attestation statement in accordance
+   with one of the following schemes:
+
+    * WebTrust for CAs v.2.0 or newer; or
+    * ETSI EN 319 411-1, which includes normative references to ETSI EN 319 401
+      (the latest version of the referenced ETSI documents should be applied).
+
```
sleevi: As noted in https://cabforum.org/pipermail/public/2020-March/014917.html , this precludes a number of existing deployed S/MIME frameworks, including inter-organization frameworks (such as the US Federal PKI or SAFE-BioPharma)

While direct integration with these other PKIs is not necessarily a goal, it's not necessarily worth precluding a priori either.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+1. An Applicant not already a member of the Forum SHALL provide the following
+   information:
+
+   * Confirmation that the applicant satisfies at least one (1) of the
+     membership eligibility criteria (and if it satisfies more than one (1),
+     indication of the single category under which the applicant wishes to apply).
+   * The organization name, as they wish it to appear on the Forum Web site and
+     in official Forum documents.
+   * URL of the applicant’s main Web site.
+   * Names and email addresses of employees who will participate in the Working
```
sleevi:
```suggestion
   * Names and email addresses of designated Representatives who will participate in the Working
```

This tries to tackle the employment issue, especially in situations where a policy management authority may contract to a third-party operator.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+     Group and Forum as Member representatives.
```
sleevi:
```suggestion
     Group and Forum.
```

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+
+In order for a ballot to be adopted by the SMCWG, two-thirds or more of the
+votes cast by the Certificate Issuers must be in favor of the ballot and more
+than 50% of the votes cast by the Certificate Consumers must be in favor of the
+ballot. At least one member of each class must vote in favor of a ballot for it
+to be adopted. Quorum is the average number of Member organizations (cumulative,
+regardless of Class) that have participated in the previous three (3) SMCWG
+Meetings or Teleconferences (not counting subcommittee meetings thereof). For
+transition purposes, if three (3) meetings have not yet occurred, quorum is
+three (3).
```
sleevi: Especially for the initial calls, this seems like a remarkably low number of participants, especially when it comes to adopting guidelines.

The simplest option is to remove this, and say that Ballots shall not be adopted until at least three meetings have been held. The moving of the the duration of the Chairship should make this reasonable.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+   Teleconference, or upon the request of any Member of the SMCWG, by a Ballot
+   among Members of the SMCWG.
+
+   In order to determine the composition of the initial set of SMCWG Members,
+   at least twenty-four (24) hours prior to the initial meeting of the SMCWG,
+   the SMCWG Chair SHALL publish a list of Members seeking to participate who
+   they determine meet the criteria set forth above. As the first order of
+   business at the first meeting of the SMCWG, those organizations on the
+   Chair’s list of proposed, qualifying Members SHALL vote to determine the
+   initial set of SMCWG Members.
```
sleevi: If the voting changes (discussed below) are adopted, this can be updated to say:

---

To confirm the list of initial Members, at least two-thirds of the organizations on the Chair's list of
proposed, qualifying Members SHALL publicly indicate vote in favor of adopting the list of Members. If the initial list fails to be adopted, the Chair SHALL consider the feedback and MAY update the initial list of proposed, qualifying Members as a result, and a second vote using these same rules shall be performed. If an initial list of Members cannot be agreed upon, the SMCWG shall be dissolved.

---

Basically, there's nothing in here that describes what happen if members don't vote in favor, and it suggests that a quorum for votes is only three votes. This attempts to make it so that 2/3 of those who express interest actually have to show up and vote. Similarly, by both public vote and through public Chair modification of the list, any shenanigans (e.g. such as removing all those with dissenting votes) are at least public.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+  to the Chair herein MAY be performed by the Vice Chair. For example, the Vice
+  Chair MAY preside at SMCWG Meetings and Teleconferences in the Chair’s
+  absence.
+- SMCWG-created Subcommittees MAY be approved either:
+
+  1. by formal ballot as described in [6.1](#6.1-voting-structure), or
+  2. by simple majority vote of those members present at a regularly
+     scheduled SMCWG Meeting or Teleconference provided that the proposal is
+     mentioned in an agenda circulated on the SMCWG Mail List at least
+     forty-eight (48) hours prior to the SMCWG Meeting or Teleconference.
```
sleevi: This does not seem to have any precedent in our Forum procedures, and we've intentionally adopted procedures that prohibit such "decide on a call" shenanigans, especially because they're hostile to members in various timezones and jurisdictions.

For example, membership votes may be forced to ballot, but this doesn't have such a provision.

While subcommittees are much more lightweight than they were, and can reasonably tackle anything in scope, they can also be used as a means to stall work (by deferring to an unproductive subcommittee chartered to stall work).

So while removal of this second bullet is optional, it would be a serious blow to the legitimacy of the Forum if this clause were ever exercised and work deferred to such a subcommittee rather than take it on. So it might be better to just remove.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+
+### 7. Summary of Major Deliverables
+
+The deliverables of the SMCWG are defined in the Scope section above and shall
+be named the "Baseline Requirements for S/MIME Certificates" proposed standard.
```
sleevi: This seems to put the cart before the horse.

I think one element of particular concern is that it presumes that any identity work MUST be tackled in the same document. I don't think that's necessarily a foregone conclusion. With the suggested approach to deferring identity until an actual deliverable is produced, this seems like this should just state that one or more Draft Guidelines shall be delivered to fulfill the objectives set out in the scope.

sleevi:
```suggestion
be named the "Baseline Requirements for S/MIME Certificates".
```

Carrying over a suggestion from @clintwilson regarding "proposed standard" not being something the CABForum produces.

docs/SMCWG-charter.md
sleevi: @wthayer Curious for any comments you might have as well

docs/SMCWG-charter.md
clintwilson: This is a good point, and one I had missed the context on. I would be reticent to support those frameworks once guidelines have been produced (transition periods accommodating, of course), but I agree there's not a need to preclude their participation in the CWG. With that understanding - assuming I have it right :| - in mind, I'd support adding that bullet back in.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+
+A Certificate Consumer Member’s membership may be suspended if any of the
+following become true:
+
+1. it stops providing support for its membership-qualifying software
+   product;
+1. its membership-qualifying software product ceases to consume
+   S/MIME certificates;
+1. it announces an end of life for its membership-qualifying software product
+   and said end of life occurs.
```
clintwilson: This third bullet point does feel duplicative of the first, at least in how it would likely be realized in practice where no shenanigans take place. I don't think it's necessary to keep either way and worried even while writing it about introducing the concept of "end of life".

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
...
+Members may resign from the SMCWG at any time. Resignation, suspension, or
+other termination of membership in the SMCWG does not prevent a Member from
+potentially having continuing obligations, under the Forum’s IPR Policy or any
+other document.
+
+A Certificate Consumer Member’s membership may be suspended if any of the
+following become true:
+
+1. it stops providing support for its membership-qualifying software
+   product;
```
clintwilson: I understand the change from "updates" to "support" to be potentially problematic. The primary concern with the "updates" is that it is entirely possible that valuable and important software in an ecosystem may not be under active and ongoing development, in the form of new feature releases and updates being made. Especially prior to the production of any draft guidelines by the CWG, I feel a requirement for semi-frequent updates by Certificate Consumers isn't appropriate.
I do agree that this change loses meeting some of the goals of Ballot 205. Perhaps it would be acceptable to directly incorporate a requirement for both Certificate Consumers and Certificate Issuers that they participate in CWG and/or Forum activities in some way, representing their continued investment and interest in the ecosystem? Depending on the wording, I could see that also being problematic, but it would allow drafting language that directly addresses the goal, rather than indirectly.

docs/SMCWG-charter.md
wthayer: There was a later conversation in Guangzhou in which Google (IMO) stated that it would not support the inclusion of identity in the scope of the charter at all, even if deferred to a later phase. Has that position changed? Does the proposed MAY imply that any member could veto work on identity validation?

docs/SMCWG-charter.md
wthayer: I thought the point of this statement was to ensure that S/MIME requirements don't prevent CAs from issuing certificates that are also used for other purposes, such as electronic signatures. If so, the proposal doesn't appear to retain that intent.

docs/SMCWG-charter.md
sleevi: @wthayer No, the proposed MAY does not imply that. It just simply doesn't commit the SMCWG to tackle that as the immediate next item, if new priorities come up.

I also don't think that's an accurate summary re: Guangzhou. Identity within the scope of the charter is certainly problematic for participation. This is because the inclusion in scope creates IP challenges and risks, and resolving those can take considerable time, preventing some members from joining or being able to contribute usefully to the core discussions unrelated to identity.

However, if identity is scoped as a fair topic from the get-go, which has been repeatedly agreed as not a goal, simply eliminates any hope of there being a useful or timely deliverable of core value, as shown by both the chartering of this WG and the SCWG experiences. If that's the case, it's likely valuable to not bother at all participating, and as a consequence of that lack of participation, not adopt any guidelines or audit criteria that might be produced.

docs/SMCWG-charter.md
wthayer: SGTM

docs/SMCWG-charter.md
sleevi: Yeah, this is really fluid depending on how the existing Certificate Consumers approach this. At the time this chartering was started, Certificate Consumers were still permissive here. If the goal is to produce something useful to the Certificate Consumers, and the Certificate Consumers do not already accept such audits / require different audits, then I suspect this isn't as valuable, since ultimately, this is about finding Guidelines that Certificate Consumers find broadly useful and acceptable for their needs.

docs/SMCWG-charter.md
wthayer: I would support changing this to "...shall include a document to..." to avoid arbitrarily limiting the WG to one deliverable, but I think it's meaningful to set the intent to deliver baseline requirements here.

sleevi:
Oh, I forgot to mention that any of the members should be able to either directly edit this PR in line, if I missed anything, or you can make the edits and have them committed to a branch in your repository and opening a PR against my PR/branch. This is in addition to commenting in line and using the suggestion tool.

I’ll create a branch-of-this-branch with all the proposed edits that might be controversial or complex, to show how this looks. Similarly, any member should be able to merge such a PR.

I mainly mention this to highlight that there are several ways we can make collaborative edits here, simultaneously, and without requiring any central approver.

docs/SMCWG-charter.md
wthayer: I disagree that "identity is scoped as a fair topic from the get-go" has been "repeatedly agreed as not a goal". I have certainly indicated a willingness to compromise on this point, but have also heard from other members who were less willing, leading me to the conclusion that the only resolution is to ballot it. I can support either approach, so I'll leave it to others to decide which route to take with the (new to me) knowledge that Google may be willing to participate initially if identity is explicitly scoped to a later phase of work.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,328 @@
+# S/MIME Certificate Working Group (SMCWG) Charter
+
+## Introduction
+
+This introduction provides general information and context with an intent to
+assist the interpretation of this Charter.
+
+An S/MIME certificate contains a public key bound to an email address; and may
+also contain the identity of a natural person or legal entity that controls such
+email address. The key pair can then be used to sign, encrypt, and decrypt
```
defacto64: I'd suggest to add "verify":     "sign, verify, encrypt, and decrypt"

docs/SMCWG-charter.md
clintwilson: I'm sure I lack a substantial amount of context, and while I would prefer identity validation be a focus of the WG only after guidelines regarding email validation and certificate profiles are produced (as the above proposed change accomplishes), this is an area I'm in agreement with Wayne on in that "identity" is an inclusion I'm willing to compromise on.

timfromdigicert: Thanks Clint and Wayne.  Let me reiterate that we support validation and profiles being the first priority, what we can't support is the exclusion of discussion of important topics from charter scope, especially given how long and difficult future charter updates will inevitably and unfortunately be.  No one is being asked to vote for or support any requirements, only to allow the topic to be discussed.  And the main reason we want to make it clear it can be discussed is because of a concerted effort to silence those who think it's an important topic.  The speed at which the group can accomplish its first priorities will depend on the willingness of participants to work together and dedicate the time and effort necessary to write a quality document; I am completely unconvinced that efforts to exclude important subjects from the WG scope will in any way help with that.  I'm very thankful for Apple, Microsoft and Mozilla's recent and substantial contributions to this charter, and look forward to actually being able to start productive work on minimum requirements validation and profiles for S/MIME certificates very soon.

sleevi:
I've created https://github.com/sleevi/cabforum-docs/pull/9 to attempt to address this feedback. I did not commit directly, in order to make sure @timfromdigicert was happy.

Tim: If you could indicate you're OK with the proposed wording on that PR, I can merge that PR into this and update the draft charter.

docs/SMCWG-charter.md
sleevi: @wthayer I tried to tweak this a little, in https://github.com/sleevi/cabforum-docs/pull/9

My understanding, and what I'm trying to capture here, is that this is trying to prevent the SMCWG from impinging on other CWGs (such as the SCWG).

That is, it's not about ensuring that a given sub-CA can issue whatever it wants, but that the scope of the SMCWG shall be focused to only those certificates that are technically capable.

In terms of a SHALL requirement, the existing language is ambiguous. What constitutes exercising caution? Is the SMCWG forbidden from taking up a TLS profile? Or could it just "exercise caution" and do so anyways? This tries to make it clear such an activity is out of scope.

That said, I agree that it doesn't and cannot prevent the SMCWG from defining a Root Certificate profile that is different from the SCWG's Root Certificate profile. The current language doesn't prevent that either, though.

I mention this, because the Browser Alignment ballot (which tries to align with existing root store policies) already has language regarding [EKUs that may be permitted](https://github.com/cabforum/documents/compare/master...sleevi:2019-10-Browser_Alignment#diff-7f6d14a20e7f3beb696b45e1bf8196f2R1559), and the current language may be seen as precluding from being adopted in this charter.

docs/SMCWG-charter.md
sleevi: @timfromdigicert Yes, the minutes from our meetings show there have been three options, basically since the very first time we were discussing:

1. Exclude from charter entirely, recharter after first deliverable (of core e-mail validation & profile)
2. Include in charter, gated upon delivering the first deliverable
3. Include in charter, without any limitations

The main distinction between 1 and 2 are the IP commitments and review that is entailed in producing work. The main distinction between 2 and 3 is making sure that we don't get bogged down in endless discussions about identity without having made some progress on the core deliverable.

(1) was Google's proposal, because that represents the broadest possible charter to allow collaboration. This is important, especially as we've seen how past SCWG chartering efforts have prevented a number of members to participate.

(2) was DigiCert's proposal, as reflected in the minutes. This attempts to avoid needing to recharter, despite this being "just" a Forum ballot to adopt language such as shown in https://github.com/sleevi/cabforum-docs/pull/9 , but otherwise scopes the SMCWG's activities to solving the core problems first.

(3) is what this charter is.

https://github.com/sleevi/cabforum-docs/pull/9 attempts to do (2). As noted, this creates more barriers for collaboration, and the value provided by avoiding a recharter is potentially greatly overstated. The primary benefit of (1) vs (2) is it allows more participation, quicker, by more members, but it's clear that's not a shared value. However, (3) is very much a non-starter, because without the limits (such as in (2)), there's a clear set of interests that prevent a useful deliverable in a timely fashion.

docs/SMCWG-charter.md
clintwilson: Are there (should there be?) any restrictions around who can be a designated Representative?

docs/SMCWG-charter.md
wthayer: @sleevi sounds like we hold different beliefs on the purpose of that sentence. Again, I think it is a weak attempt to instruct the SMCWG not to institute rules for S/MIME that revent the use of the same cert for other uses - primarily electronic signatures. If so, the #9 language doesn't help. Retaining the existing language but replacing the SHALL with a SHOULD would do that (but the meaning would still be unclear).

Having said that, I'm arguing a position that I don't really care about. I'm happy with the #9 language as long as it doesn't later become a sticking point. @timfromdigicert WDYT?

sleevi:
Entirely fair points!

The current language may be trying to capture a goal of “The SMCWG SHOULD
NOT place limits on what certificates intermediates may issue”, if I’m
understanding you correctly?

My proposed edit is trying to say “The SMCWG SHALL only write Guidelines
for S/MIME certificates”, without limiting the possibility of putting
restrictions on what intermediates may issue. As some root stores already
place limits on what intermediates may issue, it seemed important to allow
this to be reflected.

Definitely, as this is Tim’s ballot, I’m hoping he’ll review each of the
proposed edits and clarify any concerns. In the end, it may not matter:
root stores can, will, and do still limit what types of certificates
intermediates can be used to issue, and that won’t be impacted by what the
charter says.

In any event, we still need the charter to reflect that it’s only dealing
with S/MIME policies, which no interpretation of the current language does.
If we need to keep the existing restrictions on what the Guidelines can say
about intermediate issuance, we should tighten that up, but we should still
limit the scope to S/MIME only.

docs/SMCWG-charter.md
timfromdigicert: I think there's some in sleevi#9, but it needs some tweaks.  I need to go over it more carefully.

docs/SMCWG-charter.md
timfromdigicert: If someone wants to try to obstruct the work of the committee, there are plenty of options available to do regardless of what's in scope or not.  I'm much less concerned that's an actual issue, as I haven't seen or heard any evidence either publicly or privately that indicate anyone is planning to do so, and it's a pretty serious accusation.  Furthermore, in the event it happens, the best recourse is the Code of Conduct, which explicitly forbids such activities.

docs/SMCWG-charter.md
timfromdigicert: No, people should be free to be able to choose their own Representatives.

docs/SMCWG-charter.md
sleevi: @timfromdigicert I'm sorry you felt this is a "pretty serious accusation", especially since I don't know what "this" is you're referring to.

The minutes show that there is active interest in tackling identity as a co-equal, or even primary, deliverable among CA members. I'm happy to point out specific instances, if you'd like public examples. Such an approach would not be a violation of the Code of Conduct, because it would be entirely within the chartered scope. I think you may have interpreted (3) as suggesting some malicious intent, but nothing of the sort: merely, good faith attempts at debating identity will, and clearly have for two years, impeded progress on a charter.

That's exactly why a charter like (2) is essential to setting forth priorities. If there is not agreement on a consistent set of priorities, and the ballot fails, then that's a useful and important signal. The entire point of the chartering process is to help lay out those baseline priorities, so that it's clear if and when something does become a Code of Conduct problem.

This is why (1) or (2) are the only possible charters we can support, and (2) is likely a charter that will prevent us from joining for some time. (3) is a charter that we can neither join nor support, because without that prioritization, it is an unwise and unproductive use of our time.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,334 @@
...
+The primary use case under consideration for the working group is a model
+whereby senders and recipients of email messages receive “reasonable assurance”
+that the other party to the communication identified in the certificate has
+control of the domain or email address being asserted. A variation of this
+primary use case is where an individual or organization digitally signs email to
+establish its authenticity and source of origin.
+
+Therefore, in order to provide reasonable assurance, it is crucial to establish
+a standard method to validate an email address and the subject’s identity (if
+present) prior to binding it them to the email addresspublic key. “Reasonable
```
CBonnell, line 34: Not sure what's going on with the last part of this sentence. Perhaps this should be "prior to binding them to the public key"?

sleevi: Ah, yeah, thanks. This was a draft edit that got copied wrong.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,334 @@
...
+  SMCWG’s business as specified in the ballot (each, a “SMCWG Subcommittee”).
+  SMCWG Subcommittees are open to all SMCWG Members. A SMCWG Subcommittee may
+  work on and recommend ballots, complete delegated functions, or issue reports
+  to the SMCWG that are within the scope of this charter. SMCWG Subcommittees
+  must post all agendas and minutes on a public mail list.
+
+### 7. Summary of Major Deliverables
+
+The primary deliverable of the SMCWG is set forth in the Scope section, and
+shall include a Draft Guideline be named the "Baseline Requirements for S/MIME
```
CBonnell, line 306: Consider naming the deliverable "Baseline Requirements for the Issuance and Management of Publicly-Trusted S/MIME (or Email Protection) Certificates" to align with the naming convention used by the BRs and CSBRs.

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,334 @@
...
+- Identity validation for natural persons and legal entities, in the context of
+  S/MIME certificates
+
+Deliverables of the SMCWG SHALL be constrained to those certificates
+that contain the emailProtection (OID: 1.3.6.1.5.5.7.3.4) or that are technically
+capable of such issuance.
+
+### 2. Out of Scope
+
+Certificates issued under a root certificate that is not publicly trusted SHALL
```
CBonnell, line 102: It may be valuable to specify how such certificates are out of scope of any deliverables in a similar manner described in the servercert WG and code signing WG charters.

My proposed change, which borrows from the CSCWG charter:
"The SMCWG SHALL NOT develop guidelines, standards, or requirements applicable to certificates issued under a root certificate that is not publicly trusted".

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,334 @@
...
+The primary use case under consideration for the working group is a model
+whereby senders and recipients of email messages receive “reasonable assurance”
+that the other party to the communication identified in the certificate has
+control of the domain or email address being asserted. A variation of this
+primary use case is where an individual or organization digitally signs email to
+establish its authenticity and source of origin.
+
+Therefore, in order to provide reasonable assurance, it is crucial to establish
+a standard method to validate an email address and the subject’s identity (if
+present) prior to binding it them to the public key. “Reasonable
```
CBonnell, line 34: I think the "it" needs to be dropped here.

docs/SMCWG-charter.md
sleevi: I’m not sure I see the connection between your problem statement and the proposed solution. Could you explain?

Your proposed change seems to restrict what the WG can work on, but in a way that itself isn’t defined and is, at best, seemingly self-referential to itself, since publicly trusted may ultimately be decided by what’s in scope.

CBonnell: My intention is to better define how such untrusted certificates would be out of the scope of work for the WG, in a similar manner that the servercert WG and code signing WG charters exclude non-publicly trusted certificates from their deliverables. As the language currently stands, it's stated that such certificates are out of scope, but doesn't explain the net result/action of such exclusion (which, in this WG's case, would be not producing deliverables concerning them).

I drew inspiration from the language in 2 (c) of the CSWG charter as I thought that language fits the bill to both describe the certificates that are out of scope and how they will not be dealt with as part of the WG's deliverables.

sleevi: But doesn’t the current language solve that better than the proposed language?

Your proposal seems to prohibit the use of the work product, which the CABF can’t do, rather than describe the problem scope.

In order for the proposed language to work, “publicly trusted” would need to be defined here in a way that’s independent from the work product, but that also seems like an anti-goal?

I’m not trying to say the current language couldn’t be improved - it seems like it could - but merely for the problem I’m not sure I understand how the proposal is better than the current language, other than “This is what CSCWG did”. However, I’m probably misunderstanding and that’s why I wanted to try to figure the disconnect.

CBonnell: I'm not sure how my proposed language would prohibit the use of such work products; the only prohibition is on the WG's development of such deliverables.

In other words, I don't think the proposed language would prevent some arbitrary non-public S/MIME PKI for adopting the WG deliverable as its CP. However, deliverables for such a PKI would be explicitly out of scope for the WG.

sleevi: But they wouldn’t, because “publicly trusted” is ill-defined here, and presumably end up self-referential (that is, “publicly trusted = use these standards”, which then defines any new requirement as functionally out of scope)

It seems there’s two dimensions here:
1. Clarifying that these requirements are voluntary and only apply to those who use them
2. Clarifying that the WG won’t impinge on other WGs (e.g. by focusing on EKU)

It seems we want both to be clear? Is that a fair statement?

docs/SMCWG-charter.md
sleevi: You can also use "suggestion" to propose an edit inline

```suggestion
present) prior to binding them to the public key. “Reasonable
```

docs/SMCWG-charter.md
sleevi: @timfromdigicert You can make edits depending on how you feel here.

docs/SMCWG-charter.md
CBonnell: I think the definition for "publicly trusted" would be closer to "chains to a root certificate that is in the trust store of at least one Certificate Consumer", but I agree that it's better to use Defined Terms or similarly well-specified terms where possible to prevent differences in understanding.

As for the two dimensions, I think the second point is valuable to specify, although one could argue that it is already specified in section 1 ("Deliverables of the SMCWG SHALL be constrained to those certificates
that contain the emailProtection (OID: 1.3.6.1.5.5.7.3.4) or that are technically
capable of such issuance"). However, I'm not sure how the first point is relevant to defining the scope of work for a WG, as it seems to be speaking to the applicability of such deliverables to a given PKI.

timfromdigicert: Yeah, I suggested "at least one Certificate Consumer" recently.  I also think "Root Certificate" is superior to "root certificate".  I'm not a huge fan of the current definition of "Root Certificate", but at least it exists.  "Publicly-Trusted Certificate" is also a defined term in the BRs, but it's (IMO) a bad definition, and shouldn't be used here.

docs/SMCWG-charter.md
timfromdigicert: I think alignment is good.  Agree? @wthayer @benwilsonusa @clintwilson

clintwilson: I'm good with Corey's suggested edit, though don't feel strongly about it.

wthayer: Yes, I accept this change.

docs/SMCWG-charter.md
wthayer: Does someone want to propose a change here?

timfromdigicert: "Certificates issued under a Root Certificate that is not trusted by a Certificate Consumer SHALL be out of scope."

sleevi: @timfromdigicert

It would be great to avoid self-referential definitions. This seems especially important because the mere act of a single member (or one applying for membership) could change the entire scope of the WG. That's why I was trying to see if there was agreement in https://github.com/cabforum/documents/pull/167#discussion_r417736594 about what this language is trying to accomplish.

Put differently: Do we lose something of value by removing this statement entirely? It's not clear to me that we would.

docs/SMCWG-charter.md
sleevi: Fixed in 2aa376c06b45146249d0cc6b8cc5d42d08ccb177

docs/SMCWG-charter.md
timfromdigicert: Yes, but neither of your options capture what this is trying to accomplish, which is important and shouldn't be removed.  As I stated on the public list, if we can't improve it, I support leaving it as is.

sleevi: @timfromdigicert If you believe there's something of value, could you rephrase what you believe is important here. Right now, this makes no sense as to what it's trying to accomplish, and I'd like to try to find a better way of rephrasing, especially if there's a misunderstanding about something that's important to you. I haven't seen any explanation from you, but let me know if I missed something.

timfromdigicert: In my mind, the point is related to the following paragraph from the overview of the server BRs, parts of which may be worth stealing:

"This document describes an integrated set of technologies, protocols, identity-proofing, lifecycle management, and auditing requirements that are necessary (but not sufficient) for the issuance and management of Publicly-Trusted Certificates; Certificates that are trusted by virtue of the fact that their corresponding Root Certificate is distributed in widely-available application software. The requirements are not mandatory for Certification Authorities unless and until they become adopted and enforced by relying–party Application Software Suppliers."

I think it is important that the scope is explicitly limited to certificates that chain to Root Certificates that are trusted by Certificate Consumers who run root programs that voluntarily adopt the produced guidelines.

sleevi: Oh, I definitely think it's valuable context, but I'm not sure how it relates to what's out of scope.

What's an example of something that we want to use "Out of Scope" to prohibit? That might help figure out where the language is.

I can, for example, imagine someone saying "Oh, hey, we at Fido Corp use this policy, and we think it's useful". One argument is that even mentioning what Fido Corp does should be out of scope, because, see charter. But it's also in scope, because it's proposing that these certs should also consider what Fido Corp does.

Now, let's say Fido Corp wants to propose this, and we think it's not actually useful for these certs. The charter, as written, doesn't keep Fido Corp from bringing it up. All they have to do is join the Forum, and then suddenly their requirements on their certificates are relevant.

That's why I was suggesting "just delete", because everything I could think of that it would try to keep out, it wouldn't, and it would seem to keep out things it shouldn't (even though it doesn't)

docs/SMCWG-charter.md
```diff
@@ -0,0 +1,334 @@
...
+address. While there are methods for validating control of a domain that can be
+adopted from TLS certificates, there are currently no standard requirements for
+validating control of email addresses. Methods for validating identities in TLS
+certificates also exist and should be leveraged where possible, as well as other
+identity validation standards common in the industry.
+At least one method to efficiently validate an email address as well as
+developing a consistent profile for S/MIME certificates in order to facilitate
+technical interoperability across the Internet are also to be addressed by
+the SMCWG.
+
```
sleevi, line 61: @timfromdigicert Does it make sense to bring the paragraph you mentioned in https://github.com/cabforum/documents/pull/167/files#r421068417 here?

I could also see it moved to Line 99 with some tweaks, if you want it more explicitly in scope.

Independent of the discussion on "out of scope", it does seem useful.

docs/SMCWG-charter.md
timfromdigicert: What is something it seems to keep out, but shouldn't?

docs/SMCWG-charter.md
timfromdigicert: I would support that.  I would put it at a "nice to have" level of importance.

docs/SMCWG-charter.md
sleevi: @timfromdigicert Did you perhaps miss https://github.com/cabforum/documents/pull/167#discussion_r421074233 which directly answered your question?

timfromdigicert: I thought you had an explicit example beyond that.  Not only do I disagree that the language prohibits what you suggest, I can't see how an argument could be made that it does, so you'd have to walk me through why you believe that.

sleevi: @timfromdigicert 

I was asking you for an explicit example, to help find workable language. You can see it in the questions I've asked several times now:
https://github.com/cabforum/documents/pull/167#discussion_r421074233
> What's an example of something that we want to use "Out of Scope" to prohibit? That might help figure out where the language is.

https://github.com/cabforum/documents/pull/167#discussion_r420986318
> Put differently: Do we lose something of value by removing this statement entirely? It's not clear to me that we would.

I think it's useful, in order to make forward progress, if you could describe some important thing you feel that is worth keeping out, and that you're trying to do so with this language, so we can find better language.

https://github.com/cabforum/documents/pull/167#discussion_r421074233 is exactly the kind of example of how the language can be argued both ways, which you seem to be doing in https://github.com/cabforum/documents/pull/167#discussion_r424730355 if not aware, and that's why I want to try to find better language with you.

timfromdigicert: To be clear, I'm fine with the current language.  As I have said several times, I think "root certificate" -> "Root Certificate" and defining public trust in terms of Certificate Issuers are useful improvements, but if there is not enough support for that, the current language is fine, and I believe no actual problems with it have been identified.  The burden is on those who want changes to propose and justify them.

docs/SMCWG-charter.md
sleevi: https://github.com/sleevi/cabforum-docs/pull/19

docs/SMCWG-charter.md
sleevi: @timfromdigicert We're going in circles here.

I'm aware you're fine with the language. You have yet to describe what you want to accomplish with the language, which continues to be a problem to finding a productive solution to the problem that we are *not* fine with the language, and want to find a way to solve your problem without introducing new problems.

The attempt in https://github.com/cabforum/documents/pull/167#discussion_r421074233 was an attempt to understand your problem. You clearly believe it failed to do so, so again, I'm asking you to state an example of a thing you're trying to keep out of scope with this language.

Every single attempt to find something that is prohibited by this language I can find a way to make in-scope. The fact that anything can be read both ways *is a problem*, and that's the problem to solve: to make it clear and unambiguous what it is you want. The current language, unquestionably, does not do this.

BenWilson-Mozilla: I think the out-of-scope language should be removed. The resulting CABF guideline should say that it is binding on those certificate consumers when they adopt it as their standard.

timfromdigicert: Ryan, I haven't seen any objections to my email to the public list on Friday, so can you please delete this sentence?

sleevi: Done.

dzacharo:
I think this is ready to be merged. @sleevi do we need to "Resolve Conversations" before "Squash and merge"?

PR #173, opened 2020-04-08
Changed Random Number to Random Value

mtgag:
Small change to use the specified term (Random Value) in Section 1.6.1


dzacharo:
Thanks @mtgag, I opened #174 to capture this for our spring cleanup ballot.

PR #185, opened 2020-05-27
Bylaws version 2.3

dzacharo:
Ballot "Forum-12" has passed. Updating version 2.2 to 2.3.


sleevi:
@dzacharo Sorry I didn't think about this on the infrastructure call, but does it make sense that when doing these squash and merge commits, the commit description is something meaningful?

For example, the ballot text or a link to the ballot / vote?

The commit messages are useful to understanding how/why things changed, and the individual commits may not be as useful as understanding the process flow.

dzacharo:
Yes, this makes sense and we can add it in the process described on the wiki. I also asked if there is a way to add a document version as tag in individual files (BRs, EVG, Bylaws, etc). Let's continue this discussion to the infrastructure list.

PR #188, opened 2020-05-30
Update NCSSRs with Ballot SC29

BenWilson-Mozilla:
I'm sorry but I got confused. I think I merged my proposed changes into my master instead of leaving it in my branch for SC29. Is this backwards from the proposed way of doing this?


BenWilson-Mozilla:
Now I'm even more confused.  I thought there was only 1 commit.  Argh!

wthayer:
The second commit looks like you synching your fork with cabforum/documents - possibly by following the redline instructions?

I think this PR is okay, but we need to decide of we're adding 'TS' or not.

dzacharo:
Approved via ballot [SC29](https://cabforum.org/2020/05/07/ballot-sc29v3-system-configuration-management/)

PR #194, opened 2020-06-16
SC30 - Disclosure of Registration/Incorporating Agency

sleevi:
# Purpose

The EV Guidelines aim to ensure a consistent and repeatable level of validation for certificates, regardless of the CA performing the validation, providing Relying Parties consistency for all certificates complying with these Guidelines. Although the Guidelines attempt to specify objective requirements, areas remain that rely on a subjective determination by the CA. One such area is determining whether a given Incorporating Agency or Registration Agency fulfills these Requirements. 

As currently specified, it's possible for one CA to make a determination that a given Registration Agency or Incorporating Agency does meet the requirements of the EV Guidelines, while a different CA determines that same Agency does not. As the reliability of the information validated within the Certificate is tied to the reliability of the data source used to verify this information, this inconsistency undermines the assurance that EV Certificates are meant to provide.

While there is utility in being able to identify precisely what datasource(s) were used with a given Certificate, this ballot does not involve such work. It merely seeks to ensure that, for any given Organization, it can be validated consistently and to the same degree, regardless of the CA, by working to achieve consistency among all CAs in their selection of data sources.

Much like the work to remove "Any other method" from the validation of domain names, ensuring consistency, transparency, and objectivity in validating domain names, this ballot is the first step to doing the same for organization information.

A potential roadmap of ballots to to address these issues involves:
* CAs publish the list of Registration Agencies / Incorporating Agencies they use (this ballot)
* Create an allowed list of Registration Agencies / Incorporating Agencies and associated values, along with a process for updating and adding new ones, and requiring issuance exclusively use Agencies on this list.
* If useful and relevant to Relying Parties, ensure each Certificate can be tied back to their Registration Agency / Incorporating Agency, such as disclosure within the Certificate itself, so they can unambiguously and uniquely determine the organization that has been validated.

A similar process may then be repeated for other forms of verification data sources, such as the QIIS, QTIS, and QGIS within the EV Guidelines, or the Reliable Data Sources within the Baseline Requirements.

### DIFF
--- a/docs/EVG.md
+++ b/docs/EVG.md
@@ -346,6 +346,12 @@ Each CA MUST publicly disclose its Certificate Policy and/or Certification Pract
 
 Effective as of 31 May 2018, the CA's Certificate Policy and/or Certification Practice Statement MUST be structured in accordance with RFC 3647. Prior to 31 May 2018, the CA's Certificate Policy and/or Certification Practice Statement MUST be structured in accordance with either RFC 2527 or RFC 3647. The Certificate Policy and/or Certification Practice Statement MUST include all material required by RFC 3647 or, if structured as such, RFC 2527.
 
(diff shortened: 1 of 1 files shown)


sleevi:
> Would it be acceptable to allow a certain margin to disclose the agency “after x days of issuance”, instead of forcing it to be always “prior”?

https://archive.cabforum.org/pipermail/servercert-wg/2020-June/001983.html

PR #204, opened 2020-07-27
Update .gitignore

sleevi:
Add .docx to the .gitignore, now that Pandoc is integrated


sleevi:
Jos: An easy fix to merge (and delete branch when done)

sleevi:
@wthayer Looks like Jos might be busy, could you review? :)

PR #214, opened 2020-09-14
SC28v6

dzacharo:
Sync with ballot SC28v6

### DIFF
--- a/docs/NSR.md
+++ b/docs/NSR.md
@@ -90,7 +90,7 @@ Certification Authorities and Delegated Third Parties SHALL:
 
 a.	Implement a Security Support System under the control of CA or Delegated Third Party Trusted Roles that monitors, detects, and reports any security-related configuration change to Certificate Systems;
 
-b.  Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity, and enable those systems to log and continuously monitor the events specified in Section 5.4.1.3 of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates;
+b.  Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity, and enable those systems to log and continuously monitor the events specified in Section 5.4.1 (3) of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates;
 
 c.	Implement automated mechanisms under the control of CA or Delegated Third Party Trusted Roles to process logged system activity and alert personnel, using notices provided to multiple destinations, of possible Critical Security Events; 
 
--- a/docs/NSR.md
+++ b/docs/NSR.md
@@ -90,7 +90,7 @@ Certification Authorities and Delegated Third Parties SHALL:
 
 a.	Implement a Security Support System under the control of CA or Delegated Third Party Trusted Roles that monitors, detects, and reports any security-related configuration change to Certificate Systems;
 
-b. Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity and log and continuously monitor the events specified in Section 5.4.1 of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates;
+b.  Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity and log and continuously monitor the events specified in Section 5.4.1 of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates;
 
 c.	Implement automated mechanisms under the control of CA or Delegated Third Party Trusted Roles to process logged system activity and alert personnel, using notices provided to multiple destinations, of possible Critical Security Events; 
 
--- a/docs/NSR.md
+++ b/docs/NSR.md
@@ -90,7 +90,7 @@ Certification Authorities and Delegated Third Parties SHALL:
 
 a.	Implement a Security Support System under the control of CA or Delegated Third Party Trusted Roles that monitors, detects, and reports any security-related configuration change to Certificate Systems;
 
-b.  Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity and log and continuously monitor the events specified in Section 5.4.1 of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates;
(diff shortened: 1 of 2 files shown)


dzacharo (commented):
@sleevi I tried to sync Neil's ballot (https://github.com/cabforum/documents/compare/16a5a9b...neildunbar:498c5ad) into a new branch SC28. After this, I intend to create a pull request to a new branch "SC35+SC28" that will contain voted changes for both ballots once the Review period is over.

Does this make sense? If yes, please approve and merge this PR.

sleevi:
@dzacharo Sorry, I'm having trouble following what you did.

You've set the target of this PR to merge from Neil's branch into a branch in the CABF SC28. That much makes sense. However, what I'm having trouble following is your "new branch SC35+SC28".

In https://github.com/cabforum/documents/pull/213 you equally did something where you created a new branch SC35, but then merged from SC35 into SC35+SC28.

This doesn't seem to match what we have on our GitHub Redline Guide, but I suspect it's because we had two ballots in flight, and judging from your [note to the mailing list](https://archive.cabforum.org/pipermail/servercert-wg/2020-September/002252.html) that it sounds like you:

- Followed the instructions for SC35 (merging into an SC35 branch)
- Realized you forgot to address SC28
- Created a new branch SC35+SC28
- Want to merge this into a branch SC28 (why?)
- To then merge this into SC35+SC28, to produce the Redline
  - Which you've done in https://archive.cabforum.org/pipermail/servercert-wg/2020-September/002253.html

However, the fact that GitHub is showing that there are conflicts here means that SC28 was not appropriately updated for one or more ballots that modified a similar section. So I'm not sure how you produced [this redline](https://archive.cabforum.org/pipermail/servercert-wg/2020-September/002253.html), since you should have been prevented from doing it.

dzacharo:
I had to create the redline manually, using MS word because I needed to find a way to start the Review Period.

My plan was to create a common branch for the two ballots, including commits from both ballots, resolve any possible conflicts in case they made changes to the same sections (I don't think we had any such conflicts), which would then result in a single Pull Request against the Main branch, which would then produce the new BRs, EVG, NCSSRs. In the meantime, creating a combined branch would result in having the artifact of the BRs that would include the changes of SC35 and SC28, making it easier to create a redline.

I am sorry for the confusion, I'm struggling with these instructions :-) Is there another workflow that I could use?

sleevi:
> I had to create the redline manually, using MS word because I needed to find a way to start the Review Period.

I see. I think that's concerning, because if there are issues, we should sort them out to make sure they aren't fundamental.

The fact that there are conflicts here, with SC28, *should* be concerning and is a big issue.

> My plan was to create a common branch for the two ballots, including commits from both ballots, resolve any possible conflicts in case they made changes to the same sections (I don't think we had any such conflicts), 

Looking at GitHub right now, we see there are conflicts with what you've done and docs/BR.md. 

> which would then result in a single Pull Request against the Main branch, which would then produce the new BRs, EVG, NCSSRs. In the meantime, creating a combined branch would result in having the artifact of the BRs that would include the changes of SC35 and SC28, making it easier to create a redline.
> 
> I am sorry for the confusion, I'm struggling with these instructions :-) Is there another workflow that I could use?

The instructions were not designed to support the workflow you wanted, but that was intentional. It was intended that each ballot would/could result in a new version of the Guidelines, rather than incorporating multiple Ballots into a single version. I thought the last time was exceptional, and didn't realize you planned to do it continually in the future (aggregating ballots).

Among other things, it's troubling for our IP review, because now the fate of both ballots are tied in a single version, which is something our policies explicitly tried to avoid.

In any event, the reason you're having trouble is with the conflicts. The reason there are conflicts has to do with the ballot being against version 1.6.7. This actually violates our Bylaws, and means the Ballot itself was not valid: see https://cabforum.org/wp-content/uploads/CA-Browser-Forum-Bylaws-v2.3.pdf 

> Such redline
> or comparison **shall be made against the Final Guideline section(s) as they exist
> at the time a ballot is proposed**, and need not take into consideration other
> ballots that may be proposed subsequently, except as provided in Section 2.4(10)
> below.

(Emphasis added)

Specifically, Ballot SC28 was proposed on 2020-07-10, where the then-current version of the BRs was 1.7.0, per [SC26](https://cabforum.org/2020/03/30/ballot-sc26v2-pandoc-friendly-markdown-formatting-changes/)

Now, I understand the Ballot provided instructions (on sections to delete and add), but it was not against the Section(s) as they exist at the time the ballot was proposed. The inconsistency is not between the Redline and the Draft Ballot, which we have policies for.

The reason you're having trouble is that Neil's ballot conflicts with SC26, which you can see more easily with the [blame](https://github.com/cabforum/documents/blame/6b870f92d4788a52c2bbc9d96a1db17751e906b1/docs/BR.md#L1316-L1317), which changed the same sections that SC28 is trying to touch.

Now, "luckily", we're in a case where the only changes made by [SC26](https://github.com/cabforum/documents/pull/165/files) was to whitespace. SC28 is... ambiguous... as to what it proposes regarding whitespace, since the GitHub link isn't part of the Ballot, so this is further messy.

dzacharo:
Shall I use the ballot language to repeat the changes to a new SC28 branch (under the cabforum/documents repo) with a new commit? That would resolve the conflict and allow it to be merged to main or even the combined branch before that.

docs/NSR.md
```diff
@@ -90,7 +90,7 @@ Certification Authorities and Delegated Third Parties SHALL:
 
 a.	Implement a Security Support System under the control of CA or Delegated Third Party Trusted Roles that monitors, detects, and reports any security-related configuration change to Certificate Systems;
 
-b.	Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity and enable those systems to continuously monitor and log system activity;
+b.  Identify those Certificate Systems under the control of CA or Delegated Third Party Trusted Roles capable of monitoring and logging system activity, and enable those systems to log and continuously monitor the events specified in Section 5.4.1 (3) of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates;
```
sleevi, line 93: This spacing is off, relative to the surrounding sections.

sleevi (approved):

sleevi:
I'm marking this approved, but I'm not sure why the Web UI is not allowing to resolve conflicts here, since it normally should. If the GitHub had followed the Wiki, we'd be gravy.

https://github.com/sleevi/cabforum-docs/tree/SC28v6-Corrected has this corrected version, squashed into a single commit to fixup the whitespace issues.

dzacharo:
Incorporated in https://github.com/cabforum/documents/pull/224

PR #220, opened 2020-09-23
Ballot sc33

wthayer:
Notes:
- I set the date to today rather than the end of the SC33 review period
- I didn't include SC33 in the 'relevant dates' table because the ballot didn't specify an effective date for method 10's demise.


dzacharo (approved):
LGTM

PR #227, opened 2020-11-04
Cleanup non-SCWG related artifacts

sleevi:
The deleted files now live in other repositories, such as:
  * https://github.com/cabforum/forum
  * https://github.com/cabforum/code-signing
  * https://github.com/cabforum/smime
  * https://github.com/cabforum/archive


dzacharo (approved):
LGTM

PR #234, opened 2020-12-02
Add a GitHub Workflow to do Pandoc conversion

sleevi:
This adds a basic workflow that will convert `docs/BR.md`, `docs/EVG.md`, and `docs/NSR.md` to `pdf` and `docx` using the https://github.com/cabforum/build-guidelines-action GitHub Action, attaching the results as artifacts to the Workflow run.

By default, these artifacts are only maintained for 21 days, which is less than the 90 day max for GitHub Free. 21 days was chosen to support ballots in the Discussion period, which time out after three weeks of no action. The workflow can always be re-triggered manually, if necessary.

GitHub uploads all artifacts as `zip` files, so this does mean an extra step to open the zip file to extract the associated `pdf` and `docx`, but that should hopefully be easy.

Redlines are not yet supported in https://github.com/cabforum/build-guidelines-action, so these aren't yet generated.

Compared to the existing Travis-CI integration, the benefit of this approach is that these actions will also, by default, run in repository forks/clones, making it easy for CA/B Forum Members to preview their changes locally.

Note that further action is required to correct the BRs/EVGs/NCSSRs, specifically:
- Move the YAML preamble into the document itself
- Convert the EVGs/NCSSRs to a consistent flavor of Pandoc markdown (mostly around list items and extensions)

However, by landing the tooling first, it makes it easy to preview and verify that changes to address the above are correct.

### DIFF
--- a/.github/workflows/build-draft-docs.yml
+++ b/.github/workflows/build-draft-docs.yml
@@ -6,7 +6,7 @@ jobs:
       - name: Checkout the documents
         uses: actions/checkout@v2
       - name: Build the Baseline Requirements
-        uses: cabforum/build-guidelines-action@v1.0.0-rc3
+        uses: cabforum/build-guidelines-action@sleevi_scratch
         with:
           markdown-file: docs/BR.md
           pdf: true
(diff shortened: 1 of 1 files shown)


wthayer (approved):

pes2020kk (commented):
ipad

pes2020kk (commented):
ipad

PR #245, opened 2021-02-08
Correction of wrong numbering in markdown - file

RufusJWB:
In the raw mark down file there are two places where the numbering of a list is not correct. Interestingly in the rendered version it is correct.


sleevi:
@RufusJWB Unfortunately, you need to sign the IP agreement to send Pull Requests. Please note that SC41 is also in process.

sleevi:
Also, this is Markdown auto-numbering (and is fixed in SC41)

RufusJWB:
> @RufusJWB Unfortunately, you need to sign the IP agreement to send Pull Requests. Please note that SC41 is also in process.

I know, but I was hoping that such an editorial change would be acceptable.

sleevi:
Unfortunately, changes can only be proposed by Members.

However, we welcome all reports of issues, through our
questions@cabforum.org alias or, informally, through raising a GitHub issue
in the repository for the document (although this is not yet formally
recognized and not guaranteed to result in a response).

PR #250, opened 2021-02-26
SC41: Reformatting the BRs, EVGs, and NCSSRs

castillar:


.github/workflows/build-draft-docs.yml
```diff
@@ -11,19 +11,28 @@ jobs:
...
-      - uses: cabforum/build-guidelines-action@v1.0.0-rc4
+      - name: Checkout the code
+        uses: actions/checkout@v2
+      - name: Checkout old version for redline (pull request)
+        uses: actions/checkout@v2
         with:
-          markdown-file: docs/${{ matrix.document }}
+          ref: ${{ github.event.pull_request.base.sha || github.event.push.before }}
+          path: old/
+      - uses: cabforum/build-guidelines-action@v2.0.0
```
sleevi, line 21: @cabforum/servercert-chairs Y'all _should_ be able to change this, because it's not part of any of our docs, but I'll let y'all be the ones to do that. However, this should just be tweaked to one of the following:

```suggestion
      - uses: cabforum/build-guidelines-action@v2.0.1
```

or

```suggestion
      - uses: cabforum/build-guidelines-action@v2.0
```

Either of those can get the thing building again, which you can confirm before merging.

wthayer (approved):

PR #260, opened 2021-04-01
Ballot SC42: 398-day reuse period

sleevi:
(Not to be merged into SC41, but using that as the base for when SC41 is approved)

Original message from @BenWilson-Mozilla at https://archive.cabforum.org/pipermail/servercert-wg/2021-March/002586.html


sleevi:
@castillar Can you re-open this (it's greyed out for me, I think, but not sure, due to permissions)?. This was meant to be retargeted onto 'main' now that SC41 is merged.

If you can't re-open, that's fine, it means I'll need to work through the workflow in Infra and document the best way to 'layer' ballots :)

PR #263, opened 2021-04-06
Remove legacy infrastructure tooling

sleevi:
With the adoption of SC41, we can remove the Weasyprint and
Kramdown infrastructure, as well as the related in-repository tools
for building the docs. This has all been replaced with the
https://github.com/cabforum/build-guidelines-action Docker image
and GitHub Action.


castillar (approved):
Looks good! Nice to do some spring cleaning. :)

PR #265, opened 2021-04-15
Ballot sc44 clarify acceptable status codes

n0ot:
This clarifies the acceptable HTTP status codes that can result in a redirect, when CAs validate domains using methods 18 and 19. The effective date for the change is 1 Jul 2021.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -818,12 +818,13 @@ The file containing the Request Token or Random Number:
 3. MUST be retrieved via either the "http" or "https" scheme, and
 4. MUST be accessed over an Authorized Port.
 
-If the CA follows redirects the following apply:
+If the CA follows redirects, the following apply:
 
-1. Redirects MUST be initiated at the HTTP protocol layer (e.g. using a 3xx status code).
-2. Redirects MUST be the result of an HTTP status code result within the 3xx Redirection class of status codes, as defined in RFC 7231, Section 6.4.
-3. Redirects MUST be to resource URLs with either via the "http" or "https" scheme.
-4. Redirects MUST be to resource URLs accessed via Authorized Ports.
+1. Redirects MUST be initiated at the HTTP protocol layer.
+   a. Effective July 1, 2021 for new validations, Redirects MUST be the result of a 301, 302, 307, or 308 HTTP status code response and MUST be to the resource URL contained in the Location HTTP response header.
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -826,7 +826,7 @@ The file containing the Request Token or Random Number:
 If the CA follows redirects, the following apply:
 
 1. Redirects MUST be initiated at the HTTP protocol layer.
-   a. For validations performed on or after July 1, 2021, redirects MUST be the result of a 301, 302, or  307 HTTP status code response, as defined in [RFC 7231, Section 6.4](https://tools.ietf.org/html/rfc7231#section-6.4), or a 308 HTTP status code response, as defined in [RFC 7538, Section 3](https://tools.ietf.org/html/rfc7538#section-3). Redirects MUST be to the final value as defined in [RFC 7231, Section 7.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.2) specified in the Location HTTP response header.
+   a. For validations performed on or after July 1, 2021, redirects MUST be the result of a 301, 302, or 307 HTTP status code response, as defined in [RFC 7231, Section 6.4](https://tools.ietf.org/html/rfc7231#section-6.4), or a 308 HTTP status code response, as defined in [RFC 7538, Section 3](https://tools.ietf.org/html/rfc7538#section-3). Redirects MUST be to the final value of the Location HTTP response header, as defined in [RFC 7231, Section 7.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.2).
```
cardonator, line 829: @sleevi I believe this alteration makes the intended meaning of the point a little less clear. What does "final value of" mean in this context? I'm assuming that it's something along the lines of "Redirects MUST follow the Location HTTP response header and the Location HTTP response header MUST point to the page containing the random value"?

cardonator: Sorry, meant to tag @CBonnell as well.

sleevi: A redirect can be to a relative URI, at which point, the HTTP client needs to resolve that relative URI to the document URI.

The term final value is documented in https://tools.ietf.org/html/rfc7231#section-7.1.2 which is referred to by these status codes. It probably didn’t show up in a Find-in-Page for you because of the white space of the RFC, which tripped me up as well in the past.

cardonator: Thanks for recognizing my lack of searching skill, lol! That does help clarify the meaning.

Issue #242, opened 2021-01-26
BRs: Clarify the CAA requirements for CAA and .onion addresses


sleevi:
As raised in https://archive.cabforum.org/pipermail/servercert-wg/2021-January/002433.html , Section 3.2.2.4 and Section 3.2.2.8 of the BRs have a bit of an awkward intersection.

3.2.2.8 requires checking CAA, which uses DNS, but `.onion` is set aside as a special-purpose domain by [RFC 7686](https://tools.ietf.org/html/rfc7686). RFC 7686 requires that DNS servers NXDOMAIN requests for `.onion`, which means that DNSSEC validation will fail (no appropriate records), and thus CAs may still issue provided they meet the requirements. However, by requiring CAs to query the CAA record multiple times violates the SHOULD NOT of [RFC 7686, Section 2](https://tools.ietf.org/html/rfc7686#section-2), "Application Software", and so it's rather silly.

dzacharo:
Gets even more weird if you follow the definition of the Internal Name which seems to apply for .onion addresses.

_**Internal Name:** A string of characters (not an IP address) in a Common Name or Subject Alternative Name field of a Certificate that cannot be verified as globally unique within the public DNS at the time of certificate issuance because it does not end with a Top Level Domain registered in IANA's Root Zone Database._

sleevi:
As mentioned on today's call though, 3.2.2.4 states:

> The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qualified
Domain Name (FQDN) listed in the Certificate as follows:
>   1. When the FQDN does not contain “onion” as the rightmost label, the CA SHALL validate the
FQDN using at least one of the methods listed below; and
>   2. When the FQDN contains “onion” as the rightmost label, the CA SHALL validate the FQDN in
accordance with Appendix B.

And Appendix B states
> 3. When a Certificate includes an FQDN where “onion” is in the right-most label of the Domain
Name, the Domain Name shall not be considered an Internal Name provided that the
Certificate was issued in compliance with this Appendix B.

CBonnell:
SC54 added the following to BR 3.2.2.8:

> As part of the Certificate issuance process, the CA MUST retrieve and process CAA
records in accordance with RFC 8659 for each dNSName in the subjectAltName
extension that does not contain an Onion Domain Name. I

Given this, this issue can resolved.

Issue #241, opened 2021-01-25
BRs: Update section reference of Agreed-upon change to website validate


dzacharo:
APPENDIX B 2. a should refer to 3.2.2.4.18 "Agreed‐Upon Change to Website v2" instead of 3.2.2.4.6 "Agreed‐Upon Change to Website".

sleevi:
Good spot! Yes, this should be updated; Appendix B left in a provision to handle "new" methods (3.2.2.4.18 / 3.2.2.4.19), but we forgot to explicitly do that in [SC25](https://cabforum.org/2020/02/01/ballot-sc25-define-new-http-domain-validation-methods-v2/)

There are a few other issues like this where it's "normative, but hopefully matches what everyone understands" changes (Specifically, #190 and #191 come to mind), and we can probably revisit #169 and #170 around that timeframe as well. I'm thinking as a spring cleanup ballot, or we could pursue onion cleanups-and-clarifications separately if needed.

Issue #240, opened 2021-01-18
BRs: Fix language in APPENDIX B 


dzacharo:
In APPENDIX B replace:

"The CA MAY verify the Applicant's control over the .onion service by having the Applicant provide a Certificate Request signed using the .onion public key"

with

"The CA MAY verify the Applicant's control over the .onion service by having the Applicant provide a Certificate Request signed using the .onion **private** key"

CBonnell:
Fixed in BR Appendix B with SC54:
>The CA MAY verify the Applicant’s control over the .onion service by having
the Applicant provide a Certificate Request signed using the .onion service’s
private key

PR #269, opened 2021-04-23
SC45: Wildcard Domain Validation

sleevi:
## Purpose of Ballot

This Ballot addresses security issues with the use of methods 3.2.2.4.6, 3.2.2.4.18, and 3.2.2.4.19 of the Baseline Requirements to authenticate an entire domain namespace. These methods rely on an HTTP based demonstration of control, which only demonstrates control over a particular host and service, rather than the entire Domain Namespace.

Effective 2021-12-01, these methods MUST NOT be used to issue Wildcard Certificates and MUST NOT be used as Authorization Domain Names for subordinate FQDNs of the validated FQDN.

The following motion has been proposed by Ryan Sleevi of Google and endorsed by Jos Purvis of Cisco and Dimitris Zacharopoulos of HARICA.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -158,6 +158,7 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
 | 2020-09-30 | 7.1.4.1 | Subject and Issuer Names for all possible certification paths MUST be byte-for-byte identical. |
 | 2020-09-30 | 7.1.6.4 | Subscriber Certificates MUST include a CA/Browser Form Reserved Policy Identifier in the Certificate Policies extension. |
 | 2020-09-30 | 7.2 and 7.3 | All OCSP and CRL responses for Subordinate CA Certificates MUST include a meaningful reason code. |
+| 2021-12-01 | 3.2.2.4 | CAs MUST NOT use methods 3.2.2.4.6, 3.2.2.4.18, or 3.2.2.4.19 to issue wildcard certificates or with Authorization Domain Names other than the FQDN. |
 
 ## 1.3 PKI Participants
 
@@ -471,7 +472,7 @@ The script outputs:
 | DBA | Doing Business As |
 | DNS | Domain Name System |
 | FIPS | (US Government) Federal Information Processing Standard |
-| FQDN | Fully Qualified Domain Name |
+| FQDN | Fully-Qualified Domain Name |
 | IM | Instant Messaging |
 | IANA | Internet Assigned Numbers Authority |
 | ICANN | Internet Corporation for Assigned Names and Numbers |
@@ -712,8 +713,8 @@ If a Random Value is used, the CA SHALL provide a Random Value unique to the Cer
 CAs SHALL NOT perform validations using this method after June 3, 2020. CAs MAY continue to re-use information and validations for domains validated under this method per the applicable certificate data reuse periods.
 
 **Note**:
-  * For certificates issued prior to 2021-10-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
-  * For certificates issued on or after 2021-10-01, the CA MUST NOT issue Certificates for other FQDNs that end with all labels of the validated FQDN. This method is NOT suitable for validating Wildcard Domain Names, except as permitted by [Appendix B](#appendix-b--issuance-of-certificates-for-onion-domain-names) when validating a FQDN where "onion" is the right-most label of the Domain Name.
+  * For certificates issued prior to 2021-12-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+  * For certificates issued on or after 2021-12-01, the CA MUST NOT issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
 
 ##### 3.2.2.4.7 DNS Change
 
@@ -730,7 +731,7 @@ If a Random Value is used, the CA SHALL provide a Random Value unique to the Cer
 
 Confirming the Applicant's control over the FQDN by confirming that the Applicant controls an IP address returned from a DNS lookup for A or AAAA records for the FQDN in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
 
-**Note**: Once the FQDN has been validated using this method, the CA MUST NOT also issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
+**Note**: Once the FQDN has been validated using this method, the CA MUST NOT issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
 
 ##### 3.2.2.4.9 Test Certificate
 
@@ -833,8 +834,8 @@ If a Random Value is used, then:
 2. The Random Value MUST remain valid for use in a confirming response for no more than 30 days from its creation. The CPS MAY specify a shorter validity period for Random Values, in which case the CA MUST follow its CPS.
 
 **Note**:
-  * For certificates issued prior to 2021-10-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
-  * For certificates issued on or after 2021-10-01, the CA MUST NOT issue Certificates for other FQDNs that end with all labels of the validated FQDN. This method is NOT suitable for validating Wildcard Domain Names, except as permitted by [Appendix B](#appendix-b--issuance-of-certificates-for-onion-domain-names) when validating a FQDN where "onion" is the right-most label of the Domain Name.
+  * For certificates issued prior to 2021-12-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+  * For certificates issued on or after 2021-12-01, the CA MUST NOT issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
 
 ##### 3.2.2.4.19 Agreed-Upon Change to Website - ACME
 
@@ -852,8 +853,8 @@ If the CA follows redirects:
 4. Redirects MUST be to resource URLs accessed via Authorized Ports.
 
 **Note**:
-  * For certificates issued prior to 2021-10-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
-  * For certificates issued on or after 2021-10-01, the CA MUST NOT issue Certificates for other FQDNs that end with all labels of the validated FQDN. This method is NOT suitable for validating Wildcard Domain Names, except as permitted by [Appendix B](#appendix-b--issuance-of-certificates-for-onion-domain-names) when validating a FQDN where "onion" is the right-most label of the Domain Name.
+  * For certificates issued prior to 2021-12-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+  * For certificates issued on or after 2021-12-01, the CA MUST NOT issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
 
 ##### 3.2.2.4.20 TLS Using ALPN
 
@@ -861,7 +862,7 @@ Confirming the Applicant's control over a FQDN by validating domain control of t
 
 The token (as defined in RFC 8737, Section 3) MUST NOT be used for more than 30 days from its creation. The CPS MAY specify a shorter validity period for the token, in which case the CA MUST follow its CPS.
 
-**Note**: Once the FQDN has been validated using this method, the CA MUST NOT also issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
+**Note**: Once the FQDN has been validated using this method, the CA MUST NOT issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
 
 #### 3.2.2.5 Authentication for an IP Address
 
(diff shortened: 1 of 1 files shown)


sleevi:
@castillar @dzacharo Could you review this draft ballot and text to confirm before voting?

From the off-list discussion with Dimitris, there was some concern about permitting wildcard `.onion` domain names. I've flagged this as TBD, because we identified that the issues affecting these methods for subdomains equally affect `.onion` domains, and unlike DNS, Tor Hidden Service operators do not have the ability to use mechanisms like CAA to further restrict or prevent accidental authorization. If we decide to keep `.onion` consistent with DNS, which does make a degree of sense, then there are slight modifications to Appendix B to be made.

The choice to placing this exception within Appendix B is that Section 3.2.2.4 explicitly "branches" to Appendix B for validation, since `.onion` does not use DNS at all. Attempting to shoehorn these into .6, .18, and .19 felt more awkward.

sleevi:
See also #270 . I'm not sure if we want to tackle that in this, since it's "new", and I prefer to keep things more narrowly scoped. This is why I only tackled #241 but not #242 or #240, as those seem more appropriate for an overall `.onion` cleanup. I mention it here, however, in case it's dispositive towards the TBD I'm flagging here.

sleevi:
Oh, and CC'ing @CBonnell since he expressed interest in seeing the draft language and proposal on the 2021-04-22 Validation WG call.

dzacharo:
While reading the introduction of the ballot, I had to read the part related to "validation for domains" several times to understand what the intention was. 

That's because "validation of a domain" in the current version of 3.2.2.4.18, say `foo.bar.example.com`, with the use of ADNs, can be performed by choosing an ADN `foo.bar.example.com`, `bar.example.com` or `example.com`. Now that we plan to remove wildcards for these methods, the notion of ADN does not seem to apply anymore. These methods will only be able to validate the requested FQDNs to be included in the Certificates. The only exception where the notion of ADN is applicable, is for `.onion` domains, where one can use the hidden service Domain Name (two labels) as an ADN to validate any FQDN that ends with all the labels of the hidden service Domain Name.

Suggest replacing: "Authorization Domain Name" with "FQDN" in 3.2.2.4.18.

I also suggest a small change in the introduction reusing the "subordinate FQDNs" language, clarifying the ADN/FQDN language a bit, and pointing to the Appendix 2b validation method:

"One exception to this prohibition is the issuance of .onion certificates, as specified by Appendix B. For `.onion` domains, provided that the validated ADN and FQDN is the exact hidden service (i.e. two domain labels), this method MAY be used to validate a Wildcard Domain Name and MAY be used to authorize subsequent domain names. However, validation for domains such as `foo.<hidden service>.onion` MUST NOT be used to validate domains such as `bar.foo.<hidden service>.onion` unless the validation demonstrates authorization by the Tor Hidden Service's long term master key having signed a CSR." 

with 

"One exception to this prohibition is the issuance of .onion certificates, as specified by Appendix B. For `.onion` domains, if the validated FQDN is the exact Hidden Service (i.e. two domain labels), this method MAY be used to validate a Wildcard Domain Name and MAY be used to authorize subordinate FQDNs. However, validation for domains such as ` foo.<hidden service>.onion` MUST NOT be used to validate domains such as `bar.foo.<hidden service>.onion` unless the Applicant uses the validation process described in Appendix B - paragraph 2b, that demonstrates authorization by the Tor Hidden Service's long term master key having signed a CSR."

I also don't see a reason not to address #240 which seems uncontroversial.

Finally, since method 6 is deprecated since 2020-06-03, it seems weird to add a Note with a restriction date for "things issued after 2021-12-01". I guess it doesn't hurt to add for consistency but it still remains weird :)

I'm sorry these didn't come up earlier, I didn't have time to do a good review in the past weeks. I hope these are helpful.

sleevi:
> Now that we plan to remove wildcards for these methods, the notion of ADN
> does not seem to apply anymore.
>
Sorry, I should have explained more context that I considered this, and
rejected it, as part of developing this ballot. The issue here is that
because we allow reuse, I wanted to minimize the changes to this method as
much as possible, to avoid the need to introduce a new method and sunset
the old. Further, with the fact that there’s a date transition, there is
not a natural way to make the change you propose until *after* the
effective date.

I agree that it’s a potential for the future, but I do not believe there is
a clean way of making that change before December, which is why I opted for
this path. Changing to FQDN now would make it effective immediately, which
we know is not the goal.

I also suggest a small change in the introduction reusing the "subordinate
> FQDNs" language, clarifying the ADN/FQDN language a bit, and pointing to
> the Appendix 2b validation method:
>
As I mentioned privately over email, I’m still fairly uncomfortable
allowing this for .onion, because there is not a solid security reason to
justify allowing it. I think it may make sense to hold off on this language
until you can clarify if keeping wildcard for onion for these methods is
absolutely essential for your support, given that it doesn’t really provide
the same assurance.

That’s why this is still flagged TBD: both to provide an opportunity for
you to reconsider, and provide transparency for others to consider the
asymmetry here (and perhaps convince you we can drop the Appendix B bits 🙂)

We’re still discussing internally here whether to keep it, or whether to
remove it and potentially seek another endorser if that is a deal breaker
for you. But the best first step is providing fully drafted language to
discuss and explore how it would work, while we and others consider if it
should in the first place :)

Finally, since method 6 is deprecated since 2020-06-03, it seems weird to
> add a Note with a restriction date for "things issued after 2021-12-01". I
> guess it doesn't hurt to add for consistency but it still remains weird :)
>
This was drafted before SC42v2 was finished voting. If SC42 fails/failed,
then the reuse period would have been 825 days from 2020-06-03, and that’s
what this addressed. My plan is to include this language as part of the
considerations of other ballots (I.e. there are two versions of this
ballot, one with this language in case SC42v2 fails IP review, one without
this language if it succeeds), since I hope to get this voted on before IP
Review is completed for that.

dzacharo:
> As I mentioned privately over email, I’m still fairly uncomfortable allowing this for .onion, because there is not a solid security reason to justify allowing it. I think it may make sense to hold off on this language until you can clarify if keeping wildcard for onion for these methods is absolutely essential for your support, given that it doesn’t really provide the same assurance. That’s why this is still flagged TBD: both to provide an opportunity for you to reconsider, and provide transparency for others to consider the asymmetry here (and perhaps convince you we can drop the Appendix B bits 🙂) We’re still discussing internally here whether to keep it, or whether to remove it and potentially seek another endorser if that is a deal breaker for you. But the best first step is providing fully drafted language to discuss and explore how it would work, while we and others consider if it should in the first place :)

I was under the impression we agreed privately on the intent for wildcard validation using methods 18/19, and the introduction of the ballot suggests that validating the hidden service domain name exactly (two labels, `<hidden service>.onion`) with methods 18/19 would be sufficient to validate wildcards and subdomains. Did I misread? 

To recap:
- If you're validating exactly label.onion, wildcards and subdomains are OK
- If you're validating foo.label.onion, wildcards and subdomains are NOT OK
- If you're validating via Tor Hidden Service CSR (Appendix B - 2b), wildcards and subdomains are OK

HARICA supports these expectations as reasonably secure for validating onion domains. Do you consider that the current proposal defines these expectations adequately or do we need more language tweaking? We would still like to endorse this ballot and had the impression we are in agreement about the intent of the introduced language for onion domains.

I also agree with your rationale why we should not change the ADN/FQDN language of 3.2.2.4.18 and Appendix B. Same thing with the effective dates, it wasn't a strong objection anyway. The clarification I suggested was for the introduction of the ballot which added a pointer to the proper section of the BRs (Appendix B - 2b) in addition to the "authorization by the Tor Hidden Service's long term master key having signed a CSR" which sounds "intimidating" for those reviewing the ballot who are unfamiliar with this terminology :-)

dzacharo:
I forgot to add that we would still endorse even if we only allowed wildcard .onion domain validation using the Appendix B - 2b method.

sleevi:
> I was under the impression we agreed privately on the intent for wildcard validation using methods 18/19, and the introduction of the ballot suggests that validating the hidden service domain name exactly (two labels, `<hidden service>.onion`) with methods 18/19 would be sufficient to validate wildcards and subdomains. Did I misread?
> 
> To recap:
> 
> * If you're validating exactly label.onion, wildcards and subdomains are OK
> * If you're validating foo.label.onion, wildcards and subdomains are NOT OK
> * If you're validating via Tor Hidden Service CSR (Appendix B - 2b), wildcards and subdomains are OK

Right, I thought we'd agreed that validating `foo.label.onion` and issuing subdomains was NOT OK, but that you were still in favor of validating `label.onion` and subdomains being OK, while we at Google were still working through it.

If you're ok with `label.onion` being NOT OK; that is, relying only on 2.b for wildcards, treating the CSR method as effectively the same level of control over the entire DNS namespace, then we can remove the TBD and simplify this further :)

> I also agree with your rationale why we should not change the ADN/FQDN language of 3.2.2.4.18 and Appendix B. Same thing with the effective dates, it wasn't a strong objection anyway. The clarification I suggested was for the introduction of the ballot which added a pointer to the proper section of the BRs (Appendix B - 2b) in addition to the "authorization by the Tor Hidden Service's long term master key having signed a CSR" which sounds "intimidating" for those reviewing the ballot who are unfamiliar with this terminology :-)

Gotcha. If I'm sure I'm not misunderstanding, and can simplify, then yeah, I can reword the introduction to tie those concepts together better, and reference the existing language.

docs/BR.md
```diff
@@ -711,7 +712,9 @@ If a Random Value is used, the CA SHALL provide a Random Value unique to the Cer
 
 CAs SHALL NOT perform validations using this method after June 3, 2020. CAs MAY continue to re-use information and validations for domains validated under this method per the applicable certificate data reuse periods.
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+**Note**:
+  * For certificates issued prior to 2021-12-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+  * For certificates issued on or after 2021-12-01, the CA MUST NOT issue Certificates for other FQDNs that end with all the labels of the validated FQDN unless the CA performs a separate validation for that FQDN using an authorized method. This method is NOT suitable for validating Wildcard Domain Names.
```
robstradling, line 717: "that FQDN" seems to read as though it refers to "the validated FQDN", which doesn't make sense.

Suggestion: s/a separate validation for that FQDN using an authorized method/separate validations for each of those other FQDNs using authorized methods/

sleevi: So, this is for better or worse the existing language (3.2.2.4.8, 3.2.2.4.20), so I think it might be better to keep symmetry/consistency here, and perhaps clarify this language as follow-up. Do you want to file an issue for that so we can sweep it up in a spring cleanup?

robstradling: Makes sense.  I've filed Issue #272 for this.

docs/BR.md
```diff
@@ -711,7 +712,9 @@ If a Random Value is used, the CA SHALL provide a Random Value unique to the Cer
 
 CAs SHALL NOT perform validations using this method after June 3, 2020. CAs MAY continue to re-use information and validations for domains validated under this method per the applicable certificate data reuse periods.
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+**Note**:
```
CBonnell, line 715: Do we need to update this method at all? SC42 passed, so it will not be possible to reuse any method 6 validations starting October 1st, which is before the effective date of this ballot (SC45).

docs/BR.md
```diff
@@ -830,7 +833,9 @@ If a Random Value is used, then:
 1. The CA MUST provide a Random Value unique to the certificate request.
 2. The Random Value MUST remain valid for use in a confirming response for no more than 30 days from its creation. The CPS MAY specify a shorter validity period for Random Values, in which case the CA MUST follow its CPS.
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+**Note**:
+  * For certificates issued prior to 2021-12-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
```
CBonnell, line 837:
```suggestion
  * For Certificates issued prior to 2021-12-01, the CA MAY also issue Certificates for other FQDNs that end with all the labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
```

docs/BR.md
```diff
@@ -2567,7 +2574,23 @@ This appendix defines permissible verification procedures for including one or m
...
+
+      ii. [Section 3.2.2.4.18 - Agreed-Upon Change to Website v2](#322418-agreed-upon-change-to-website-v2):
+
+          The CA SHALL only use this method to validate Wildcard Certificates or as an Authorization Domain Name as specified by that section, except as follows:
+
+          When the CA performs the validation with the FQDN containing exactly two labels, the hidden service name and "onion", the CA MAY use this validation to validate a Wildcard Domain Name for this FQDN and MAY use this validation to issue Certificates for other FQDNs that end with all the labels, without regard to the prohibition of such certificates in [Section 3.2.2.4.18](#322418-agreed-upon-change-to-website-v2).
+
+      iii. [Section 3.2.2.4.19 - Agreed-Upon Change to Website - ACME](#322419-agreed-upon-change-to-website-acme):
+
+           The CA SHALL only use this method to validate Wildcard Certificates or as an Authorization Domain Name as specified by that section, except as follows:
```
CBonnell, line 2591:
```suggestion
          Validations for .onion Domain Names SHALL be performed according to [Section 3.2.2.4.19](#322419-agreed-upon-change-to-website-acme). If the method prohibits its use for the validation of Wildcard Domain Names for the .onion Domain Name or other FQDNs that end with the .onion Domain Name and the .onion Domain Name consists of exactly two labels (the hidden service name and "onion"), then the following exception applies:
```

docs/BR.md
```diff
@@ -2567,7 +2574,23 @@ This appendix defines permissible verification procedures for including one or m
...
+
+           When the CA performs the validation with the FQDN containing exactly two labels, the hidden service name and "onion", the CA MAY use this validation to validate a Wildcard Domain Name for this FQDN and MAY use this validation to issue Certificates for other FQDNs that end with all the labels, without regard to the prohibition of such certificates in [Section 3.2.2.4.19](#322419-agreed-upon-change-to-website-acme).
```
CBonnell, line 2593:
```suggestion
          The CA MAY use validations performed using this method to validate a Wildcard Domain Name for the validated .onion Domain Name and MAY use validations performed using this method to validate other FQDNs that end with all labels of the validated .onion Domain Name.
```

CBonnell (changes requested):

docs/BR.md
sleevi: Technically SC42 hasn't passed the IP review period.

It's a little tricky to do the "two versions", but my approach with other ballots has been to "include the language as if it fails" as one commit, and then put another comment "and here's the language if it succeeds", and use two different commit ranges for the ballot.

So yeah, if SC42 succeeds, I think the updated version (and I need to double check the math, but I'm sure you're right) would simply be to "must not be used" this portion because of the SC42 interaction (i.e. we could have done it in SC42 itself, but doing it now will at least ensure there's no ambiguity about wildcards from folks who don't implement/notice SC42)

docs/BR.md
```diff
@@ -2567,7 +2574,23 @@ This appendix defines permissible verification procedures for including one or m
...
+   a. The CA MAY verify the Applicant's control over the .onion service by using one of the following methods from [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control):
+
+      i. [Section 3.2.2.4.6 - Agreed-Upon Change to Website](#32246-agreed-upon-change-to-website):
+
+         The CA SHALL only use this method when reusing a previously completed domain validation permitted by that method and [Section 4.2.1](#421-performing-identification-and-authentication-functions). The CA MUST NOT perform new validations using this method.
+
+      ii. [Section 3.2.2.4.18 - Agreed-Upon Change to Website v2](#322418-agreed-upon-change-to-website-v2):
+
+          The CA SHALL only use this method to validate Wildcard Certificates or as an Authorization Domain Name as specified by that section, except as follows:
```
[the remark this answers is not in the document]
sleevi: Sorry, this was language I'm meaning to remove, but haven't had a chance to pick this up yet :(

docs/BR.md
```diff
@@ -2567,7 +2574,23 @@ This appendix defines permissible verification procedures for including one or m
...
+
+          When the CA performs the validation with the FQDN containing exactly two labels, the hidden service name and "onion", the CA MAY use this validation to validate a Wildcard Domain Name for this FQDN and MAY use this validation to issue Certificates for other FQDNs that end with all the labels, without regard to the prohibition of such certificates in [Section 3.2.2.4.18](#322418-agreed-upon-change-to-website-v2).
```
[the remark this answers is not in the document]
sleevi: So this is a substantial change in the language (although, as mentioned above, it's being removed)

The issue here is that your proposed language changes to allow `foo.bar.onion` to authenticate `*.foo.bar.onion`, and that's explicitly not something that is intended nor good, as covered on the list, the call, and the PR discussion.

But, like I said, this is being removed, so that wildcard authorization/subdomain authorization is only done with an equivalent "proof of control over entire domain namespace" (i.e. the onion hidden service key)

sleevi:
@dzacharo @castillar Could you take another look and see if you're happy endorsing.

This gets rid of the asymmetry (as discussed on list) addresses the feedback from @CBonnell about capitalization.

castillar:
Looks good to me!

Manjitsing:
Someone hacked my personal information without my permission someone hack my personal on my all personal data on my all Gmail all iCloud my Samsung my Google Drive my anything￼My name of the domain any mama you are a number of my IP address on my server address my host address for my anything hack my personal service Canada account my tax regarding an all information my banking information please help me protect me this happening March 28,2020 to go by the time I can face on that I have Panaway every day please help me protect me me protect me￼

Manjitsing:
Someone hacked my personal information without my permission someone hack my personal on my all personal data on my all Gmail all iCloud my Samsung my Google Drive my anything￼My name of the domain any mama you are a number of my IP address on my server address my host address for my anything hack my personal service Canada account my tax regarding an all information my banking information please help me protect me this happening March 28,2020 to go by the time I can face on that I have Panaway every day please help me protect me me protect me￼

Manjitsing:
Someone hacked my personal information without my permission someone hack my personal on my all personal data on my all Gmail all iCloud my Samsung my Google Drive my anything￼My name of the domain any mama you are a number of my IP address on my server address my host address for my anything hack my personal service Canada account my tax regarding an all information my banking information please help me protect me this happening March 28,2020 to go by the time I can face on that I have Panaway every day please help me protect me me protect me￼

Manjitsing:
Someone hacked my personal information without my permission someone hack my personal on my all personal data on my all Gmail all iCloud my Samsung my Google Drive my anything￼My name of the domain any mama you are a number of my IP address on my server address my host address for my anything hack my personal service Canada account my tax regarding an all information my banking information please help me protect me this happening March 28,2020 to go by the time I can face on that I have Panaway every day please help me protect me me protect me￼

PR #285, opened 2021-06-10
SC48 - Domain Name and IP Address Encoding

CBonnell:
This ballot clarifies several points that were left unanswered in the wake of the ballot 202 failure. The following items are proposed with an immediate effective date (but of course can be discussed):

* Prohibition on U-labels in the CN
* More clarity surrounding Reserved IP Addresses

For a future effective date (TBD):
* Prohibition on Reserved LDH domain labels that are not XN-labels
* Requirement that all XN-labels must be validated as containing valid Punycode

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -161,7 +161,7 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
 | 2020-09-30 | 7.1.6.4 | Subscriber Certificates MUST include a CA/Browser Form Reserved Policy Identifier in the Certificate Policies extension. |
 | 2020-09-30 | 7.2 and 7.3 | All OCSP and CRL responses for Subordinate CA Certificates MUST include a meaningful reason code. |
 | 2021-07-01 | 3.2.2.4.18 and 3.2.2.4.19 | Redirects MUST be the result of one of the HTTP status code responses defined. | 
-| 202x-yy-zz | 7.1.4.2.1 | Fully-Qualified Domain Names MUST consist solely of P-Labels and Non-Reserved LDH Labels. |
+| 2021-10-01 | 7.1.4.2.1 | Fully-Qualified Domain Names MUST consist solely of P-Labels and Non-Reserved LDH Labels. |
 
 ## 1.3 PKI Participants
 
@@ -2071,7 +2071,7 @@ __Contents:__ This extension MUST contain at least one entry. Each entry MUST be
 
    The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of LDH Labels joined together by a U+002E FULL STOP (".") character. Namely, entries MUST NOT contain underscore characters ("_").
 
-   Effective xxx 1, 202x, the Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name MUST consist solely of Domain Labels that are P-Labels or Non-Reserved LDH Labels.
+   Effective 2021-10-01, the Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name MUST consist solely of Domain Labels that are P-Labels or Non-Reserved LDH Labels.
 
 * `iPAddress`: The entry MUST contain an IPv4 or IPv6 address that the CA has validated in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address). The entry MUST NOT contain a Reserved IP Address.
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -161,7 +161,7 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
 | 2020-09-30 | 7.1.6.4 | Subscriber Certificates MUST include a CA/Browser Form Reserved Policy Identifier in the Certificate Policies extension. |
 | 2020-09-30 | 7.2 and 7.3 | All OCSP and CRL responses for Subordinate CA Certificates MUST include a meaningful reason code. |
 | 2021-07-01 | 3.2.2.4.18 and 3.2.2.4.19 | Redirects MUST be the result of one of the HTTP status code responses defined. | 
-| 202x-yy-zz | 7.1.4.2.1 | Fully-Qualified Domain Names MUST consist solely of P-Labels and Non-reserved LDH Labels. |
+| 202x-yy-zz | 7.1.4.2.1 | Fully-Qualified Domain Names MUST consist solely of P-Labels and Non-Reserved LDH Labels. |
 
 ## 1.3 PKI Participants
 
@@ -352,7 +352,7 @@ No stipulation.
 
 **Legal Entity**: An association, corporation, partnership, proprietorship, trust, government entity or other entity with legal standing in a country's legal system.
 
-**Non-reserved LDH Label**: From RFC 5890 (<http://tools.ietf.org/html/rfc5890>): "The set of valid LDH labels that do not have '--' in the third and fourth positions."
+**Non-Reserved LDH Label**: From RFC 5890 (<http://tools.ietf.org/html/rfc5890>): "The set of valid LDH labels that do not have '--' in the third and fourth positions."
 
 **Object Identifier**: A unique alphanumeric or numeric identifier registered under the International Organization for Standardization's applicable standard for a specific object or object class.
 
@@ -2067,13 +2067,13 @@ __Certificate Field:__ `extensions:subjectAltName`
 __Required/Optional:__ Required  
 __Contents:__ This extension MUST contain at least one entry. Each entry MUST be one of the following types:
 
-1. `dNSName`: The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT not contain an Internal Name.
+* `dNSName`: The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT not contain an Internal Name.
 
    The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be in the "preferred name syntax", as specified in RFC 5280. Namely, entries MUST NOT contain underscore characters ("_").
 
-   Effective xxx 1, 202x, the Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name MUST consist solely of Domain Labels that are P-Labels or Non-reserved LDH Labels.
+   Effective xxx 1, 202x, the Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name MUST consist solely of Domain Labels that are P-Labels or Non-Reserved LDH Labels.
 
-2. `iPAddress`: The entry MUST contain an IPv4 or IPv6 address that the CA has validated in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address). The entry MUST NOT contain a Reserved IP Address.
+* `iPAddress`: The entry MUST contain an IPv4 or IPv6 address that the CA has validated in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address). The entry MUST NOT contain a Reserved IP Address.
 
 ##### 7.1.4.2.2 Subject Distinguished Name Fields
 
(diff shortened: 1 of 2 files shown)


docs/BR.md
```diff
@@ -2043,19 +2065,25 @@ Subject attributes MUST NOT contain only metadata such as '.', '-', and ' ' (i.e
...
 __Required/Optional:__ Required  
-__Contents:__ This extension MUST contain at least one entry. Each entry MUST be either a `dNSName` containing the Fully-Qualified Domain Name or an `iPAddress` containing the IP address of a server. The CA MUST confirm that the Applicant controls the Fully-Qualified Domain Name or IP address or has been granted the right to use it by the Domain Name Registrant or IP address assignee, as appropriate.
+__Contents:__ This extension MUST contain at least one entry. Each entry MUST be one of the following types:
+
+* `dNSName`: The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT not contain an Internal Name.
 
-Wildcard FQDNs are permitted.
+   The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be in the "preferred name syntax", as specified in RFC 5280. Namely, entries MUST NOT contain underscore characters ("_").
```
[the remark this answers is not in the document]
CBonnell: Thanks, I agree using "LDH Labels" throughout is clearer and more concise.

CBonnell:
> So, I think you've hit all the right notes, and we'd be supportive of this.

Thanks for the review and feedback!

> I think one question I have is: should we fold this into the profiles work? I think that should be ready to go shortly (outstanding items are SRVName constraints, fixing up the backdating, and any discussion on validity periods that may get proposed)

I think we're looking to push this separately from the profiles ballot, for two reasons:
1. The concepts introduced here all should be familiar to the Forum since they were included in Ballot 202. It's essentially a "self-contained" ballot that CAs can assess independently from the larger profiles work.
2. Proposing small improvements such as this independently from the profiles work will make evaluation of the profiles less burdensome when it is ready for wider discussion/review, as there will be less normative changes introduced.

sleevi:
> 1. The concepts introduced here all should be familiar to the Forum since they were included in Ballot 202. It's essentially a "self-contained" ballot that CAs can assess independently from the larger profiles work.
> 2. Proposing small improvements such as this independently from the profiles work will make evaluation of the profiles less burdensome when it is ready for wider discussion/review, as there will be less normative changes introduced.

But I think both these points are true for most of the profiles work, right?

Re: 2, maybe I'm miscalibrated - I've been trying to circulate the profiles work for feedback for a while for that wider discussion/review. Because of that, I was thinking the review period would only be a week - but it sounds like you're thinking the review period may be longer (i.e. that CAs have not yet started reviewing that work), and that's why this would benefit clearing sooner / being separable?

I'm not wanting to insist this be part of profiles, but mostly trying to think through the sequencing of how multiple ballots in flight will be managed, given the overlap here in terms of sections touched. It makes it possible that I'll get it wrong (particularly, the handling of the "if this ballot succeeds" / "if this ballot fails" variants), so, selfishly, it's a little easier if integrated. It's the same reason I haven't (yet) pushed a spring (haha, summer) cleanup ballot yet - same risk.

What's your thought on the timing of this ballot? That is, I realize there's an open question for what the effective date should be for the new changes, and I wasn't sure when you see that effective date being resolved (because it means that time, +6 weeks, will have to deal with the conflict in the profiles work)

dzacharo:
@CBonnell, I reviewed the ballot last week and we'd be happy to endorse.

dzacharo (approved):

sleevi (approved):

dzacharo (approved):

sleevi (approved):

docs/BR.md
```diff
@@ -931,9 +952,10 @@ Confirming the Applicant's control over the IP Address by performing the procedu
...
+Before issuing a Wildcard Certificate, the CA MUST establish and follow a documented procedure that determines if the FQDN portion of any
+Wildcard Domain Name in the Certificate is “registry-controlled” or is a “public suffix” (e.g. “\*.com”, “\*.co.uk”, see RFC 6454 Section 8.2 for further explanation).
 
-If a wildcard would fall within the label immediately to the left of a registry-controlled or public suffix, CAs MUST refuse issuance unless the applicant proves its rightful control of the entire Domain Namespace. (e.g. CAs MUST NOT issue "\*.co.uk" or "\*.local", but MAY issue "\*.example.com" to Example Co.).
+If the FQDN portion of any Wildcard Domain Name is "registry-controlled" or is a "public suffix", CAs MUST refuse issuance unless the Applicant proves its rightful control of the entire Domain Namespace. (e.g. CAs MUST NOT issue "\*.co.uk" or "\*.local", but MAY issue "\*.example.com" to Example Co.).
```
aarongable, line 958: Minor nit: there is an inconsistency here between smart quotes (line 956: `“registry-controlled”`) and straight quotes (line 958: `"registry-controlled"`).

CBonnell: Unless others disagree, it appears that the BRs use both types of quoting (straight vs. smart) so I think we should decide on which type of quotes we want in the documents and then fix all occurrences at once. Given that the documents are already inconsistent and mixed usage doesn't impair readability, I don't think we should delay the ballot further to address instance of mixed use.

sleevi: Yeah, we can leave these as-is for this ballot. Pandoc will automatically convert these to ligatures for the PDF rendering, so it makes no difference.

aarongable: Agreed, no need to delay this ballot to fix it. I only brought it up because both of these (mismatched) quotes were introduced in this ballot, not just a simple matter of this ballot using one style adjacent to where the other style already existed.

docs/BR.md
```diff
@@ -2051,19 +2073,25 @@ Subject attributes MUST NOT contain only metadata such as '.', '-', and ' ' (i.e
...
+   __Contents:__ If present, this field MUST contain exactly one entry that is one of the values contained in the Certificate's `subjectAltName` extension (see [Section 7.1.4.2.1](#71421-subject-alternative-name-extension)). The value of the field MUST be encoded as follows:
+
+   * If the value is an IPv4 address, then the value MUST be encoded as an IPv4Address as specified in RFC 3986, Section 3.2.2.
+   * If the value is an IPv6 address, then the value MUST be encoded in the text representation specified in RFC 5952, Section 4.
+   * If the value is a Fully-Qualified Domain Name or Wildcard Domain Name, then the value MUST be encoded as a character-for-character copy of the `dNSName` entry value from the `subjectAltName` extension. Specifically, all Domain Labels of the Fully-Qualified Domain Name or FQDN portion of the Wildcard Domain Name must be encoded as LDH Labels, and P-Labels MUST NOT be converted to their Unicode representation.
```
robplee, line 2094: Hi all, can I just ask for a clarification on this line.  Does this mean that the commonName field must be _exactly_ found in a SAN dNSName entry or would it be reasonable for the common name to be EXAMPLE.COM and the SAN dNSName to be example.com?  DNS names are not case-sensitive so it shouldn't matter but I just want to clarify what precisely is meant by a character-for-character copy?

CBonnell: Thanks for raising this. I agree this isn't clear and I should have worded this better.
As you mentioned, DNS is case-insensitive so differences in casing aren't material. Given this, I think it's reasonable to interpret this as a "case-insensitive character-for-character copy". Nonetheless, we should make this explicit in the next "Spring cleanup" ballot.

dougbeattie: Yes, I agree, we should interpret as @CBonnell recommends and fix in a clean up ballot.

sleevi: I'm not sure I agree with Corey here with how we should interpret it. A character in an IA5String is an unambiguous value from the ASCII/IA5 character table, and an 'a' character is not equivalent to an 'A' character.

I don't believe our current language permits this case differential. I understand Corey reasonably suggesting we can consider changing that, but I don't believe that's currently what the text reads.

CBonnell: The intent of my writing "character-for-character copy" was to prohibit transformation of P-labels in a SAN dNSName to U-labels in the CN. The next sentence, "Specifically, " clarifies that is the intent. Even if we did mandate a case-sensitive match, I'm not sure what goal that would accomplish from a practical standpoint.

sleevi: The element of "character-for-character" is still relevant, because a CN is not an IA5String, but a DirectoryString.  If you're using the allowances in RFC 4.1.2.6 ( https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.6 ), then while the PrintableString and UTF8String encodings for the characters remain the same as IA5String, different encodings are used if you're using BMPString/UniversalString.

The existing BRs have the language:
> If present, this field MUST contain a single IP address or Fully‐Qualified Domain Name that is one of the values contained in the Certificate’s subjectAltName extension 

And, just like eliminating ambiguity about whether a U-Label is "contained" in a P-label-expressed `dNSName`, I thought the explicit goal here was to ensure a "byte for byte" copy, while accounting for encoding - thus, "character for character". This was certainly the understanding when endorsing, because I'm not sure how we can say "a" is a "character by character copy". It's considered equivalent under DNS matching, yes, but that is not a copy. Allowing 'a' for 'A' is functionally no different than allowing a U-label to be a substitution of the A-label, by arguing that it is a "character by character" copy, simply expressed as a U-label.

There is no legitimate need to express the CN in a separate case syntax than a SAN, and the mere existence of such a permutation is indicative of a point of concern.

CBonnell: > Allowing 'a' for 'A' is functionally no different than allowing a U-label to be a substitution of the A-label, by arguing that it is a "character by character" copy, simply expressed as a U-label.

A U-label will by definition not be a LDH label. So the P-label to U-label conversion case is prohibited by this language.

> There is no legitimate need to express the CN in a separate case syntax than a SAN, and the mere existence of such a permutation is indicative of a point of concern.

I think I agree it seems "cleaner" from an engineering purity standpoint that they are binary equal, but I'm not sure of the practical benefit of mandating. Is there a concern about confusable characters or something similar that such case mismatches would cause?

sleevi: > So the P-label to U-label conversion case is prohibited by this language.

Just to make sure we're on the same page: I *believe* we're saying that, even if we drop the "Specifically" language, the requirement still exists that P-label to U-Label conversions are prohibited. I'm saying that with, or without, the "Specifically" language, case conversions are *similarly* prevented, on the same principle: they are character transformations, and thus *not* character-by-character copies.

> Is there a concern about confusable characters or something similar that such case mismatches would cause?

You're going for the "why not" (default allow), I'm going from the "why" (default deny). There is no legitimate need for this flexibility, and the existence of this flexibility can easily lead to implementation bugs by the CA. That is, if the value of the CN and SAN are both populated by the CA by their systems, then there's no reason to ever expect a case mismatch. The only time one should expect a case mismatch is if one, or both, fields come from user-supplied or manually-entered data. While it's _possible_ that the CA performed the necessary case-folding comparison to ensure this requirement is met, it is unfortunately both possible and demonstrably likely that the CA merely transposed one or both values from user-supplied data. Such transposition can easily lead to issues (e.g. if a CA mistakenly uses `str{n}casecmp` to compare the CN and SAN, while failing to account for the possibility of embedded `NUL`s in the ASN.1 string).

This is why I say that this flexibility is intrinsically bad: the only time it's useful is for when a CA is doing something that is known error prone. I don't believe the current language permits that badness, on the same basis that just because certain Unicode characters may compare equivalent does not mean that they are identical characters/codepoints (which is why we require Unicode normalization, by way of RFC 5280).

CBonnell: > Just to make sure we're on the same page: I believe we're saying that, even if we drop the "Specifically" language, the requirement still exists that P-label to U-Label conversions are prohibited. I'm saying that with, or without, the "Specifically" language, case conversions are similarly prevented, on the same principle: they are character transformations, and thus not character-by-character copies.

I agree that as presently worded, it can be read that case mismatches are prohibited. However, my intent wasn't to prohibit that and was merely to prohibit conversion to U-labels.

To be clear: are you asserting that the Chrome Root Program views such case mismatches as mis-issuance when the ballot clears IPR?

> There is no legitimate need for this flexibility, and the existence of this flexibility can easily lead to implementation bugs by the CA

I agree that such case mismatches are a code smell, absent someone providing a compelling reason that such mismatches are valuable. That being said, this prohibition would affect many CAs with short notice. If that were the intent, I would have proposed a future effective date to give time to transition, especially since this wasn't a previously stated expectation by Root Programs.

sleevi: > To be clear: are you asserting that the Chrome Root Program views such case mismatches as mis-issuance when the ballot clears IPR?

Yes, we will view this as misissuance, especially since we felt the (new) wording was unambiguous in its expectations, even if that may not have been what you personally intended.

I think the best path forward here would be to discuss with the broader SCWG, particularly to solicit feedback from other root programs for their perspectives.

As I mentioned, while you view this as a new requirement, this seems to be a clarification of the existing requirement. While I can understand your intent may have been different, I'm also not sure how there can be any supported interpretation other than what I mentioned. In the broader conversation, it may be useful to explain how to support the case-insensitive interpretation (not just because of intent), as that will help better inform our handling of any misissuances by CAs.

We've been here before with other ballots, so I don't want to seem inflexible: we handle misissuances and failures to revoke based on the information available to us, so if, for example, a CA fails to adopt the expected action, because of a legitimate misinterpretation of the requirement, it's useful for us to understand how the misinterpretation came to be, so we can both correct it and prevent such issues in the future.

mike-reilly: Ryan, I assume you would potentially be looking at certs issued after the ballot passes IPR review as mis-issued?  Or would consider all certificates mis-issued regardless of issue date?

sleevi: @mike-reilly Effective date (i.e [~3 weeks from now](https://cabforum.org/2021/07/22/ballot-sc48-v2-domain-name-and-ip-address-encoding/))

mike-reilly: Understand.  Certs would be considered mis-issued by Google Chrome after effective date (about 3 weeks from now).  Thanks, Mike

mike-reilly: To clarify......only certs issued **AFTER** the IPR review ends would be potentially be considered mis-issued by Google Chrome correct?

sleevi: Correct. Certificates issued after the effective date are misissued if they are not exact character by character copies (that is, preserving case)

PR #292, opened 2021-07-12
SC45: Wildcard Domain Validation (#269)

castillar:
* Attempt to cleanup wildcard rules

Address longstanding feedback from the Validation Summit in Herdon
(Meeting 43) regarding wildcard domain names. This limits the
ability to use a certificate as an Authorization Domain Name (aka
parent & all child and grandchild domains) or to issue Wildcard
Certificates (which cover all child domains, but not grandchild
domains) to those methods which are linked to DNS-based proofs of
control/authorization.

* Adjust based on feedback

* Add titles for .18/.19 to Appendix B, to match .6

* Remove two-label exclusion and tighten language

* Restore 3.2.2.4.6 to handle if SC42 passes

If SC42 passes, the change to 3.2.2.4.6 is unnecessary, because no reused validations will be performed after the effective date.

Co-authored-by: Jos <castillar@melete.org>


sleevi:
@castillar Would it be possible that you use the Ballot description as the commit message when merging this? Thanks!

wthayer (approved):

PR #296, opened 2021-07-29
Adjust the workflow file to build the actions

sleevi:
This addresses a few requests that recently came up from the certificate
profiles work:

- Remove the explicit retention period (of 21 days) to allow the GitHub
  default of 90 days.
- Change the generated ZIP file from being "BR.md-hash" to being
  "BR-hash".
- Allow manually invoking the workflow (via workflow_dispatch), in the
  event folks want to re-run for a particular branch (e.g. profiles)
- Attempt to resolve the "non-deterministic redline" noted by Jos. When
  a given commit is on cabforum/servercert, it may be both a commit (to
  a branch) and part of a pull request (to main). We want the pull
  request redline to be against main, while the commit redline to be
  against the previous commit. Because both jobs run, and both upload
  the same file name, this results in a non-deterministic clobbering,
  where the commit-redline may clobber the pr-redline. This changes
  the generated zip file to be "file-hash-event_type", so that it
  will generate redlines for both PRs and commits and attach both.

### DIFF
--- a/.github/workflows/build-draft-docs.yml
+++ b/.github/workflows/build-draft-docs.yml
@@ -1,19 +1,20 @@
 name: Build Draft Guidelines
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 jobs:
   build_docs:
     strategy:
       matrix:
         document:
-          - 'BR.md'
-          - 'EVG.md'
-          - 'NSR.md'
+          - 'BR'
+          - 'EVG'
+          - 'NSR'
     name: Build ${{ matrix.document }}
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout the code
         uses: actions/checkout@v2
-      - name: Checkout old version for redline (pull request)
+      - name: Checkout old version for redline
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}
         uses: actions/checkout@v2
         with:
           ref: ${{ github.event.pull_request.base.sha || github.event.push.before }}
@@ -21,18 +22,17 @@ jobs:
       - uses: docker://ghcr.io/cabforum/build-guidelines-action:2.0.2
         id: build_doc
         with:
-          markdown_file: docs/${{ matrix.document }}
-          diff_file: old/docs/${{ matrix.document }}
+          markdown_file: docs/${{ matrix.document }}.md
+          diff_file: old/docs/${{ matrix.document }}.md
           pdf: true
           docx: true
           lint: true
           draft: ${{ !(github.event_name == 'push' && github.repository == 'cabforum/servercert' && github.ref == 'refs/heads/main') }}
       - uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.document }}-${{ github.event.pull_request.head.sha || github.sha }}
+          name: ${{ matrix.document }}-${{ github.event.pull_request.head.sha || github.sha }}-${{ github.event_name }}
           path: |
             ${{ steps.build_doc.outputs.pdf_file }}
             ${{ steps.build_doc.outputs.docx_file }}
             ${{ steps.build_doc.outputs.pdf_redline_file }}
           if-no-files-found: 'error'
-          retention-days: 21


.github/workflows/build-draft-docs.yml
```diff
@@ -1,38 +1,38 @@
 name: Build Draft Guidelines
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
```
sleevi, line 2: This only allows manual triggering if it's part of the `main` branch, and runs the version from `main`

So this won't, for example, allow manual trigger of a modified workflow from a branch or PR. It also means that newer versions (in main) may not be compatible with older targets (e.g. old tagged releases) - because the tagged release expects to be run with the tagged version.

So this is not a catchall, but is mostly useful for folks working on their own repos, who pick up this file, and want to run for their local branches. As I did with the profiles work, since I hadn't committed recently.

.github/workflows/build-draft-docs.yml
```diff
@@ -1,38 +1,38 @@
...
-          - 'NSR.md'
+          - 'BR'
+          - 'EVG'
+          - 'NSR'
     name: Build ${{ matrix.document }}
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout the code
         uses: actions/checkout@v2
-      - name: Checkout old version for redline (pull request)
```
sleevi, line 16: This was a left-over I meant to cleanup.

.github/workflows/build-draft-docs.yml
```diff
@@ -1,38 +1,38 @@
...
   build_docs:
     strategy:
       matrix:
         document:
-          - 'BR.md'
-          - 'EVG.md'
-          - 'NSR.md'
+          - 'BR'
+          - 'EVG'
+          - 'NSR'
```
sleevi, line 10: This was to satisfy a request from @vanbroup who found the "BR.md-hash" naming confusing, which was entirely reasonable.

.github/workflows/build-draft-docs.yml
```diff
@@ -1,38 +1,38 @@
...
     name: Build ${{ matrix.document }}
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout the code
         uses: actions/checkout@v2
-      - name: Checkout old version for redline (pull request)
+      - name: Checkout old version for redline
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}
```
sleevi, line 17: This is so that we don't produce redlines for manually triggered events, since we have no "thing to redline against" (yes, we could do the previous commit here, but that ended up being messy to get so I didn't bother)

If the `diff_file` doesn't exist (line 26), then the job will still continue, and just skip the redline production.

.github/workflows/build-draft-docs.yml
```diff
@@ -1,38 +1,38 @@
...
+          markdown_file: docs/${{ matrix.document }}.md
+          diff_file: old/docs/${{ matrix.document }}.md
           pdf: true
           docx: true
           lint: true
           draft: ${{ !(github.event_name == 'push' && github.repository == 'cabforum/servercert' && github.ref == 'refs/heads/main') }}
       - uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.document }}-${{ github.event.pull_request.head.sha || github.sha }}
+          name: ${{ matrix.document }}-${{ github.event.pull_request.head.sha || github.sha }}-${{ github.event_name }}
```
sleevi, line 33: I *think* this should fix your issue; for the stuff you're working on in branches, where you make some changes (e.g. to the informative table), you'll get two *separate* attachments here - one sufficed `push` (which just has your commit) and one sufficed `pull_request` (which will have all the changes since `main`)

To be fair, I haven't fully tested (since that'd require merging to _my_ main), but hey, commits are cheap, right? :)

.github/workflows/build-draft-docs.yml
```diff
@@ -1,38 +1,38 @@
...
           path: |
             ${{ steps.build_doc.outputs.pdf_file }}
             ${{ steps.build_doc.outputs.docx_file }}
             ${{ steps.build_doc.outputs.pdf_redline_file }}
           if-no-files-found: 'error'
-          retention-days: 21
```
sleevi, line 38: I mean, 21 days is nice in the YAGNI, but for folks hacking on branches, it's nice to leave them the full 90 days.

castillar (approved):
This looks great!

PR #322, opened 2021-10-07
Typo correction

hablutzel1:


sleevi:
@hablutzel1 All PRs to any of the Forum Guidelines need to go through the ballot process explained in Sections 2.3 and 2.4 of the Bylaws - https://cabforum.org/wp-content/uploads/CA-Browser-Forum-Bylaws-v2.3.pdf

Thus any edits (even typographical) need to be balloted and endorsed. Generally, we track these as issues for cleanup ballots, to capture all of the typographical changes needed and to reduce the ballot overhead.

PR #323, opened 2021-10-15
SC50 - Remove the requirements of 4.1.1

clintwilson:
When attempting to reduce the retention period required for audit logs and data archives, the NetSec Subcommittee also identified gaps in which data a CA is required to retain which make it somewhat difficult to make the desired adjustments to retention period.

Specifically, a CA is currently required to retain, but not use, data as defined in 4.1.1.

While reviewing the intent, purpose, and real-world usage around section 4.1.1 it became apparent that there’s little value in requiring CAs to maintain a database for which there is no prescribed purpose or required action.

This ballot seeks to address the above gap by replacing section 4.1.1 with "No stipulation." as is appropriate based on current expectations here.


castillar:
Merging after ballot passage.

PR #328, opened 2021-11-22
Ballot SC50 - Remove the requirements of 4.1.1

castillar:
# BALLOT SC50: Remove the requirements of 4.1.1 #

## PURPOSE OF BALLOT ##

When attempting to reduce the retention period required for audit logs and data archives, the NetSec Subcommittee also identified gaps in which data a CA is required to retain which make it somewhat difficult to make the desired adjustments to retention period. Specifically, a CA is currently required to retain, but not use, data as defined in 4.1.1 of the BRs.
While reviewing the intent, purpose, and real-world usage around section 4.1.1, it became apparent that there’s little value in requiring CAs to maintain a database for which there is no prescribed purpose or required action. This ballot seeks to address this gap by replacing section 4.1.1 with "No stipulation." as is appropriate based on current expectations here.

The following motion has been proposed by Clint Wilson of Apple and endorsed by Trevoli Ponds-White of Amazon and Dustin Hollenback of Microsoft.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1,9 +1,9 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates
-subtitle: Version 1.8.0
+subtitle: Version 1.8.1
 author:
   - CA/Browser Forum
-date: 25 August, 2021  
+date: 23 December, 2021  
 copyright: |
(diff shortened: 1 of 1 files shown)


dougbeattie (approved):
The changes align with the ballot - I approve.

PR #330, opened 2021-12-20
Sunset SHA-1 for OCSP signing

### LINKED ISSUE #324: Disallow SHA-1 for delegated OCSP response signing
The TLS BRs currently allow signing OCSP responses with the SHA-1 hash algorithm if delegated signing is used.

Given the risks of the continued use of SHA-1, we should consider prohibiting SHA-1 for delegated response signing as well.

CBonnell:
The BRs currently allow signing of OCSP responses if they are signed using delegated responders. Given the weaknesses with SHA-1 that have been known for several years, we should sunset the use of SHA-1 for OCSP signing altogether.

Fixes #324.


castillar (commented):
LGTM

PR #392, opened 2022-09-25
Update to allow multiple instances of subject attributes

dzacharo:
To allow for multiple instances of the domainContact attributes until we can address at an upcoming ballot.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -2689,7 +2689,7 @@ The following table details the acceptable `AttributeType`s that may appear with
 | `stateOrProvinceName`    | MAY             | If present, the CA's state or province information. | [Section 3.2.2.1](#3221-identity) |
 | `localityName`           | MAY             | If present, the CA's locality. | [Section 3.2.2.1](#3221-identity) |
 | `postalCode`             | MAY             | If present, the CA's zip or postal information. | [Section 3.2.2.1](#3221-identity) |
-| `streetAddress`          | MAY             | If present, the CA's street address. Multiple instances MAY be present. | [Section 3.2.2.1](#3221-identity) |
+| `streetAddress`          | MAY             | If present, the CA's street address. | [Section 3.2.2.1](#3221-identity) |
 | `organizationName`       | MUST            | The CA's name or DBA. The CA MAY include information in this field that differs slightly from the verified name, such as common variations or abbreviations, provided that the CA documents the difference and any abbreviations used are locally accepted abbreviations; e.g. if the official record shows "Company Name Incorporated", the CA MAY use "Company Name Inc." or "Company Name". | [Section 3.2.2.2](#3222-dbatradename) |
 | `commonName`             | MUST            | The contents SHOULD be an identifier for the certificate such that the certificate's Name is unique across all certificates issued by the issuing certificate. | |
 | Any other attribute      | NOT RECOMMENDED | -           | See [Section 7.1.4.3](#7143-other-subject-attributes) |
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -3031,7 +3031,6 @@ When encoding a `Name`, the CA SHALL ensure that:
   * Each `RelativeDistinguishedName` MUST contain exactly one `AttributeTypeAndValue`.
   * Each `RelativeDistinguishedName`, if present, is encoded within the `RDNSequence` in the order that it appears in [Section 7.1.4.2](#7142-subject-attribute-encoding).
     * For example, a `RelativeDistinguishedName` that contains a `countryName` `AttributeTypeAndValue` pair MUST be encoded within the `RDNSequence` before a `RelativeDistinguishedName` that contains a `stateOrProvinceName` `AttributeTypeAndValue`.
-  * Each `Name` MUST NOT contain more than one instance of a given `AttributeTypeAndValue` across all `RelativeDistinguishedName`s.
```
clintwilson, line 3034: I had understood the consensus to be that we'd add a carve-out to specifically add domainComponent as another field that allows multiple, as that was the only other field identified as being impacted by this change. Are there other fields which need to be present multiple times?

CBonnell: I think it makes sense to add a carve out to allow multiple instances of `streetAddress` and `domainComponent` and prohibit all other attributes from appearing more than once.

Since OU is now prohibited, I can't think of any other attribute types that would need to appear more than once in an RDN.

dzacharo: > We've been discussing multiple instances for a few years now, and streetAddress, OU, and domainComponent are the only three that have been identified as needing multiples. The situation is less clear in other ecosystems, but here on the TLS side, we have certificate transparency, so we're fairly confident there aren't other things hiding out here.
> 

As others have stated, the OU is currently not prohibited for CA Certificates.


> We're better off calling out explicitly the fields where it makes sense for repeats, instead of allowing arbitrary repeats for no reason, that may even have security implications (imagine, for example, your firewall and browser disagree on which of two needlessly repeated elements is the right one, and make security decisions that are inconsistent with each other).

I assume you are referring to multiple `commonName` subject attributes, because that's the only field that used for TLS which would still be a validated entry :)

Overall, I have no objection to call out explicit fields that allow multiple instances. As long as people are OK with adding the `domainComponent` field without any explicit validation rules, I can make the change.

dzacharo: I believe the latest commits address the concerns raised about the `domainComponent` and `streetAddress` attributes. I made no changes to address the OU field in CA Certificates. If we want to explicitly allow that, please let me know.

docs/BR.md
```diff
@@ -2289,13 +2289,14 @@ Table: Individual Validated `subject` Attributes
...
 | `surname`                      | MUST NOT        | -           | -           |
 | `givenName`                    | MUST NOT        | -           | -           |
 | `organizationalUnitName`       | -               | -           | -           |
 | \ \ \ \ _Prior to 2022-09-01_  | NOT RECOMMENDED | If present, the CA MUST implement a process that prevents an OU attribute from including a name, DBA, tradename, trademark, address, location, or other text that refers to a specific natural person or Legal Entity unless the CA has verified this information. | [Section 3.2](#32-initial-identity-validation) |
 | \ \ \ \ _Effective 2022-09-01_ | MUST NOT        | -           | -           |
+| `domainComponent`       | MAY | If present, this field MUST contain a label from a Domain Name. The `domainComponent` fields for each Domain Name MUST be in a single ordered sequence containing all labels from the Domain name. The labels MUST be encoded in the reverse order to the on-wire representation of domain names in the DNS protocol, so that the label closest to the root is encoded first. Multiple instances MAY be present. | [Section 3.2]
```
CBonnell, line 2299:
```suggestion
| `domainComponent`       | MAY | If present, this field MUST contain a Domain Label from a Domain Name. The `domainComponent` fields for the Domain Name MUST be in a single ordered sequence containing all Domain Labels from the Domain Name. The Domain Labels MUST be encoded in the reverse order to the on-wire representation of domain names in the DNS protocol, so that the Domain Label closest to the root is encoded first. Multiple instances MAY be present. | [Section 3.2]
```

PR #402, opened 2022-11-09
Proposed Ballot: Make OCSP Optional, Require CRLs, and Incentivize Automation

ryancdickson:
In response to #389 and referenced [here](https://lists.cabforum.org/pipermail/servercert-wg/2022-November/003420.html) and [here](https://docs.google.com/document/d/180T6cDSWPy54Rb5d6R4zN7MuLEMShaZ4IRLQgdPqE98/edit)- this PR proposes changes to the BRs related to making OCSP optional, requiring CRLs, and incentivizes automation and the issuance of short-lived certificates.

This PR is being opened to help collect feedback and facilitate discussion.

PR #414, opened 2023-01-18
Proposal: Make OCSP Optional, Require CRLs, and Incentivize Automation / Short-Lived Certificates

ryancdickson:
This pull request is a follow-up to #402, which inadvertently made edits to the in-progress "profiles" branch. 

Due to conflicts, #402 was closed. Since closing #402, a separate branch (i.e., "make-ocsp-optional") has been created dedicated to this specific proposal. Comments made on #402 will be transposed here.

Background, motivation, and additional information is available [here](https://docs.google.com/document/d/180T6cDSWPy54Rb5d6R4zN7MuLEMShaZ4IRLQgdPqE98/edit).

This "Pull Request" is intended to be used to foster discussion within the community in advance of requesting endorsors and beginning the balloting process.


docs/BR.md
```diff
@@ -452,6 +455,8 @@ The script outputs:
 
 **Root Certificate**: The self-signed Certificate issued by the Root CA to identify itself and to facilitate verification of Certificates issued to its Subordinate CAs.
 
+**Short-lived Subscriber Certificate**: A server Certificate with a Validity Period that is less than ten (10) days (864,000 seconds).
```
ryancdickson, line 458: On #402, @BenWilson-Mozilla commented...

```I would add "that is" so that it reads, "with a Validity Period that is less than ..."```

This change has been incorporated.

File: docs/BR.md [hunk 2]
```diff
@@ -1272,55 +1279,64 @@ No stipulation.
...
+- a full and complete; OR 
+- partitioned (i.e., "sharded") CRLs, that when aggregated, represent the equivalent of the full and complete CRL.
 
-The CA SHALL update and reissue CRLs at least:
+CRLs must be available via a publicly-accessible HTTP URL. 
 
-  i. once every twelve months; and
-  ii. within 24 hours after revoking a Subordinate CA Certificate.
+CAs issuing Subscriber Certificates (i.e., Certificates for servers) SHALL:
+- update and reissue CRLs at least once daily; and
```
XolphinMartijn, line 1291: If the intent is indeed to move from "at least once every seven days", to "at least daily", should we rephrase this to "at least once every 24 hours"?

docs/BR.md
```diff
@@ -1205,27 +1210,29 @@ No stipulation.
...
 
 1. The Subscriber requests in writing that the CA revoke the Certificate;
 2. The Subscriber notifies the CA that the original certificate request was not authorized and does not retroactively grant authorization;
 3. The CA obtains evidence that the Subscriber's Private Key corresponding to the Public Key in the Certificate suffered a Key Compromise;
 4. The CA is made aware of a demonstrated or proven method that can easily compute the Subscriber's Private Key based on the Public Key in the Certificate (such as a Debian weak key, see <https://wiki.debian.org/SSLkeys>);
 5. The CA obtains evidence that the validation of domain authorization or control for any Fully-Qualified Domain Name or IP address in the Certificate should not be relied upon.
 
-The CA SHOULD revoke a certificate within 24 hours and MUST revoke a Certificate within 5 days if one or more of the following occurs:
+With the exception of Short-lived Subscriber Certificates, the CA SHOULD revoke a certificate within 24 hours and MUST revoke a Certificate within 5 days if one or more of the following occurs:
```
ryancdickson, line 1223: On #402, @BenWilson-Mozilla commented:

```Same comment about 10-day validity period for short-lived certificates vs. 24-hour or 5-day revocation (assuming that revocation can cause a TLS connection to hard-fail).```

In response, please see my comment above. [Open for continued discussion!]

File: docs/BR.md [hunk 4]
```diff
@@ -1272,55 +1279,64 @@ No stipulation.
...
+- partitioned (i.e., "sharded") CRLs, that when aggregated, represent the equivalent of the full and complete CRL.
 
-The CA SHALL update and reissue CRLs at least:
+CRLs must be available via a publicly-accessible HTTP URL. 
 
-  i. once every twelve months; and
-  ii. within 24 hours after revoking a Subordinate CA Certificate.
+CAs issuing Subscriber Certificates (i.e., Certificates for servers) SHALL:
+- update and reissue CRLs at least once daily; and
+- the value of the `nextUpdate` field MUST NOT be more than ten (10) days beyond the value of the `thisUpdate` field.
```
ryancdickson, line 1292: On #402, @BenWilson-Mozilla commented:

```Can this 10-day window be shortened?```

In part, the validity of the as-proposed "short-lived certificate" was intended to align with this timeline (i.e., 10 days) as it represented how long a client might cache and rely on a CRL.

Given CAs would be obligated to generate and publish new CRLs every 24 hours ("at least once daily"), applications like CRLSet/CRLite/OneCRL would be capable of obtaining fresh revocation status information every 24 hours, regardless of the nextUpdate.

Given the clarification above, do you still have concern about the 10-day timeline?

File: docs/BR.md (hunk 1 above)
Comment (ryancdickson) [inline docs/BR.md:458]: In #402, @BenWilson-Mozilla commented:

```Can the Validity Period be defined as a shorter period, e.g. five days, 432,000 seconds?```

Open for discussion. 

10 days was in attempt to:
- align with existing timelines (e.g.,  CRL "nextUpdate") and implementations observed in some browsers  
- offer an attractive incentive for CAs to adopt automation. Reducing the period too far may result in limited adoption due to the impact of unforeseen outages or operational issues. IMO, a validity of 10 days is more "forgiving" of an unexpected 2-3 day service outage than a validity of 5 days).

What are your thoughts on a phased approach (sample below):
- Until 2026: 10 days
- Effective 1/1/26: 5 days

File: docs/BR.md [hunk 5]
```diff
@@ -1272,55 +1279,64 @@ No stipulation.
...
-3. For OCSP responses with validity intervals less than sixteen hours, then the CA SHALL update the information provided via an Online Certificate Status Protocol prior to one-half of the validity period before the nextUpdate.
-4. For OCSP responses with validity intervals greater than or equal to sixteen hours, then the CA SHALL update the information provided via an Online Certificate Status Protocol at least eight hours prior to the nextUpdate, and no later than four days after the thisUpdate.
-
-For the status of Subordinate CA Certificates:
+- Prior to 2020-09-30:
+     1. The CA SHALL update information provided via an Online Certificate Status Protocol at least every four days. 
+     2. OCSP responses from this service MUST have a maximum expiration time of ten days.
 
-* The CA SHALL update information provided via an Online Certificate Status Protocol
+- Effective 2020-09-30:
```
ryancdickson, line 1330: On #402, @BenWilson-Mozilla commented:

```Should this be shortened, too?```

Open for community discussion. This proposal does not change the existing OCSP requirements, it merely makes OCSP optional.

File: docs/BR.md [hunk 6]
```diff
@@ -1368,7 +1384,7 @@ Revocation entries on a CRL or OCSP Response MUST NOT be removed until after the
 
 ### 4.10.2 Service availability
 
-The CA SHALL operate and maintain its CRL and OCSP capability with resources sufficient to provide a response time of ten seconds or less under normal operating conditions.
+The CA SHALL operate and maintain its CRL and optional OCSP capability with resources sufficient to provide a response time of ten seconds or less under normal operating conditions.
```
ryancdickson, line 1387: In #402, @BenWilson-Mozilla commented:

I think there has been discussion about eliminating or re-phrasing this requirement.

[will leave open for community discussion]

docs/BR.md
```diff
@@ -2858,7 +2876,18 @@ This section contains several fields that are common among multiple certificate
 
 ##### 7.1.2.11.2 CRL Distribution Points
 
-If present, the CRL Distribution Points extension MUST be formatted as follows:
+The CRL Distribution Points extension MUST be present in:
+- Subordinate CA Certificates; and 
+- Subscriber Certificates with validity greater than or equal to ten days.
```
ryancdickson, line 2881: In #402, @BenWilson-Mozilla commented:

```Change to 5 days, 432,000 seconds?```

Open for discussion (described in open comment above). This value will be set based on the validity of a short-lived certificate.

File: docs/BR.md (hunk 1 above)
Comment (ryancdickson) [reply docs/BR.md:458]: @BenWilson-Mozilla - added a phased reduction in 6.3.2.

```Subscriber Certificates issued on or after 1 September 2020 SHOULD NOT have a Validity Period greater than 397 days and MUST NOT have a Validity Period greater than 398 days. To qualify as a Short-lived Subscriber Certificate, Subscriber Certificates issued after 15 January 2024, but prior to 15 January 2026, MUST NOT have a Valdity Period greater than 10 days. Short-lived Subscriber Certificates issued on or after 15 January 2026 MUST NOT have a Validity Period greater than 7 days.```

File: docs/BR.md [hunk 8]
```diff
@@ -3147,45 +3174,88 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+
+A full and complete CRL is a list of all revoked certificates issued by the CA for any and all reasons.
+
+A partitioned CRL (sometimes referred to as a "sharded CRL") is a list of revoked certificates issued by the CA for any and all reasons constrained to a specific scope (e.g., temporal sharding).
+
+Minimally, CAs MUST issue either a "full and complete" CRL - or "partitioned" CRLs. Aside from the presence of the `IssuingDistributionPoint` extension (OID 2.5.29.28) in partitioned CRLs, both CRL formats are syntactically the same from the perspective of this profile.
+
+If using only partitioned CRLs, the full set of partitioned CRLs MUST cover the complete set of public-key certificates issued by the CA. Thus, the complete set of partitioned CRLs MUST be equivalent to a full CRL for the same set of public-key certificates, if the CA was not using partitioned CRLs.
```
aarongable, line 3193: This is wildly editorial, so feel free to ignore, but I would combine/rephrase/reorder these paragraphs. I works better for my brain this way, but that's definitely not a claim that my way works better for everyone or even a majority of people:

```suggestion
A full and complete CRL is CRL whose scope includes all certificates issued by the CA.

A partitioned CRL (sometimes referred to as a "sharded CRL") is a CRL with a a constrained scope, such as all certificates issued by the CA during a certain period of time ("temporal sharding"). Aside from the presence of the `IssuingDistributionPoint` extension (OID 2.5.29.28) in partitioned CRLs, both CRL formats are syntactically the same from the perspective of this profile.

Minimally, CAs MUST issue either a "full and complete" CRL or a set of "partitioned" CRLs which cover the complete set of certificates issued by the CA. In other words, if issuing only partitioned CRLs, the combined scope of those CRLs must be equivalent to that of a full and complete CRL. 
```

docs/BR.md
```diff
@@ -3147,45 +3174,88 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+| ---                        | ------          | ------          |
+| `tbsCertList`              |                 |                 |
+|     `version`              | MUST     | MUST be v2(1) |
+|     `signature`            | MUST     | See [Section 7.1.3.2](#7132-signature-algorithmidentifier) |
+|     `issuer`               | MUST     | MUST be byte-for-byte identical to the `subject` field of the Issuing CA. |
+|     `thisUpdate`           | MUST     | UTCTime (YYMMDDHHMMSSZ) MUST be used for dates up to and including 2049. GeneralizedTime (YYYYMMDDHHMMSSZ) MUST be used for dates after 2049.|
+|     `nextUpdate`           | MUST     | UTCTime (YYMMDDHHMMSSZ) MUST be used for dates up to and including 2049. GeneralizedTime (YYYYMMDDHHMMSSZ) MUST be used for dates after 2049.|
+|     `revokedCertificates`  | *        | MUST only be present if the CA has issued a certificate that is both revoked and unexpired. An entry MUST NOT be removed from the CRL until it appears on one regularly scheduled CRL issued beyond the revoked certificate's validity period. See the "revokedCertificates Component" table for additional requirements. |
```
aarongable, line 3208: This description is too strict. CRL entries are required to remain on a CRL for one CRL issuance cycle _after_ they expire (so that a cert which is revoked seconds before it expires appears on at least one CRL anyway). This requirement would force a CA to violate _that_ requirement if the CA has only expired-and-revoked certificates left, which is a common scenario for the very last CRL published at the end of a CA's issuing lifetime.

docs/BR.md
```diff
@@ -3147,45 +3174,88 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+| `signatureAlgorithm`       | MUST     | Encoded value MUST be byte-for-byte identical to the `tbsCertList.signature`. |
+| `signature`                | MUST     | - |
+| Any other value            | NOT RECOMMENDED | - |
+
+Table: CRL Extensions
+
+| __Extension__                     | __Presence__    | __Critical__          | __Description__ |
+| ----                              | -               | -                     | ----- |
+| `authorityKeyIdentifier`          | MUST            | N                     | The `keyIdentifier` field MUST be included and its value MUST be byte-for-byte identical to the Subject Key Identifier of the Issuing CA's Certificate. <br><br>The `authorityCertIssuer` and `authorityCertSerialNumber` fields MUST NOT be populated.      |
+| `CRLNumber`                       | MUST            | N                     | MUST contain an INTEGER greater than or equal to zero (0) and less than 2¹⁵⁹, and convey a monotonically increasing sequence.       |
```
aarongable, line 3219: "Monotonic" just means non-decreasing; CRL Numbers are required to be _strictly_ increasing (they cannot stay the same between CRL generations).
```suggestion
| `CRLNumber`                       | MUST            | N                     | MUST contain an INTEGER greater than or equal to zero (0) and less than 2¹⁵⁹, and convey a strictly increasing sequence.       |
```

docs/BR.md
```diff
@@ -1752,7 +1765,8 @@ The CA SHALL protect its Private Key in a system or device that has been validat
 
 ### 6.3.2 Certificate operational periods and key pair usage periods
 
-Subscriber Certificates issued on or after 1 September 2020 SHOULD NOT have a Validity Period greater than 397 days and MUST NOT have a Validity Period greater than 398 days.
+Subscriber Certificates issued on or after 1 September 2020 SHOULD NOT have a Validity Period greater than 397 days and MUST NOT have a Validity Period greater than 398 days. To qualify as a Short-lived Subscriber Certificate, Subscriber Certificates issued after 15 March 2024, but prior to 15 March 2026, MUST NOT have a Valdity Period greater than 10 days. Short-lived Subscriber Certificates issued on or after 15 March 2026 MUST NOT have a Validity Period greater than 7 days.
```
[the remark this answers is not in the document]
ryancdickson: Addressed in https://github.com/ryancdickson/staging/commit/df5e689b2b098b21aa997299b1c8dd1b927a74f4.

docs/BR.md
```diff
@@ -1280,27 +1284,32 @@ No stipulation.
...
-If the CA publishes a CRL, then the CA SHALL update and reissue CRLs at least once every seven days, and the value of the `nextUpdate` field MUST NOT be more than ten days beyond the value of the `thisUpdate` field.
+CRLs must be available via a publicly-accessible HTTP URL.
 
-For the status of Subordinate CA Certificates:
-
-The CA SHALL update and reissue CRLs at least:
-
-  i. once every twelve months; and
-  ii. within 24 hours after revoking a Subordinate CA Certificate.
+CAs issuing Subscriber Certificates (i.e., Certificates for servers) SHALL:
```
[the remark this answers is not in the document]
ryancdickson: Overcome by events, modified presentation format in https://github.com/ryancdickson/staging/blob/make-ocsp-optional-updates/docs/BR.md#497-crl-issuance-frequency. Will close this comment. Future comments should be made on https://github.com/ryancdickson/staging/pull/3 which will serve as the basis for the next round of public discussion.

docs/BR.md
```diff
@@ -1280,27 +1284,32 @@ No stipulation.
...
+- update and reissue CRLs at least once every 24 hours; and
+- the value of the `nextUpdate` field MUST NOT be more than ten (10) days beyond the value of the `thisUpdate` field.
 
-The value of the `nextUpdate` field MUST NOT be more than twelve months beyond the value of the `thisUpdate` field.
+CAs issuing CA Certificates SHALL:
```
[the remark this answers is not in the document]
ryancdickson: Same as response to prior comment. (OBE).

docs/BR.md
ryancdickson: Accepted suggestion with minor tweaks.  https://github.com/ryancdickson/staging/commit/12c2db11346481fd8efd150144692950827f968f

docs/BR.md
ryancdickson: Addressed in https://github.com/ryancdickson/staging/commit/f65d2470e4777136d567c5a4ff5efa42c216b206.

docs/BR.md
ryancdickson: Tweaked in https://github.com/ryancdickson/staging/commit/5cc135f2eec7241dd3868369e50b200b641cf813.

"MUST be present if the CA has issued a certificate that has been revoked and the corresponding entry has yet to appear on at least one regularly scheduled CRL beyond the revoked certificate's validity period. The CA SHOULD remove an entry for a corresponding certificate after it has appeared on at least one regularly scheduled CRL beyond the revoked certificate's validity period."

CBonnell: Roots that are under both TLS BR and CS BR audit scope will need to violate the SHOULD for revoked code signing ICAs (CSBRs require CRL revocation services for 10 years after expiry), but I think that following the CSBRs is compelling justification for violating this SHOULD-level requirement.

docs/BR.md
```diff
@@ -2887,7 +2910,18 @@ This section contains several fields that are common among multiple certificate
...
 ##### 7.1.2.11.2 CRL Distribution Points
 
-If present, the CRL Distribution Points extension MUST contain at least one `DistributionPoint`; containing more than one is NOT RECOMMENDED. All `DistributionPoint` items must be formatted as follows:
+The CRL Distribution Points extension MUST be present in:
+- Subordinate CA Certificates; and
+- Subscriber Certificates that 1) do not qualify as "Short-lived Subscriber Certificates" and 2) do not include an Authority Information Access extension with an id-ad-ocsp accessMethod.
+
+The CRL Distribution Points extension SHOULD NOT be present in:
+- Root CA Certificates; and
+- Short-lived Subscriber Certificates.
```
robstradling, line 2919: I think "SHOULD NOT be present in...Short-lived Subscriber Certificates" is a step too far.

CAs that issue <=398-day certificates may sometimes issue <10-day certificates, but not because they want to avoid having to provide revocation information.  For example: if a Subscriber suffers a key compromise 7 days before their certificate is due to expire, the CA might issue a 7-day replacement certificate (with a different public key) that will expire at the same time, in order to cover (but not exceed) the validity period that the Subscriber had already paid for.

```suggestion
The CRL Distribution Points extension is OPTIONAL in:
- Short-lived Subscriber Certificates.
```

docs/BR.md
```diff
@@ -3147,45 +3181,85 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
-  * **cessationOfOperation (RFC 5280 CRLReason #5):** Indicates that the website with the Certificate is shut down prior to the expiration of the Certificate, or if the Subscriber no longer owns or controls the Domain Name in the Certificate prior to the expiration of the Certificate; or
-  * **privilegeWithdrawn (RFC 5280 CRLReason #9):** Indicates that there has been a subscriber-side infraction that has not resulted in keyCompromise, such as the Certificate Subscriber provided misleading information in their Certificate Request or has not upheld their material obligations under the Subscriber Agreement or Terms of Use.
+**Note:** Backdating the revocationDate field is an exception to best practice described in RFC 5280 (Section 5.3.2); however, these requirements specify the use of the revocationDate field to support TLS implementations that process the revocationDate field as the date when the Certificate is first considered to be compromised.
```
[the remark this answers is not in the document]
robstradling: Also, I'm curious: Do there really exist "TLS implementations that process the revocationDate field as the date when the Certificate is first considered to be compromised"?  Or has this language accidentally been imported from the Code Signing BRs, which permit the RFC5280-non-compliant backdating of the revocationDate field as a workaround for the fact that Microsoft's stack doesn't support the RFC5280 [Invalidity Date](https://www.rfc-editor.org/rfc/rfc5280.html#section-5.3.2) extension?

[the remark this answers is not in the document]
robstradling: ISTM that TLS implementations only care about the revocation status of a certificate and pay no heed to any timestamps that indicate how far in the past the revocation occurred, whereas Code Signing does care about the "revocation time" and "invalidity date".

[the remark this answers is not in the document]
ryancdickson: Thanks for this comment, @robstradling.

I attempted to address the backdating concerns [here](https://github.com/ryancdickson/staging/pull/5/files).

Rather than accepting your suggestion as written (i.e., move the complete paragraph beginning with "When a CA obtains..." to the beginning of the note) - I split the paragraph such that discussion re: CRLReasons follows the CRLReason table. Otherwise, we're introducing language related to CRLReasons that are not yet enumerated. 

Any concerns with how I handled this?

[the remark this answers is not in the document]
ryancdickson: Re: your other comments - I agree with you re: "TLS implementations only care about the revocation status of a certificate and pay no heed to any timestamps that indicate how far in the past the revocation occurred".

It's not clear to me the motivation of the existing language re: backdating. @BenWilson-Mozilla - as the author of SC-61, can you offer any background?

docs/BR.md
ryancdickson: No concerns with this recommendation. 

Addressed in [this](https://github.com/ryancdickson/staging/pull/5/files) PR.

docs/BR.md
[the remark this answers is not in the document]
robstradling: @ryancdickson That change looks good to me.  Thanks.

(I still think that the BRs should not be permitting the backdating of revocationDate, but tackling that existing language might be considered scope creep for this particular ballot).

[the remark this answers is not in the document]
ryancdickson: Thanks for confirming, Rob!

I agree with your comment re: "scope creep," but will take action to create a separate issue such that we can explore this further in a future cleanup ballot.

docs/BR.md
```diff
@@ -3147,45 +3181,85 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+|     `extensions`           | MUST     | See the "CRL Extensions" table for additional requirements. |
+| `signatureAlgorithm`       | MUST     | Encoded value MUST be byte-for-byte identical to the `tbsCertList.signature`. |
+| `signature`                | MUST     | - |
+| Any other value            | NOT RECOMMENDED | - |
+
+Table: CRL Extensions
+
+| __Extension__                     | __Presence__    | __Critical__          | __Description__ |
+| ----                              | -               | -                     | ----- |
+| `authorityKeyIdentifier`          | MUST            | N                     | The `keyIdentifier` field MUST be included and its value MUST be byte-for-byte identical to the Subject Key Identifier of the Issuing CA's Certificate. <br><br>The `authorityCertIssuer` and `authorityCertSerialNumber` fields MUST NOT be populated.      |
```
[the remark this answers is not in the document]
ryancdickson: Addressed in: https://github.com/ryancdickson/staging/commit/7767d3da5eabcdfec13a6a7bd2975a7dd6bcf14e

docs/BR.md
```diff
@@ -3147,45 +3181,85 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+| `IssuingDistributionPoint`        | *           | Y                     | Partitioned CRLs MUST include at least one of the names from the corresponding distributionPoint field of the cRLDistributionPoints extension of every certificate that is within the scope of this CRL. The encoded value MUST be byte-for-byte identical to the encoding used in the distributionPoint field of the certificate. <br><br>The `indirectCRL` and `onlyContainsAttributeCerts` fields MUST be set to `FALSE` (i.e., not asserted). <br><br>The CA MAY set either of the `onlyContainsUserCerts` and `onlyContainsCACerts` fields to `TRUE`, depending on the scope of the CRL. <br><br>The CA MUST NOT assert both of the `onlyContainsUserCerts` and `onlyContainsCACerts` fields. <br><br>The `onlySomeReasons` field SHOULD NOT be included; if included, then the CA MUST provide another CRL whose scope encompasses all revocations regardless of reason code. <br><br>This extension is NOT RECOMMENDED for full and complete CRLs. |
```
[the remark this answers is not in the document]
ryancdickson: I had considered that, too. Thanks for the nudge. 

Addressed in https://github.com/ryancdickson/staging/commit/8d0022567dad16910c7486721403beb40b97b448

docs/BR.md
```diff
@@ -3147,45 +3181,85 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
-   If a `reasonCode` CRL entry extension is present, the `CRLReason` MUST indicate the most appropriate reason for revocation of the Certificate.
-   
-    CRLReason MUST be included in the `reasonCode` extension of the CRL entry corresponding to a Subscriber Certificate that is revoked after July 15, 2023, unless the CRLReason is "unspecified (0)". Revocation reason code entries for Subscriber Certificates revoked prior to July 15, 2023, do NOT need to be added or changed.
+Table: revokedCertificates Component
 
-Only the following CRLReasons MAY be present in the CRL `reasonCode` extension for Subscriber Certifificates:
+| __Component__                     | __Presence__    | __Description__ |
+| ----                              | -               | ----- |
+| `serialNumber`                    | MUST            | MUST be byte-for-byte identical to the serialNumber contained in the revoked certificate. |
+| `revocationDate`                  | MUST            | The date and time which revocation occurred. UTCTime (YYMMDDHHMMSSZ) MUST be used for dates up to and including 2049. GeneralizedTime (YYYYMMDDHHMMSSZ) MUST be used for dates after 2049.  |
```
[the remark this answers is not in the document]
ryancdickson: Agree - addressed in https://github.com/ryancdickson/staging/commit/09a563c73dfb4c33e285754a441a112d5507848c

docs/BR.md
```diff
@@ -3147,45 +3181,85 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+|     `nextUpdate`           | MUST     | UTCTime (YYMMDDHHMMSSZ) MUST be used for dates up to and including 2049. GeneralizedTime (YYYYMMDDHHMMSSZ) MUST be used for dates after 2049.|
+|     `revokedCertificates`  | *        | MUST be present if the CA has issued a certificate that has been revoked and the corresponding entry has yet to appear on at least one regularly scheduled CRL beyond the revoked certificate's validity period. The CA SHOULD remove an entry for a corresponding certificate after it has appeared on at least one regularly scheduled CRL beyond the revoked certificate's validity period. See the "revokedCertificates Component" table for additional requirements.  |
+|     `extensions`           | MUST     | See the "CRL Extensions" table for additional requirements. |
+| `signatureAlgorithm`       | MUST     | Encoded value MUST be byte-for-byte identical to the `tbsCertList.signature`. |
+| `signature`                | MUST     | - |
+| Any other value            | NOT RECOMMENDED | - |
```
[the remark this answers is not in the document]
ryancdickson: Addressed in: https://github.com/ryancdickson/staging/commit/c2dadd3a97c3b68a9a7363ed78c6c279ba89ca75

docs/BR.md
ryancdickson: Closing stale comments.

docs/BR.md
ryancdickson: Closing stale comments.

docs/BR.md
ryancdickson: Closing stale comments.

docs/BR.md
ryancdickson: Closing stale comments.

docs/BR.md
ryancdickson: Closing stale comments.

docs/BR.md
ryancdickson: Closing stale comments.

docs/BR.md
```diff
@@ -1205,27 +1210,29 @@ No stipulation.
 
 #### 4.9.1.1 Reasons for Revoking a Subscriber Certificate
 
-The CA SHALL revoke a Certificate within 24 hours if one or more of the following occurs:
+The CA MAY support revocation of Short-lived Subscriber Certificates.
+
+With the exception of Short-lived Subscriber Certificates, the CA SHALL revoke a Certificate within 24 hours if one or more of the following occurs:
```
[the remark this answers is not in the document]
ryancdickson: Closing stale comments.

docs/BR.md
```diff
@@ -3147,45 +3181,85 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
-   If a CRL entry is for a Certificate not subject to these Requirements and was either issued on-or-after 2020-09-30 or has a `notBefore` on-or-after 2020-09-30, the `CRLReason` MUST NOT be certificateHold (6).
-   If a CRL entry is for a Certificate subject to these Requirements, the `CRLReason` MUST NOT be certificateHold (6).
+| __Field__                  | __Presence__    | __Description__ |
+| ---                        | ------          | ------          |
+| `tbsCertList`              |                 |                 |
+|     `version`              | MUST     | MUST be v2(1) |
+|     `signature`            | MUST     | See [Section 7.1.3.2](#7132-signature-algorithmidentifier) |
+|     `issuer`               | MUST     | MUST be byte-for-byte identical to the `subject` field of the Issuing CA. |
+|     `thisUpdate`           | MUST     | UTCTime (YYMMDDHHMMSSZ) MUST be used for dates up to and including 2049. GeneralizedTime (YYYYMMDDHHMMSSZ) MUST be used for dates after 2049.|
+|     `nextUpdate`           | MUST     | UTCTime (YYMMDDHHMMSSZ) MUST be used for dates up to and including 2049. GeneralizedTime (YYYYMMDDHHMMSSZ) MUST be used for dates after 2049.|
```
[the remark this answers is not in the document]
aarongable: This language works for me.

docs/BR.md
```diff
@@ -2887,7 +2907,20 @@ This section contains several fields that are common among multiple certificate
...
+The CRL Distribution Points extension SHOULD NOT be present in:
+- Root CA Certificates.
+
+The CRL Distribution Points extension is OPTIONAL in:
+- Short-lived Subscriber Certificates.
+
+The CRL Distribution Points extension MUST NOT be present in:
+- OCSP Responder Certificates.
+
+When present, the CRL Distribution Points extension MUST contain at least one `DistributionPoint`; containing more than one is NOT RECOMMENDED. All `DistributionPoint` items must be formatted as follows:
```
aarongable, line 2923:
```suggestion
When present, the CRL Distribution Points extension MUST contain at least one `DistributionPoint`; containing more than one is NOT RECOMMENDED. All `DistributionPoint` items MUST be formatted as follows:
```
maybe? I'm not sure why this "must" wasn't capitalized previously.

docs/BR.md
```diff
@@ -2748,7 +2768,7 @@ The `AuthorityInfoAccessSyntax` MAY contain multiple `AccessDescription`s with t
 
 | __Access Method__ | __OID__            | __Access Location__         | __Presence__ | __Maximum__ | __Description__ |
 | --                | --                 | ----                        | -            | -          | ---             |
-| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | SHOULD       | \*         | A HTTP URL of the Issuing CA's OCSP responder. |
+| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | MAY       | \*         | A HTTP URL of the Issuing CA's OCSP responder. |
```
aarongable, line 2771: nit: column alignment
```suggestion
| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | MAY          | \*         | A HTTP URL of the Issuing CA's OCSP responder. |
```

docs/BR.md
```diff
@@ -3147,44 +3180,96 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+|     `thisUpdate`           | MUST     | Indicates the issue date of the CRL. |
+|     `nextUpdate`           | MUST     | Indicates the date by which the next CRL will be issued. For CRLs covering Subscriber Certificates, at most 10 days after the `thisUpdate`. For other CRLs, at most 12 months after the `thisUpdate`. |
+|     `revokedCertificates`  | *        | MUST be present if the CA has issued a certificate that has been revoked and the corresponding entry has yet to appear on at least one regularly scheduled CRL beyond the revoked certificate's validity period. The CA SHOULD remove an entry for a corresponding certificate after it has appeared on at least one regularly scheduled CRL beyond the revoked certificate's validity period. See the "revokedCertificates Component" table for additional requirements.  |
+|     `extensions`           | MUST     | See the "CRL Extensions" table for additional requirements. |
+| `signatureAlgorithm`       | MUST     | Encoded value MUST be byte-for-byte identical to the `tbsCertList.signature`. |
+| `signature`                | MUST     | - |
+| Any other value            | NOT RECOMMENDED | - |
```
aarongable, line 3209: nit: column alignment
```suggestion
| __Field__                  | __Presence__    | __Description__ |
| ---                        | ------          | ------          |
| `tbsCertList`              |                 |                 |
|     `version`              | MUST            | MUST be v2(1), see [Section 7.2.1](#721-version-numbers) |
|     `signature`            | MUST            | See [Section 7.1.3.2](#7132-signature-algorithmidentifier) |
|     `issuer`               | MUST            | MUST be byte-for-byte identical to the `subject` field of the Issuing CA. |
|     `thisUpdate`           | MUST            | Indicates the issue date of the CRL. |
|     `nextUpdate`           | MUST            | Indicates the date by which the next CRL will be issued. For CRLs covering Subscriber Certificates, at most 10 days after the `thisUpdate`. For other CRLs, at most 12 months after the `thisUpdate`. |
|     `revokedCertificates`  | *               | MUST be present if the CA has issued a certificate that has been revoked and the corresponding entry has yet to appear on at least one regularly scheduled CRL beyond the revoked certificate's validity period. The CA SHOULD remove an entry for a corresponding certificate after it has appeared on at least one regularly scheduled CRL beyond the revoked certificate's validity period. See the "revokedCertificates Component" table for additional requirements.  |
|     `extensions`           | MUST            | See the "CRL Extensions" table for additional requirements. |
| `signatureAlgorithm`       | MUST            | Encoded value MUST be byte-for-byte identical to the `tbsCertList.signature`. |
| `signature`                | MUST            | - |
| Any other value            | NOT RECOMMENDED | - |
```

docs/BR.md
```diff
@@ -3147,44 +3180,96 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
 
-1. `reasonCode` (OID 2.5.29.21)
+Table: CRL Extensions
+
+| __Extension__                     | __Presence__    | __Critical__          | __Description__ |
+| ----                              | -               | -                     | ----- |
+| `authorityKeyIdentifier`          | MUST            | N                     | See [Section 7.1.2.11.1](#712111-authority-key-identifier) |
+| `CRLNumber`                       | MUST            | N                     | MUST contain an INTEGER greater than or equal to zero (0) and less than 2¹⁵⁹, and convey a strictly increasing sequence.        |
+| `IssuingDistributionPoint`        | *           | Y                     | See [Section 7.2.2.1 CRL Issuing Distribution Point](#7221-crl-issuing-distribution-point) |
+| Any other extension               | NOT RECOMMENDED        | -                     |       |
```
aarongable, line 3224: nit: column alignment
```suggestion
| __Extension__                     | __Presence__    | __Critical__          | __Description__ |
| ----                              | -               | -                     | ----- |
| `authorityKeyIdentifier`   | MUST            | N                     | See [Section 7.1.2.11.1](#712111-authority-key-identifier) |
| `CRLNumber`                | MUST            | N                     | MUST contain an INTEGER greater than or equal to zero (0) and less than 2¹⁵⁹, and convey a strictly increasing sequence.        |
| `IssuingDistributionPoint` | *               | Y                     | See [Section 7.2.2.1 CRL Issuing Distribution Point](#7221-crl-issuing-distribution-point) |
| Any other extension        | NOT RECOMMENDED | -                     |       |
```

docs/BR.md
```diff
@@ -3147,44 +3180,96 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
 
 The privilegeWithdrawn reasonCode SHOULD NOT be made available to the Subscriber as a revocation reason option, because the use of this reasonCode is determined by the CA and not the Subscriber.
 
-When a CA obtains verifiable evidence of Key Compromise for a Certificate whose CRL entry does not contain a reasonCode extension or has a reasonCode extension with a non-keyCompromise reason, the CA SHOULD update the CRL entry to enter keyCompromise as the CRLReason in the reasonCode extension. Additionally, the CA SHOULD update the revocation date in a CRL entry when it is determined that the private key of the certificate was compromised prior to the revocation date that is indicated in the CRL entry for that certificate.
+When a CA obtains verifiable evidence of Key Compromise for a Certificate whose CRL entry does not contain a reasonCode extension or has a reasonCode extension with a non-keyCompromise reason, the CA SHOULD update the CRL entry to enter keyCompromise as the CRLReason in the reasonCode extension. 
+
+#### 7.2.2.1 CRL Issuing Distribution Point
```
aarongable, line 3261: nit: empty line after header
```suggestion
#### 7.2.2.1 CRL Issuing Distribution Point

```

docs/BR.md
```diff
@@ -3147,44 +3180,96 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+| Any other extension               | NOT RECOMMENDED        | -                     |       |
+
+Table: revokedCertificates Component
 
-   If present, this extension MUST NOT be marked critical.
+| __Component__                     | __Presence__    | __Description__ |
+| ----                              | -               | ----- |
+| `serialNumber`                    | MUST            | MUST be byte-for-byte identical to the serialNumber contained in the revoked certificate. |
+| `revocationDate`                  | MUST            | Normally, the date and time revocation occurred. See the footnote following this table for circumstances where backdating is permitted.  | 
+| `crlEntryExtensions`              | *               | See the "crlEntryExtensions Component" table for additional requirements. |
```
aarongable, line 3232: nit: column width
```suggestion
| __Component__        | __Presence__ | __Description__ |
| ----                              | -               | ----- |
| `serialNumber`       | MUST         | MUST be byte-for-byte identical to the serialNumber contained in the revoked certificate. |
| `revocationDate`     | MUST         | Normally, the date and time revocation occurred. See the footnote following this table for circumstances where backdating is permitted.  | 
| `crlEntryExtensions` | *            | See the "crlEntryExtensions Component" table for additional requirements. |
```

aarongable (approved):
LGTM. Looking at this with fresh post-vacation eyes caught one more phrasing thing, and then I finally allowed myself to comment on the formatting nits that don't affect the meaning of the ballot.

docs/BR.md
```diff
@@ -1280,27 +1286,39 @@ No stipulation.
...
-If the CA publishes a CRL, then the CA SHALL update and reissue CRLs at least once every seven days, and the value of the `nextUpdate` field MUST NOT be more than ten days beyond the value of the `thisUpdate` field.
+Within twenty-four (24) hours of issuing its first Certificate, the CA MUST generate and publish either:
+- a full and complete CRL; OR
+- partitioned (i.e., "sharded") CRLs that, when aggregated, represent the equivalent of a full and complete CRL.
 
-For the status of Subordinate CA Certificates:
+CAs issuing Subscriber Certificates:  
+1. MUST update and publish a new CRL within at least: 
+     - seven (7) days if all Certificates include an Authority Information Access extension with an id-ad-ocsp accessMethod (“AIA OCSP pointer”); or
+     - four (4) days in all other cases;
```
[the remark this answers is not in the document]
dzacharo: If we don't use "**at least** every X days" then if a CA decides to publish a new CRL sooner it would violate the requirement.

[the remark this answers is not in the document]
aarongable: Ah good catch, then I'm in favor of replacing "within at least" with "at least every".

docs/BR.md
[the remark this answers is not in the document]
ryancdickson: Given it seems unlikely we'll get through the Ballot in 3 rounds of discussion due to some recent comments, I might explore this in a Version 4 (hopefully the last). 

I'll follow-up with @BenWilson-Mozilla to understand why the backdating discussion was pulled into the TLS BRs via SC-61. 

My preference would be to stipulate the revocation date listed on the CRL must be the _actual_ date of revocation. Any opinion on that @robstradling?

While studying the CRLs disclosed to CCADB, we noted a small number of instances where:
- Certificates appeared to be revoked before they were issued  ([example](https://crt.sh/?q=4e48fab0934f50cd3a42d00653089f1a90d56f6157cad82d9afce2e6dc81be4b))
- Certificates appeared to be revoked at the same time as when they were issued ([example](https://crt.sh/?id=7085490113))

As an outside observer, it's impossible to know the _actual_ revocation date - however we can gain some context  by observing when an entry lands on the CRL (not easily monitored using existing public tools).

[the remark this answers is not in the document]
ryancdickson: Update: Discussed a bit with @CBonnell and @wthayer during lunch at the F2F. 

In cases where a pre-certificate is signed and the corresponding certificate is _never_ issued, setting revocationDate to equal the notBefore date makes sense. 

I'd still be interested in learning whether there are other valid use cases for backdating revocationDate.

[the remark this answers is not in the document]
BenWilson-Mozilla: My understanding about the language re: backdating is that it was solely related to situations where the revocation reason code was being changed to keyCompromise from some other reason - to say that was allowed. Does that clarification help us navigate a path forward on this issue?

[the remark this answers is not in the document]
aarongable: Confirming Ben's understanding: The paragraph this comment thread is on is copied from [here](https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/#6-revocation:~:text=Note%3A%20Backdating%20the%20revocationDate%20field%20is%20an%20exception%20to%20best%20practice%20described%20in%20RFC%205280%20(section%205.3.2)) in the MRSP. That version contains a hyperlink to [here](https://wiki.mozilla.org/CA/Revocation_Reasons#Updating_CRL_Entries) which provides some clarifying language.

docs/BR.md
```diff
@@ -3147,44 +3180,96 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
+When a CA obtains verifiable evidence of Key Compromise for a Certificate whose CRL entry does not contain a reasonCode extension or has a reasonCode extension with a non-keyCompromise reason, the CA SHOULD update the CRL entry to enter keyCompromise as the CRLReason in the reasonCode extension. 
+
+#### 7.2.2.1 CRL Issuing Distribution Point
+Partitioned CRLs MUST include at least one of the names from the corresponding distributionPoint field of the cRLDistributionPoints extension of every certificate that is within the scope of this CRL. The encoded value MUST be byte-for-byte identical to the encoding used in the distributionPoint field of the certificate. 
+
```
wthayer, line 3263: What about CAs that shard CRLs but do not include the cRLDP in the certs? I don't think the intent is to prohibit that but this language seems to have that effect? Can we add something like 'or at least one of the URLs published in CCADB'? Also, in either case is there a chicken/egg problem here if CRLs are issued before certs or before CCADB is updated?

docs/BR.md
```diff
@@ -3147,44 +3180,96 @@ The following Certificate Policy identifiers are reserved for use by CAs as an o
...
-  * **cessationOfOperation (RFC 5280 CRLReason #5):** Indicates that the website with the Certificate is shut down prior to the expiration of the Certificate, or if the Subscriber no longer owns or controls the Domain Name in the Certificate prior to the expiration of the Certificate; or
-  * **privilegeWithdrawn (RFC 5280 CRLReason #9):** Indicates that there has been a subscriber-side infraction that has not resulted in keyCompromise, such as the Certificate Subscriber provided misleading information in their Certificate Request or has not upheld their material obligations under the Subscriber Agreement or Terms of Use.
+| __RFC 5280 reasonCode__   | __RFC 5280 reasonCode value__ | __Description__ |
+| ---                       | -    | ------          |
+| unspecified               | 0    |  MUST be omitted if the CRL entry is for a Certificate not technically capable of causing issuance unless the CRL entry is for a Subscriber Certificate subject to these Requirements revoked prior to July 15, 2023.
```
wthayer, line 3247: Nit, but I had to read this language a few times to get comfortable with it. I think a statement like 'the unspecified reason is represented by the omission of a reasonCode.' would add clarity.

docs/BR.md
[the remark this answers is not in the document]
ryancdickson: @aarongable and @dzacharo - any issues with [this](https://github.com/ryancdickson/staging/commit/eb974081f456e73e68fd6df5382efabf4bc898f4) proposed update?

docs/BR.md
ryancdickson: It was a typo on my part dating back to the [early versions](https://github.com/ryancdickson/staging/blob/36bd26adf1346fcb9ff23e43468501778a7d3d7c/docs/BR.md#:~:text=When%20present%2C%20the%20CRL%20Distribution%20Points%20extension%20MUST%20contain%20at%20least%20one%20DistributionPoint%3B%20containing%20more%20than%20one%20is%20NOT%20RECOMMENDED.%20All%20DistributionPoint%20items%20must%20be%20formatted%20as%20follows%3A) of the proposal.

docs/BR.md
```diff
@@ -1280,27 +1286,39 @@ No stipulation.
...
+- a full and complete CRL; OR
+- partitioned (i.e., "sharded") CRLs that, when aggregated, represent the equivalent of a full and complete CRL.
 
-For the status of Subordinate CA Certificates:
+CAs issuing Subscriber Certificates:  
+1. MUST update and publish a new CRL within at least: 
+     - seven (7) days if all Certificates include an Authority Information Access extension with an id-ad-ocsp accessMethod (“AIA OCSP pointer”); or
+     - four (4) days in all other cases; 
+2. MUST update and publish a new CRL within twenty-four (24) hours after recording a Certificate as revoked; and
+3. MUST include a `nextUpdate` field value that is no more than ten (10) days beyond the value of the `thisUpdate` field.
```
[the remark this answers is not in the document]
ryancdickson: I appreciate that perspective. 

I originally kept the language in 4.9.7 for consistency with the existing location in the BRs. However, now that this requirement is addressed in the CRL profile, the redundancy may create more headache than it's worth (i.e., if the nextUpdate period were to change in the future, we'd need to address it in two sections).

Addressed in: https://github.com/ryancdickson/staging/commit/96153723ca783676a1b1f0316bc8610435b6c79b

docs/BR.md
ryancdickson: Addressed in https://github.com/ryancdickson/staging/commit/bbf5155dd87de476eb360a79feb8e49fdd0eefe6.

docs/BR.md
ryancdickson: Addressed in https://github.com/ryancdickson/staging/commit/b2df5dc24cdb9333ffa166edfe10c09445b3eeff.

docs/BR.md
ryancdickson: Addressed in https://github.com/ryancdickson/staging/commit/bbf5155dd87de476eb360a79feb8e49fdd0eefe6.

docs/BR.md
ryancdickson: Addressed in https://github.com/ryancdickson/staging/commit/530bc887dfd9553d8a77e1de5c5f20e1ed9d67b1

The "Description" column now reads: "_Represented by the omission of a reasonCode. MUST be omitted if the CRL entry is for a Certificate not technically capable of causing issuance unless the CRL entry is for a Subscriber Certificate subject to these Requirements revoked prior to July 15, 2023._"

@wthayer, does this work for you?

wthayer: LGTM

docs/BR.md
wthayer: My fundamental concern is that someone could come along in the future and interpret this section as an implicit requirement to include the cRLDP extension in certs when a CA is issuing sharded CRLs. Perhaps an easy clarification that aligns with your interpretation of the language is to prepend "If certificates contain a distributionPoints field and that name points to a partitioned CRL,..."

ryancdickson: Fair point. Re-phrased, with that concern in mind:

"_If a Certificate contains a distributionPoint field that does not point to the URI of a full and complete CRL, the corresponding CRL MUST contain the issuingDistributionPoint extension (OID 2.5.29.28). The encoded value of the issuingDistributionPoint extension MUST be byte-for-byte identical to the encoding used in the distributionPoint field of the Certificate._"

Any concerns with this approach?

wthayer: LGTM

ryancdickson: Thanks @wthayer!

CC @aarongable and @CBonnell - as I believe during past discussions you both had opinions on the language re: IDP. Any concerns with the above change? Candidly, I'd like this next round of discussion to be the last, so the more opinions, the better.

CBonnell: I understand the intent behind the change, but the concern I have with the proposed language is that other GeneralNames cannot be included in the IDP extension (as the value of the extension itself must be binary equal to the single distributionPoint).

I suggest:

Partitioned CRLs MUST contain an Issuing Distribution Point extension. The `distributionPoint` field of the Issuing Distribution Point extension MUST be present. Additionally, the `fullName` field of the DistributionPointName value MUST be present, and its value MUST conform to the following requirements:

1. If a Certificate within the scope of the CRL contains a CRL Distribution Points extension, then at least one of the `uniformResourceIdentifier`s in the CRL Distribution Points's `fullName` field MUST be included in the `fullName` field of the CRL's Issuing Distribution Point extension. The encoding of the `uniformResourceIdentifier` value in the Issuing Distribution Point extension SHALL be identical to the encoding used in the Certificate's CRL Distribution Points extension.
2. Other GeneralNames of type `uniformResourceIdentifier` MAY be included.
3. Non-`uniformResourceIdentifier` GeneralName types MUST NOT be included.

ryancdickson: Thanks, @CBonnell. I think that works and still addresses @wthayer's earlier concern.

Aside from one small change (using "byte-for-byte" in lieu of "octet-for-octet" --- to avoid introducing a new term), I believe I [incorporated](https://github.com/ryancdickson/staging/commit/15eab80d50a7b6b0e5dce7d413c7692610d4ac93) your feedback "as-is." 

I made some other small tweaks to improve consistency in how extensions introduced in this update are represented.

docs/BR.md
```diff
@@ -1,10 +1,9 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates
```
Pongsaphat17, line 2: ryancdickson:make-ocsp-optional

PR #440, opened 2023-07-14
SC65: Convert EVGs into RFC 3647 format v2

barrini:
The Extended Validation Certificates guidelines (EVGs) were developed and written in a specific format. Since then, the RFC 3647 has been the basis (and the de-facto standard) for the CA/Browser Forum to develop other documents.

This ballot aims to update the EVGs to follow the RFC 3647 format without changing any content, just moving current sections to those defined in the RFC 3647. There are no normative requirements changes.

This change also affects the Baseline Requirements for TSL certificates (BRs) which needs to point to the new sections of the EVGs. Both documents will be updated according to the latest version published.


ryancdickson:
@barrini - will the scope of the future ballot focus strictly on aligning with the format of 3647, or will other updates be considered?

Issue #453, opened 2023-08-24
Clarify CA Subscribers


BenWilson-Mozilla:
Modified from https://lists.cabforum.org/pipermail/validation/2022-November/001826.html:

Here is draft language that could be added to BR Section 1.3.3, Subscribers.

In some situations, a CA acts as an Applicant or Subscriber, for instance, when it generates and protects a Private Key, requests a Certificate, demonstrates control of a Domain, or obtains a Certificate for its own use.

BenWilson-Mozilla:
Perhaps this language might make it more clear that a CA must store validation records under BR section 5.4.1 when it is also the Subscriber.

barrini:
to include in the next clean-up ballot

Issue #450, opened 2023-08-15
Clarify "specified above" in section 6.2


wthayer:
The second sentence of BR section 6.2 states: Protection of the CA Private Key outside the validated system or device specified above MUST consist of physical security, encryption, or a combination of both, implemented in a manner that prevents disclosure of the Private Key.

It's not clear to me what "the validated system or device specified above" is referring to. Looking back 8 years, I see that this sentence used to be in section 16.6 and was preceded by the statement "The CA SHALL protect its Private Key in a system or device that has been validated as meeting at least FIPS 140 level 3 or an appropriate Common Criteria Protection Profile or Security Target, EAL 4 (or higher), which includes requirements to protect the Private Key and other assets against known threats." When the BRs were restructured to RFC 3647 format, that sentence was moved below (section 6.2.7).

I believe the fix is to simply replace "specified above" with "specified in section 6.2.7"

barrini:
to include in the next clean-up ballot

Issue #446, opened 2023-07-25
IP validation via ACME


timfromdigicert:
The BRs reference a draft RFC for IP validation (https://tools.ietf.org/html/draft‐ietf‐acme‐ip‐04#section‐4).  That draft has since been published as RFC 8738.  We should look if the reference can be updated to point to the published RFC instead of the draft.

barrini:
to include in the next clean-up ballot

barrini:
Reflected in #169 
Already included in the clean-up

Issue #444, opened 2023-07-20
BR v2.0.0 Missing Section 7.1.5


AaronP-ATS:
Section 7.1.5 was removed with BR v2.0.0. It should remain as an empty section 1) to ensure contiguous numbering between sections, and 2) to stay compliant with its own assertion that all RFC 3647 sections are included.

From v2.0.0 **1.1 Overview**:

> In accordance with RFC 3647 and to facilitate a comparison of other certificate policies and CPSs (e.g. for policy mapping), this document includes all sections of the RFC 3647 framework.

barrini:
Thanks Aaron, I´ve labelled this issue as "clean-up" if you don´t mind and therefore be included in the next clean-up ballot.

barrini:
to include in the next clean-up ballot

BenWilson-Mozilla:
A cross-reference should be added to BR section 7.1.5 to point readers to where in the BRs we discuss name constraints.

dzacharo:
Addressed in SC095

PR #460, opened 2023-09-13
Fall 2023 clean up

barrini:
Issues included are:
#169 and #446 
#174
#337
#415 
#423 
#430 
#444 
#450 
#453


docs/BR.md
```diff
@@ -613,7 +619,7 @@ Section 4.2 of a CA's Certificate Policy and/or Certification Practice Statement
 
 The CA SHALL publicly give effect to these Requirements and represent that it will adhere to the latest published version. The CA MAY fulfill this requirement by incorporating these Requirements directly into its Certificate Policy and/or Certification Practice Statements or by incorporating them by reference using a clause such as the following (which MUST include a link to the official version of these Requirements):
 
-> [Name of CA] conforms to the current version of the Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates published at <http://www.cabforum.org>. In the event of any inconsistency between this document and those Requirements, those Requirements take precedence over this document.
+> [Name of CA] conforms to the current version of the Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates published at <http://www.cabforum.org>. In the event of any inconsistency between this document and those Requirements, those Requirements take precedence over this document.
```
CBonnell, line 622: I *think* the document name change and associated change in the statement of compliance need not be immediately reflected in all CP/CPS documents upon this ballot's passage, since the statement of compliance need not be a verbatim copy of this text.

Wondering if anyone (especially auditors) might have a different opinion.

docs/BR.md
```diff
@@ -957,11 +963,11 @@ The Random Value SHALL remain valid for use in a confirming response for no more
 
 ##### 3.2.2.5.6 ACME “http-01” method for IP Addresses
 
-Confirming the Applicant's control over the IP Address by performing the procedure documented for an “http-01” challenge in draft 04 of “ACME IP Identifier Validation Extension,” available at <https://tools.ietf.org/html/draft-ietf-acme-ip-04#section-4>.
+Confirming the Applicant's control over the IP Address by performing the procedure documented for an “http-01” challenge in RFC 8738.
```
CBonnell, line 966: The only potential normative requirement change I see with this is that the RFC specifies a particular format for IPv4 and IPv6 addresses in the Host header, whereas the -04 draft does not: https://author-tools.ietf.org/iddiff?url1=draft-ietf-acme-ip-04&url2=rfc8738&difftype=--html.

I doubt this is material though, since implementations will already be using the RFC-prescribed format.

docs/BR.md
```diff
@@ -2938,7 +2944,7 @@ Each `SignedCertificateTimestamp` included within the `SignedCertificateTimestam
 
 ##### 7.1.2.11.4 Subject Key Identifier
 
-If present, the `subjectKeyIdentifier` MUST be set as defined within [RFC 5280, Section 4.2.1.2](https://tools.ietf.org/html/rfc5280#section-4.2.1.2). The CA MUST generate a `subjectKeyIdentifier` that is unique within the scope of all Certificates it has issued for each unique public key (the `subjectPublicKeyInfo` field of the `tbsCertificate`). For example, CAs may generate the subject key identifier using an algorithm derived from the public key, or may generate a sufficiently-large unique number, such by using a CSPRNG.
+If present, the `subjectKeyIdentifier` MUST be set as defined within [RFC 5280, Section 4.2.1.2](https://tools.ietf.org/html/rfc5280#section-4.2.1.2). The CA MUST generate a `subjectKeyIdentifier` that is unique within the scope of all Certificates it has issued for each unique public key (the `subjectPublicKeyInfo` field of the `tbsCertificate`). For example, CAs may generate the subject key identifier using an algorithm derived from the public key, or may generate a sufficiently-large unique number, by using a CSPRNG.
```
CBonnell, line 2947:
```suggestion
If present, the `subjectKeyIdentifier` MUST be set as defined within [RFC 5280, Section 4.2.1.2](https://tools.ietf.org/html/rfc5280#section-4.2.1.2). The CA MUST generate a `subjectKeyIdentifier` that is unique within the scope of all Certificates it has issued for each unique public key (the `subjectPublicKeyInfo` field of the `tbsCertificate`). For example, CAs may generate the subject key identifier using an algorithm derived from the public key, or may generate a sufficiently-large unique number, such as by using a CSPRNG.
```

There is no need to require the use of a CSPRNG, as randomness of the SKI provides no security value.

docs/BR.md
barrini: Agree. Missed that one.

docs/BR.md
barrini: Also agree. Same as above

castillar (approved):
This looks good!

dougbeattie (approved):
Looks good to me!

PR #470, opened 2023-12-21
Ballot SC-XX: Measure all hours and days to the second

aarongable:
In light of https://bugzilla.mozilla.org/show_bug.cgi?id=1865080, this ballot ensures that all readers of the BRs understand that time periods measured in days (such as validation document reuse periods, random value usage periods, and revocation timelines) are measured precisely, not in calendar days.

Notes:
- This ballot bears some similarity to Ballot SC-52, which never came to a vote.
- This ballot does not strictly define a "month", allowing infrequent tasks to continue to be executed on the same numeric day of each month, regardless of the number of days in that month.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -460,7 +460,7 @@ The script outputs:
 
 **Root Certificate**: The self-signed Certificate issued by the Root CA to identify itself and to facilitate verification of Certificates issued to its Subordinate CAs.
 
-**Short-lived Subscriber Certificate**: For Certificates issued on or after 15 March 2024 and prior to 15 March 2026, a Subscriber Certificate with a Validity Period less than or equal to 10 days (864,000 seconds). For Certificates issued on or after 15 March 2026, a Subscriber Certificate with a Validity Period less than or equal to 7 days (604,800 seconds).
+**Short-lived Subscriber Certificate**: For Certificates issued on or after 15 March 2024 and prior to 15 March 2026, a Subscriber Certificate with a Validity Period less than or equal to 10 days. For Certificates issued on or after 15 March 2026, a Subscriber Certificate with a Validity Period less than or equal to 7 days.
 
 **Sovereign State**: A state or country that administers its own government, and is not dependent upon, or subject to, another power.
 
@@ -594,6 +594,8 @@ X.509, Recommendation ITU-T X.509 (08/2005) \| ISO/IEC 9594-8:2005, Information 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in these Requirements shall be interpreted in accordance with RFC 2119.
 
 By convention, this document omits time and timezones when listing effective requirements such as dates. Except when explicitly specified, the associated time with a date shall be 00:00:00 UTC.
+
+For purposes of measuring periods of time, one hour shall be defined to be exactly 3,600 seconds, and one day shall be defined to be exactly 86,400 seconds, ignoring leap-seconds. Any amount of time greater than this, including fractional seconds, shall represent an additional unit of measure, such as an additional hour or an additional day.
 
 # 2. PUBLICATION AND REPOSITORY RESPONSIBILITIES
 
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -1772,7 +1776,7 @@ The CA SHALL protect its Private Key in a system or device that has been validat
 
 Subscriber Certificates issued on or after 1 September 2020 SHOULD NOT have a Validity Period greater than 397 days and MUST NOT have a Validity Period greater than 398 days. 
 
-For the purpose of calculations, a day is measured as 86,400 seconds. Any amount of time greater than this, including fractional seconds and/or leap seconds, shall represent an additional day. For this reason, Subscriber Certificates SHOULD NOT be issued for the maximum permissible time by default, in order to account for such adjustments.
+Due to the precision with which Certificate validity periods are measured, Subscriber Certificates SHOULD NOT be issued for the maximum permissible time by default, to prevent off-by-one-second errors.
```
dzacharo, line 1779:
```suggestion
Due to the precision with which Certificate Validity Period is measured, Subscriber Certificates SHOULD NOT be issued for the maximum permissible time by default, to prevent off-by-one-second errors.
```

docs/BR.md
```diff
@@ -490,6 +490,8 @@ The script outputs:
 
 **Validation Specialist**: Someone who performs the information verification duties specified by these Requirements.
 
+**Validity Interval**: For CRLs and OCSP Responses, the period of time from thisUpdate through nextUpdate, inclusive.
```
dzacharo, line 493: In order to avoid confusion, perhaps we could rename the "Validity Period" to "Certificate Validity Period" since it seems to apply only to Certificates, and then introduce "Validity Interval" that focuses on CRL and OCSP Responses.

docs/BR.md
```diff
@@ -595,6 +597,10 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 By convention, this document omits time and timezones when listing effective requirements such as dates. Except when explicitly specified, the associated time with a date shall be 00:00:00 UTC.
 
+All statements of time periods (for example, "5 days") SHALL be taken to mean exactly that period of time, and not a fractional period more (e.g. not "5 and a half days").
```
dzacharo, line 600: I'm afraid this will extend beyond the definition of "days" and "hours" within the BRs, EVGs and NSRs, and cause some confusion. For example:
- NSR 1l: "within six (6) months"
- NSR 2j: "every three (3) months"
- NSR 2g: "two years"
- NSR 4c: "every three (3) months"
- BR 4.9.7: "every twelve (12) months"
- BR 5.4.3: "two (2) years"
- BR 5.5.2: "two (2) years"
- BR 8.1: "one year"
- BR 8.6: "three months"
- EV 11.2.2: "six months"
- EV 14.1.1: "three months"

I'm not sure how to expand this example into the time periods of the cases above. If we want more precision on any time period that is not stated as "X number of days" or "X hours" in the documents, we probably need to update those time periods and call out specific days. 

For example, NSR 3e states "once every 31 days", which covers months that have 30 days or February that has 28 or 29 days. Similarly, we could change 3 months to 92 days and so on.

docs/BR.md
aarongable: Hmm, interesting idea. I kinda like it, but I also don't want to rename "Validity Period", because that defined term (and its definition) is taken directly from RFC 5280.

dzacharo: We could just add the word "Certificate" in front of it, which is already the subject of the term, so in essence you are not changing the definition :)
In any case, I don't have any strong feelings about this.

docs/BR.md
aarongable: On the whole, I agree. I would love for all time periods throughout these documents to be measured in days. But the last time I advocated for that, I received pushback from folks who think that having different granularities is useful.

I think this will resolve more confusion than it will cause.

Right now, someone could read EV 14.1.1 "three months" and interpret that as over 100 days: "We conducted the verification on Jan 1, it's now April 15, and April is three months after January". In contrast, I think it would be very very difficult for someone to interpret "three months" as (say) 28*3=84 days, since there is no point in the calendar in which there are three 28-day months.

My original plan here was to apply this only to "hours" and "days", as noted by the draft title. But during the validation call, others expressed that they'd prefer that we treat all periods of time equally, and not clarify hours/days differently from months/years. I'm okay doing it either way, but the community will have to come to consensus.

timfromdigicert:
Oh, please no.

 

I don’t want to have to deal with all the “fun” that will ensue if CABF defines “Certificate Validity Period” to be something other than the “Validity Period” from RFC 5280.  Terminology is confusing enough without unnecessary complications like that.

 

-Tim

Issue #274, opened 2021-05-04
BRs: Clarify whether 3.2.2.8 can be delegated


sleevi:
CA/B Forum [Ballot 187](https://cabforum.org/2017/03/08/ballot-187-make-caa-checking-mandatory/) made CAA checking Mandatory for CAs, a revision of [Ballot 125](https://cabforum.org/2014/10/14/ballot-125-caa-records/) which had CAs document their CAA policies.

CA/B Forum [Ballot 204](https://cabforum.org/2017/07/11/ballot-204-forbid-dtps-domainip-ownership/), adopted 4 months after, forbade CAs from delegating the performance of 3.2.2.4 and 3.2.2.5 (DNS and IP validation).

@CBonnell raised on the management list that this suggests that CAs are allowed to delegate part, or all, of the CAA validation to third parties, as this is located within section 3.2.2.8, which Ballot 204 does not speak to.

Ballot 187 is structured in a way to describe when a CA may choose to **not** check CAA (i.e. if doing as part of a precert/final cert issuance and the precert was checked and logged, if issuing from a technically constrained sub CA, or if an affiliate of the CA, notwithstanding [draft ballot SC26](https://github.com/cabforum/servercert/pull/271)). However, it doesn't explicitly address the delegation to a third-party (e.g. a non-technically constrained sub-CA) and how that performance is measured.

Given the critical security importance of 3.2.2.8, which was the reason and rationale for not allowing 3.2.2.4/3.2.2.5 to be delegated, it seems important to clarify whether or not 3.2.2.8 can or should be delegated to third parties.

barrini:
Let´s wait for PAG outcome

PR #475, opened 2024-01-18
Ballot SC-070: Clarify the use of DTPs for domain control validation

aarongable:
Add a paragraph to BRs Section 3.2.2 clarifying that all Domain Control Validation DNS queries must be conducted by the CA itself, without the use of third-party recursive resolvers. Similarly clarify that looking up information for Domain Contacts and IP Address Contacts must also be done without third-party services.

Require that CAA checks be performed by the CA itself, and not delegated to a third party.

Fixes https://github.com/cabforum/servercert/issues/274

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -199,7 +199,7 @@ Certification Authority (CA) is defined in [Section 1.6](#16-definitions-and-acr
 
 ### 1.3.2 Registration Authorities
 
-With the exception of [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) and [Section 3.2.2.5](#3225-authentication-for-an-ip-address), the CA MAY delegate the performance of all, or any part, of [Section 3.2](#32-initial-identity-validation) requirements to a Delegated Third Party, provided that the process as a whole fulfills all of the requirements of [Section 3.2](#32-initial-identity-validation).
+With the exception of [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control), [Section 3.2.2.5](#3225-authentication-for-an-ip-address), and [Section 3.2.2.8](#3228-caa-records) the CA MAY delegate the performance of all, or any part, of [Section 3.2](#32-initial-identity-validation) requirements to a Delegated Third Party, provided that the process as a whole fulfills all of the requirements of [Section 3.2](#32-initial-identity-validation).
 
 Before the CA authorizes a Delegated Third Party to perform a delegated function, the CA SHALL contractually require the Delegated Third Party to:
 
@@ -334,7 +334,7 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **DNS TXT Record Phone Contact**: The phone number defined in [Appendix A.2.2](#a22-dns-txt-record-phone-contact).
 
-**Domain Contact**: The Domain Name Registrant, technical contact, or administrative contact (or the equivalent under a ccTLD) as listed in the WHOIS record of the Base Domain Name or in a DNS SOA record, or as obtained through direct contact with the Domain Name Registrar.
+**Domain Contact**: The Domain Name Registrant, technical contact, or administrative contact (or the equivalent under a ccTLD) for a Base Domain Name.
 
 **Domain Label**: From RFC 8499 (<http://tools.ietf.org/html/rfc8499>): "An ordered list of zero or more octets that makes up a portion of a domain name. Using graph theory, a label identifies one node in a portion of the graph of all possible domain names."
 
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -704,6 +704,8 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 1. When the FQDN is not an Onion Domain Name, the CA SHALL validate the FQDN using at least one of the methods listed below; and
 2. When the FQDN is an Onion Domain Name, the CA SHALL validate the FQDN in accordance with Appendix B.
 
+All DNS queries conducted in the course of validation MUST be made from the CA to authoritative nameservers, i.e. without the use of recursive resolvers operated outside the CA's audit scope.
```
XolphinMartijn, line 707: Could we move this to 3.2.2 instead, and rephrase as:

"All DNS queries conducted in the course of validation in Section 3.2.2.4, Section 3.2.2.5 and Section 3.2.2.8 MUST be made from the CA to authoritative nameservers, i.e. without the use of recursive resolvers operated outside the CA's audit scope."

MadZeg: Today, a CA may use an external DNS resolver for the purpose of checking CAA according to 3.2.2.8 as long as this is done by a DTP (and according to the requirements for using a DTP) . Do we really want to change this? I think this is out-of-scope for this ballot. See BRF 1.3.2 for details - we should restrict the ballot to 3.2.2.4 and 3.2.2.5.

aarongable (commented):
Feedback from Validation subcommittee meeting 2024-01-25:
- "All DNS queries conducted in the course of..." sounds like it could include, say, DNS queries made by a validation specialist's browser while looking up the requirements they have to abide by. One suggestion is "All DNS queries _required to be_ conducted in the course of...", but this then perhaps exempts DNS SOA lookups to get Domain Contact information because those are technically optional (could be replaced by WHOIS lookups).
- The phrasing of "from the allowed sources" doesn't make it immediately clear that those allowed sources are listed in the Section 1.6 definition of Domain Contact.
- The sentences added to 3.2.2.4 and 3.2.2.5 don't cover CAA lookups, which are described in 3.2.2.8. We need to ensure that CAA is covered, and probably end up resolving https://github.com/cabforum/servercert/issues/274 at the same time.

aarongable:
> * "All DNS queries conducted in the course of..." sounds like it could include, say, DNS queries made by a validation specialist's browser while looking up the requirements they have to abide by. One suggestion is "All DNS queries _required to be_ conducted in the course of...", but this then perhaps exempts DNS SOA lookups to get Domain Contact information because those are technically optional (could be replaced by WHOIS lookups).

Addressed in the latest commit by saying "in the course of satisfying the requirements of 3.2.2.4, 3.2.2.5, and 3.2.2.8".

> * The phrasing of "from the allowed sources" doesn't make it immediately clear that those allowed sources are listed in the Section 1.6 definition of Domain Contact.

Addressed in the latest commit by modifying the definition of Domain Contact to say what one _is_ rather than where the info comes from, and explicitly laying out the requirements for how to obtain a Domain Contact's info in Section 3.2.2.

> * The sentences added to 3.2.2.4 and 3.2.2.5 don't cover CAA lookups, which are described in 3.2.2.8. We need to ensure that CAA is covered, and probably end up resolving [BRs: Clarify whether 3.2.2.8 can be delegated #274](https://github.com/cabforum/servercert/issues/274) at the same time.

Addressed in the latest commit by moving the verbiage to Section 3.2.2, and explicitly calling out 3.2.2.8 as one of the covered sections.

docs/BR.md
MadZeg: I understand and agree, then it makes sense to include CAA lookups in this ballot.

docs/BR.md
```diff
@@ -199,7 +199,7 @@ Certification Authority (CA) is defined in [Section 1.6](#16-definitions-and-acr
 
 ### 1.3.2 Registration Authorities
 
-With the exception of [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) and [Section 3.2.2.5](#3225-authentication-for-an-ip-address), the CA MAY delegate the performance of all, or any part, of [Section 3.2](#32-initial-identity-validation) requirements to a Delegated Third Party, provided that the process as a whole fulfills all of the requirements of [Section 3.2](#32-initial-identity-validation).
```
dzacharo, line 202: Although this seems to have been implicitly required, if we are to make it explicit, we probably will need some sort of effective date (something like ~30 days after the IPR period clears).

dzacharo (commented):
Changes look good to me. Please check the comment about an effective date for 3.2.2.8.

docs/BR.md
aarongable: I've added an effective date of 2024-05-15, expected to be between 1 and 2 months after the IPR period and nicely aligned with the "middle of odd months" idea.

Issue #466, opened 2023-11-29
Unify sections regarding CAA


aarongable:
Currently:

- Section 2.2 states
> Section 4.2 of a CA's Certificate Policy and/or Certification Practice Statement SHALL state the CA's policy or practice on processing CAA Records for Fully-Qualified Domain Names; that policy shall be consistent with these Requirements. It shall clearly specify the set of Issuer Domain Names that the CA recognizes in CAA "issue" or "issuewild" records as permitting it to issue. The CA SHALL log all actions taken, if any, consistent with its processing practice.

- Section 3.2.2.8 is entitled "CAA Records" and contains numerous requirements regarding CAA checking

- And Section 4.2 has nothing to say about CAA, despite the fact that CA's are required to document their own CAA-checking practices here, as per Section 2.2.

These should be unified. In my opinion, the paragraph in Section 2.2 should be removed, and anything we want to keep from it should be moved into Section 3.2.2.8. Then CAs can similarly move all of their CAA practices into Section 3.2.2.8 of their CP/CPS.

ryancdickson:
Added this to the scope of the Multi-Perspective Issuance Corroboration work thanks to the [reminder](https://github.com/ryancdickson/staging/pull/8#discussion_r1408590704) from @aarongable.

Updated language can be found [here](https://github.com/ryancdickson/staging/blob/require-mpdv-v2/docs/BR.md#3228-caa-records) via these commits: https://github.com/ryancdickson/staging/pull/8/commits/0525044a051bad888caf1072111a8c5717fe1cfa
and https://github.com/ryancdickson/staging/commit/8315b6a3a338ea6b47cf99979b0d97f61e5f3363

Summary:
- Deleted the CAA paragraph from 2.2
- Added what I felt was relevant content from 2.2 into 3.2.2.8
- Did some minor housekeeping

@bcmorton / @srdavidson - your review and feedback are welcome, along with other members of the community!

srdavidson:
Here's the S/MIME BR (SBR) equivalent.  https://github.com/cabforum/smime/pull/228/file
For issuers, it's preferable have similar requirements in similar sections.
SBR has no 3.2.2.8 so its CAA content is is 4.2.2.1 ... however both the SBR and this text require documentation of CAA practices in section 4.2 so that should be fine.

barrini:
@aarongable can you confirm that we can close this issue?

PR #487, opened 2024-02-22
SC-067: Require Multi-Perspective Issuance Corroboration (Version 1)

ChristopherRC:
**Summary**: This is the first version of SC-067, which intends to add "Multi-Perspective Domain Validation" and "Multi-Perspective CAA Checking" (together referred to as “Multi-Perspective Issuance Corroboration”) requirements to the CA/Browser Forum TLS Server Certificate Baseline Requirements. Official public discussion is expected to begin the week after CA/Browser Forum F2F 61.

<br>

**This Pull Request:**
- compares the latest draft of SC-067 against Version 2.0.2 of the TLS Server Certificate Baseline Requirements.
- addresses issues and comments made against draft Version 2 (pre-ballot) ([clean](https://github.com/ryancdickson/staging/blob/require-mpdv-v2/docs/BR.md)) of this effort. A separate branch was created to help make changes across versions of the document clear to readers, and due to owner of the earlier versions being on leave.

<br>

**How can you help?**
- Better: Add comments to this Pull Request.
- _Best_: Add suggested edits directly to this Pull Request.

<br>

**Version History:**
- Pre-Ballot Version 1 [ [branch](https://github.com/ryancdickson/staging/tree/require-mpdv-v1), [compare against SC-63 v3 (obsolete)](https://github.com/ryancdickson/staging/pull/6)]
- Pre-Ballot Version 2 [ [branch](https://github.com/ryancdickson/staging/tree/require-mpdv-v2), [compare against SC-63 v3 (obsolete)](https://github.com/ryancdickson/staging/pull/8)]

<br>

**Summary of recent updates:**
- Establishes requirements for the use of recursive DNS resolvers for remote Network Perspectives following discussions related to Ballot SC-070.
- Adds flexibility regarding how Network Perspectives may be operated to corroborate domain control validation.
- Permits re-use of MPIC CAA quorum for subdomains under certain conditions.
- Clarifies that Onion Names are outside of scope and exempt from MPIC requirements.
- Moves CAA-related requirements from Section 2.2 into Section 4.2.2.8 to address issue [466](https://github.com/cabforum/servercert/issues/466).
- Shifts back required implementation dates.

<br>

**Additional Notes:**
- Some changes in this PR represent clean-ups (e.g., smart quotes -> regular quotes) unrelated to the primary scope of this Ballot (Multi-Perspective Issuance Corroboration).
- "MPDV Work Team" [Work Plan](https://docs.google.com/document/d/1Vg8qwH_slheGofmj-lKt0P-bMNezlML3n89g8dSVSGo/edit) (contains useful background and additional context).
- Previous Validation Subcommittee [Update](https://drive.google.com/file/d/1LTwtAwHXcSaPVSsqKQztNJrV2ozHJ7ZL/view?usp=drive_link) introducing this work.

<br>

**Additional Resources:**

On MPDV:
- [How Effective is Multiple-Vantage-Point Domain Control Validation?](https://arxiv.org/abs/2302.08000)
- [Experiences Deploying Multi-Vantage-Point Domain Validation at Let’s Encrypt.](https://www.usenix.org/conference/usenixsecurity21/presentation/birge-lee)
- [Let's Downgrade Let's Encrypt](https://dl.acm.org/doi/10.1145/3460120.3484815)
- [Domain Validation++ For MitM-Resilient PKI](https://dl.acm.org/doi/10.1145/3243734.3243790)
- APIs:
     - Cloudflare: [Multipath DCV service](https://blog.cloudflare.com/secure-certificate-issuance)  [API](https://docs.google.com/document/d/19wvjk7lcK1TCQpJrjEljosTEe8A0We1ayRp_1Ou3r4s/edit#heading=h.9kf5j5tsn6i7)
     - Princeton: [Open MPIC Project](https://freedom-to-tinker.com/2024/02/13/announcing-the-open-multi-perspective-issuance-corroboration-project/)

On the problem space:
- A [video](https://drive.google.com/file/d/15e4Z9InYbThwJsDuH0oS7vfXKvdSBzi9/view) describing the vulnerabilities these requirements intend to prevent.
- [Attackers exploit fundamental flaw in the web’s security to steal $2 million in cryptocurrency](https://freedom-to-tinker.com/2022/03/09/attackers-exploit-fundamental-flaw-in-the-webs-security-to-steal-2-million-in-cryptocurrency/)
- [Celer Bridge incident analysis](https://www.coinbase.com/blog/celer-bridge-incident-analysis)
- [Bamboozling Certificate Authorities with BGP](https://www.usenix.org/conference/usenixsecurity18/presentation/birge-lee)
- [Face-to-Face #58](https://cabforum.org/wp-content/uploads/13-CAB-Forum-face-to-face-multiple-vantage-points.pdf) (Presentation from Princeton Team)
- [Securing Internet Applications from Routing Attacks](https://cacm.acm.org/magazines/2021/6/252822-securing-internet-applications-from-routing-attacks/fulltext?mobile=false)


docs/BR.md
```diff
@@ -380,6 +382,10 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **Legal Entity**: An association, corporation, partnership, proprietorship, trust, government entity or other entity with legal standing in a country's legal system.
 
+**Multi-Perspective Issuance Corroboration**: A process by which the determinations made during domain validation and CAA checking by the Primary Network Perspective are corroborated from other Network Perspectives before Certificate issuance.
```
XolphinMartijn, line 385:
```suggestion
**Multi-Perspective Issuance Corroboration**: A process by which the determinations made during domain validation and CAA checking by the Primary Network Perspective are corroborated by other Network Perspectives before Certificate issuance. 
```

docs/BR.md
```diff
@@ -380,6 +382,10 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **Legal Entity**: An association, corporation, partnership, proprietorship, trust, government entity or other entity with legal standing in a country's legal system.
 
+**Multi-Perspective Issuance Corroboration**: A process by which the determinations made during domain validation and CAA checking by the Primary Network Perspective are corroborated from other Network Perspectives before Certificate issuance. 
+
+**Network Perspective**: Related to Multi-Perspective Issuance Corroboration. A system for sending outbound Internet traffic associated with a domain control validation method and CAA check. The location of a Network Perspective is determined by the point where unencapsulated outbound Internet traffic is first handed off to the network infrastructure providing Internet connectivity to that perspective.
```
XolphinMartijn, line 387:
```suggestion
**Network Perspective**: Related to Multi-Perspective Issuance Corroboration. A system for sending outbound Internet traffic associated with a domain control validation method and/or CAA check. The location of a Network Perspective is determined by the point where unencapsulated outbound Internet traffic is first handed off to the network infrastructure providing Internet connectivity to that perspective.
```

docs/BR.md
```diff
@@ -994,25 +1026,88 @@ Databases maintained by the CA, its owner, or its affiliated companies do not qu
...
+
+#### 3.2.2.9 Multi-Perspective Issuance Corroboration
+
+Multi-Perspective Issuance Corroboration attempts to corroborate the determinations (i.e., domain validation pass/fail, CAA permission/prohibition) made by the Primary Network Perspective from multiple remote Network Perspectives before Certificate issuance. This process can improve protection against equally-specific prefix Border Gateway Protocol (BGP) attacks or hijacks.
+
+Each corroborating Network Perspective's response MUST provide the CA with the necessary information to allow it to affirmatively assess:
+- a. the presence of the expected 1) Random Value, 2) Request Token, 3) IP Address, or 4) Contact Address, as required by the relied upon validation method specified in Sections 3.2.2.4 and 3.2.2.5; and
+- b. the CA's authority to issue to the requested domain(s), as specified in Section 3.2.2.8.
+
```
XolphinMartijn, line 1059: I worry that this can be read as that every response from a network perspective MUST contain both the presense of the Random Value (or likewise for DCV), AND the CAA confirmation. 

I expect the intent here is to state that the CA MUST get a confirming response on both of these items prior to the issuance of certificate, however, I can imagine that CAs may want to do this by using two different calls, thus getting 2 different responses, one for DCV confirmation, one for CAA confirmation. 

```suggestion
Each corroborating Network Perspective's response(s) MUST provide the CA with the necessary information to allow it to affirmatively assess:
- a. the presence of the expected 1) Random Value, 2) Request Token, 3) IP Address, or 4) Contact Address, as required by the relied upon validation method specified in Sections 3.2.2.4 and 3.2.2.5; and
- b. the CA's authority to issue to the requested domain(s), as specified in Section 3.2.2.8.

The CA MAY use either the same set, or different sets of Network Perspectives for Domain Authorization or Control and CAA Record checks
```

XolphinMartijn: I'm not fully happy with the suggestion either, if it inspires someone for a better solution, please comment

docs/BR.md
```diff
@@ -994,25 +1026,88 @@ Databases maintained by the CA, its owner, or its affiliated companies do not qu
...
+The "Quorum Requirements" Table describes quorum requirements related to Multi-Perspective Issuance Corroboration. Network Perspectives are considered distinct when the straight-line distance between the two States, Provinces, or Countries they reside in is at least 500 km. Network Perspectives are considered "remote" when they are distinct from the Primary Network Perspective and the other Network Perspectives represented in a quorum.
+
+A CA MAY reuse corroborating evidence for CAA record quorum compliance for a maximum of 398 days. After issuing a Certificate to a domain, remote Network Perspectives MAY omit retrieving and processing CAA records for its subdomains in subsequent Certificate requests from the same Applicant for up to a maximum of 398 days.
```
XolphinMartijn, line 1070:
```suggestion
A CA MAY reuse corroborating evidence for CAA record quorum compliance for a maximum of 398 days. After issuing a Certificate to a domain, remote Network Perspectives MAY omit retrieving and processing CAA records for the same domain or its subdomains in subsequent Certificate requests from the same Applicant for up to a maximum of 398 days.
```

docs/BR.md
ChristopherRC: We have no issues using “by" in place of “from.” The original use of “from" was intended to describe the location *from* which MPIC lookups take place, but if this is clearer for the community - we’re happy to use it. 

We pre-staged this change in https://github.com/ChristopherRC/servercert/commit/07f9c3bac410a8af53ef60755b2788e67f7ab556 in preparation for a subsequent round of discussion.

docs/BR.md
ChristopherRC: This appears to add flexibility to how CA Owners implement MPIC, and we think that’s a good thing. Using specific Network Perspectives for either DCV or CAA checks could create opportunities for operational efficiencies, which we had no intention of limiting.

We pre-staged this change in https://github.com/ChristopherRC/servercert/commit/07f9c3bac410a8af53ef60755b2788e67f7ab556 in preparation for a subsequent round of discussion.

docs/BR.md
ChristopherRC: As described in response to your last comment, we think the added flexibility proposed in this update is positive. 

While we have no issues with the proposed language “The CA MAY use either the same set, or different sets of Network Perspectives for Domain Authorization or Control and CAA Record checks.”, we tweaked this a bit as described below. 

Do you feel the following adequately addresses your concern re: implementation flexibility, while also making it clear that both domain validation and CAA MPIC checks are expected? The bold language is modified, the rest of the italics are unchanged. 

_**The CA MAY use either the same set, or different sets of Network Perspectives when performing Multi-Perspective Issuance Corroboration for the required 1) Domain Authorization or Control and 2) CAA Record checks.**_

_**The set of responses from the relied upon Network Perspectives MUST provide the CA with the necessary information to allow it to affirmatively assess:**_

_a. the presence of the expected 1) Random Value, 2) Request Token, 3) IP Address, or 4) Contact Address, as required by the relied upon validation method specified in Sections 3.2.2.4 and 3.2.2.5; and
b. the CA's authority to issue to the requested domain(s), as specified in Section 3.2.2.8._

While making this change, it seems we need to consider the impact of the “and/or" approach to the Quorum Requirements Table. For example, it seems we’ll need to add something like the following bolded text to make it clear that if not using the same set of perspectives for both DV and CAA checks, two quorums are necessary. We will also do a more holistic review of the ballot text to verify whether other changes need to be made. 

_The "Quorum Requirements" Table describes quorum requirements related to Multi-Perspective Issuance Corroboration. **If the CA does NOT rely on the same set of Network Perspectives for both Domain Authorization or Control and CAA Record checks, the quorum requirements MUST be met for both sets of Network Perspectives (i.e.,the Domain Authorization or Control set and the CAA record check set).** Network Perspectives are considered distinct when the straight-line distance between the two States, Provinces, or Countries they reside in is at least 500 km. Network Perspectives are considered "remote" when they are distinct from the Primary Network Perspective and the other Network Perspectives represented in a quorum._

Do you (or others) have any thoughts on this approach?

We pre-staged these changes in https://github.com/ChristopherRC/servercert/commit/07f9c3bac410a8af53ef60755b2788e67f7ab556 in preparation for a subsequent round of discussion.

docs/BR.md
```diff
@@ -994,25 +1026,88 @@ Databases maintained by the CA, its owner, or its affiliated companies do not qu
...
+
+Table: Quorum Requirements
+
+| # of Distinct Remote Network Perspectives Used | # of Allowed non-Corroborations |
+| --- | --- |
+| 2-5 |  1  |
+| 6+  |  2  |
+
+Remote Network Perspectives performing Multi-Perspective Issuance Corroboration:
```
jdkasten, line 1079: I agree with the spirit of this requirement, but I am concerned that it isn't entirely feasible. I recently spoke to a team that deploys RPKI-invalid route filtering and they mentioned that they may not enable it for certain types of peers. They will also disable it if there is an issue with the filtering system itself or they may grant a temporary exception for a customer. I would assume this is common?

As I currently read the text, the CA would be required to file an incident and halt issuance if any of these exceptions were made. CAs are very unlikely to have an SLA with their network operator to always support RPKI-invalid route filtering. 

How should this situation be handled ideally?

docs/BR.md
XolphinMartijn: @ChristopherRC  The proposed language above looks great! Thank you

romanf:
I have a question to possible architectural solutions: Would it be acceptable to establish VPN tunnels with exit points in different parts of the world and still keep all the processing in the central location(s)? Would such VPN tunnels be considered "delegated third parties" since the VPN provider could potentially be malicious and exit all the traffic at the same location?

tobij:
> I have a question to possible architectural solutions: Would it be acceptable to establish VPN tunnels with exit points in different parts of the world and still keep all the processing in the central location(s)? Would such VPN tunnels be considered "delegated third parties" since the VPN provider could potentially be malicious and exit all the traffic at the same location?

This is an interesting and indeed possibly relevant question. My first intuitive thought on this would be that it should possibly not be considered allowed under the current language, or at least I would err on the side of it not being allowed as an implementing CA. However, I think there is no fundamental reason not to allow it if certain constraints are met. For example, I think such a scenario should - if at all - only be allowed if it is the CA that is operating the VPN endpoints, at the same level it would operate more straight forward MPIC implementations, in particular such that the VPN solution is subject to only the same threat model as the straightforward MPIC implementation would be. This could indeed in principle be done by setting up a VPN endpoint in the cloud, but would not (known to) be met in the same way if a CA contracted a VPN endpoint from a third party. This is not to preclude that there might be VPN endpoint types provided by certain CSPs that would approach a straight-forward MPIC implementation in threat and risk model, however that also would be a special case.

timfromdigicert:
It’s important to remember what I pointed out in Ottawa about the security level that is being proposed for these additional vantage points: that they have no security or audit requirements whatsoever.  If you want to install a default Windows 98 install and set the password to “hackme”, that’s acceptable for these additional validation points.

 

My first impression is that the implementation you are proposing is actually somewhat clever, as using geographically diverse set of VPN-protected exit points actually potentially achieves the same goal we have been trying to achieve while exposing far less than you would by throwing up poorly-protected versions of your validation implementation all over the planet.

 

And before someone claims no one would ever do that, my experience is that in the absence of any security requirements, someone will attempt to implement the minimum (and, often, fail to achieve even that).

 

-Tim

docs/BR.md
```diff
@@ -994,25 +1026,88 @@ Databases maintained by the CA, its owner, or its affiliated companies do not qu
...
+
+A Network Perspective MAY use a recursive DNS resolver that is NOT co-located with the Network Perspective. However, the DNS resolver used by the Network Perspective MUST fall within the same Regional Internet Registry service region as the Network Perspective relying upon it. Furthermore, for any pair of DNS resolvers used on a Multi-Perspective Issuance Corroboration attempt, the straight-line distance between the two States, Provinces, or Countries the DNS resolvers reside in MUST be at least 500 km. The location of a DNS resolver is determined by the point where unencapsulated outbound DNS queries are first handed off to the network infrastructure providing Internet connectivity to that DNS resolver.
```
romanf, line 1064: Do I understand this paragraph correctly in that if a remote perspective uses a DNS resolver NOT run by the CA itself, this would still be considered a Delegated Third Party and thus be forbidden?

romanf:
@tobij Do I understand your point correctly: Your interpretation is that it would be preferrable if a CA rents a bunch of VMs from a Cloud provider and runs a VPN endpoint software on them, than renting the VPN connection?

tobij:
@romanf You are correct - this is my point indeed, but I am open to being wrong about my point.

docs/BR.md
ryancdickson: @jdkasten - thanks for you comment!

Is it safe to assume the specific comment is focused on the following bullet?:

- [Remote Network Perspectives performing Multi-Perspective Issuance Corroboration MUST]:  _Forward all Internet traffic via a network or set of networks that filter all RPKI-invalid BGP routes as defined by RFC 6811_"

**Proposed Update:** 

If so, what if instead the language was modified to something like the following:

- [Remote Network Perspectives performing Multi-Perspective Issuance Corroboration MUST]: _**under normal operating conditions**, forward all Internet traffic via a network or set of networks that filter **all or some** RPKI-invalid BGP routes as defined by RFC 6811._

Feedback on the above is welcome. 

**Change Justification:** 

As I understand it, there are two distinct categories of exceptions to RPKI filtering: 
- the first is temporary exceptions where certain peers are identified as potentially having an issue with RPKI ROV enforcement and these sessions are given a temporary exception. 
- the second is that some types of sessions are considered exempt from a network's filtering policy. There are documented cases of networks that filter routes from peers but not from customers (see [https://isbgpsafeyet.com/](https://isbgpsafeyet.com/).

The approach proposed above would consider both types of exceptions described above, while also addressing your concern re: risk of blocking issuance when filtering is beyond the CA Owner's control. 

FWIW, it would seem there are some tools available that would help CA owners evaluate whether ISPs are performing filtering, for example https://rpkitest.nlnetlabs.net/ from https://nlnetlabs.nl/research/projects/

docs/BR.md
ryancdickson: @romanf - One important clarification - the proposal does not prohibit Delegated Third Parties from participating in MPIC. MPIC Network Perspectives should be considered as having the potential of blocking issuance, not authorizing it (this conclusion must be derived by the Primary Network Perspective) as described in 3.2.2.4 and 3.2.2.5.

ryancdickson: @romanf / @tobij / @timfromdigicert 

[Sorry for the odd formatting/threading of these messages, I'm responding to these comments directly in GitHub rather than via email.]

Concerning the use of VPNs:
- My interpretation of the proposal is that use of VPNs would not be considered prohibited. 
- My position is that both 1st (CA-operated) and 3rd party (non-CA operated) VPNs should be permitted. At a network level, a VPN achieves the same effect of causing traffic egress at a different location as actually running validator code at remote vantage perspectives. While care must be taken to ensure that the DNS cache is not shared between different MPIC calls as described required in the proposal, fundamentally, the VPN implementation can achieve the exact same security properties against BGP attacks as remote validators (i.e., those hosted on a cloud VM).
- It's not clear to me how relying on a VPN provider for MPIC is meaningfully different than when considering how a CA relies on their existing Internet Service Provider for outbound internet connectivity. 

Comments and considerations welcome!

docs/BR.md
romanf: I'm not a network specialist.. how does one meet this MUST requirement?

ryancdickson: Great, thanks for confirming! 

We'll:
1) stage these changes in https://github.com/ChristopherRC/servercert/blob/SC-067-Version-2/docs/BR.md in preparation for the next round of discussion, and
2) leave this comment thread open in the event there's any other discussion.

ryancdickson: @romanf - the expectation is that the lookups performed by the remote network perspectives used for MPIC would rely on an ISP that performs RPKI-invalid route filtering.

Some examples are located at: https://isbgpsafeyet.com/ (note: I cannot speak to the completeness of this list, however, I would expect any ISP to be able to describe whether filtering is in place).

romanf: So the CA would need written confirmation from all of their ISPs that they're using for the various perspectives that they perform RPKI-invalid route filtering. Correct?

docs/BR.md
timfromdigicert: @ryancdickson I haven't thought about it too much in detail but I think your bullet points are at least in the neighborhood of correct, and would be a great start.

timfromdigicert:
If the cloud provider can't be trusted, then renting a VM instead of a VPN connection doesn't stop their ability to mess with you.

At the end of the day, what matters is that the DNS for the additional perspective is performed from a geographically diverse location, and that the implementation passes some extremely low security bar (I'm adding that last point to avoid just dropping the VPN entirely).  If the wrong VPN or provider or CDN is relied upon, then a CA might find that that result is not achieved, then they're in compliance failure / incident land.  They should choose an architecture and solution providers to minimize the chances of that happening, because that's their job.  I don't think we can do that part for them.

docs/BR.md
ryancdickson: @romanf - I think there are a few ways CA Owners could collect evidence to demonstrate a service provider performs invalid-route filtering. Written confirmation from the service provider(s) seems like one way this could be addressed. Tools like https://rpkitest.nlnetlabs.net/ could also be used by CA Owners and their auditors to corroborate statements from service providers or lists like we see on https://isbgpsafeyet.com/.

wthayer: The changes proposed by @jdkasten are significant improvements! Having said that, Google cloud apparently does not yet meet the RPKI filtering requirement. There is currently a lot of active RPKI deployment activity, and I wonder if we should phase in this requirement by starting with a SHOULD and perhaps some language that forces CAs to think about how they can meet the requirement?

ChristopherRC: @wthayer 

While we prefer the MUST due to the demonstrated security benefits of invalid route filtering, we understand your concern and we have proposed a phased approach at https://github.com/ChristopherRC/servercert/commit/d6125463d370f57a78d29c072c49c88dda7d579d. Feedback is welcome! The effective date for invalid route filtering in this proposal aligns with when we would expect to see blocking behavior for MPIC (September 15, 2025). Of course, CAs are encouraged to pursue earlier adoption.

wthayer: Thank you @ChristopherRC. I am particularly interested in having the ability to run perspectives on Google cloud, so I am trying to get some timing from them to determine if this change helps. In general it might be better to include something like this right above your proposed change:

Under normal operating conditions, remote Network Perspectives performing Multi-Perspective Issuance Corroboration SHOULD forward all Internet traffic via a network or set of networks that filter all or some RPKI-invalid BGP routes as defined by RFC 6811. The CA MUST document the reason(s) for all non-compliance.

Maybe this is going too far, but my intent is to force CAs to think about meeting this before the requirement actually comes into effect. Also, I'd want to improve that paragraph to remove ambiguity around the meaning of "non-compliance" before incorporating it, should you like to idea.

job: Hi Folks - I work in Fastly's network engineering department and am available as RPKI and BGP Routing Security subject matter expert.

Reviewing the draft BR requirements from a network engineering perspective, to me it seems RFC 6811 might not be the best reference. RFC 6811 merely is an algorithm description. The outcome of the algorithm depends on the algorithm's inputs, which are the RPKI Trust Anchor Locators the network operator has chosen to use. For example, if zero RPKI Trust Anchors are used, the outcome of the RFC 6811 validation algorithm will always be `not-found`. Would a network applying RFC 6811 _without any Trust Anchors_ be compliant with the spirit of the requirement? I don't think so! :-)

However, explicitly prescribing which RPKI Trust Anchors a CA's network must/should use also has its challenges: Some Regional Internet Registries (RIRs) employ a Relying Party Agreement (RPA) to restrict access to their online Resource Certification PKI services. The CA/B BR mandating that a particular RPKI Trust Anchor MUST be used, also means the CA's network operator MUST agree the current and any future versions of such RPKI Relying Party Agreements. Some wiggle room probably is needed to allow network operators to make independent choices.

There is a non-zero number of networks which use only 4 of the 5 RIR Trust Anchors due to legal concerns with the 5th RIR's RPA. The world's second most popular RPKI validator [rpki-client](https://man.openbsd.org/rpki-client#FILES) does not include all 5 RIR Trust Anchor TALs in its distribution. But, I'd consider network operations using 4 out of 5 RIR Trust Anchor Locators to be compliant with the spirit of this requirement.

I'd like to echo @jdkasten's comments about RPKI-filtering generally speaking not being a binary state of 'on' or 'off'. In this [NANOG](https://mailman.nanog.org/pipermail/nanog/2021-April/213346.html) posting I've outlined reasons why a network at any given time might not have RPKI enabled on all its routers. If RPKI-filtering is disabled on just 1 router in a global IP network, it might appear to outsiders the network is not doing any RPKI-filtering at all.

It is also good to keep in mind that RPKI-based filtering is not the **only** approach to improve routing security and attempt to mitigate routing incidents. For example, RFC 9234 specifies an approach using in-band BGP signalling to help avoid route leaks without RPKI. 

I'd like to propose text along these lines:

> Networks used by Remote Network Perspectives performing Multi-Perspective Issuance Corroboration MUST implement measures to mitigate BGP routing incidents in the global Internet routing system. Network operators are RECOMMENDED to use mechanisms based on Secure Inter-Domain Routing (RFC 6480), for example, BGP Prefix Origin Validation (RFC 6811), make use of other non-RPKI route-leak prevention mechanisms (such as RFC 9234), and apply best current practices described in BCP 194.

I understand the above is a bit of a departure from the original text, but I think the original text's hard requirements might not be entirely realistic and might distract from the ultimate goal: encourage networks to undertake efforts to improve routing security. I think it is important to encourage RNP's to make use of networks that make an effort to avoid routing incidents.

docs/BR.md
aaomidi: This can be read as "well, we did MP for CAA, so we don't need to do it for DV". Is that the intention here?

ryancdickson: @aaomidi - the definition is intended to serve the most general use case and not inadvertently limit CA Owner flexibility re: implementation. 

For example, if we assume I need 5 "successful" DCV/CAA lookups, I should be able to:
- perform the lookups from 5 Network Perspectives (each perspective evaluates both DCV and CAA), or
- perform the lookups from 10 Network Perspectives (5 perspectives evaluate DCV, 5 others CAA).

Regardless of the above, the requirements in Section 3.2.2.9 make expectations clear (especially those [staged](https://github.com/ChristopherRC/servercert/blob/SC-067-Version-2/docs/BR.md#3229-multi-perspective-issuance-corroboration) for the next round of discussion) that both DCV and CAA need to be considered.

docs/BR.md
ryancdickson: Hi @job,

1) We are familiar with and very much appreciate your contributions in the routing security space. Thank you for reviewing the proposal, offering concrete suggestions for improvement, and volunteering your expertise to help amplify the positive impact this ballot can have on the Web PKI and user security!

2) On the proposed update:

> Networks used by Remote Network Perspectives performing Multi-Perspective Issuance Corroboration MUST implement measures to mitigate BGP routing incidents in the global Internet routing system. Network operators are RECOMMENDED to use mechanisms based on Secure Inter-Domain Routing (RFC 6480), for example, BGP Prefix Origin Validation (RFC 6811), make use of other non-RPKI route-leak prevention mechanisms (such as RFC 9234), and apply best current practices described in BCP 194.

One challenge with this approach is that it imposes requirements on service providers that are not obligated to adhere to the CA/Browser Forum policies. Do you think the following re-framing could achieve the same security outcome as your proposal?

> Network Perspectives performing Multi-Perspective Issuance Corroboration MUST rely upon networks that implement measures to mitigate BGP routing incidents in the global Internet routing system. In support of performing Multi-Perspective Issuance Corroboration, CAs SHOULD rely upon network operators that: 

> 1. use mechanisms based on Secure Inter-Domain Routing (RFC 6480), for example, BGP Prefix Origin Validation (RFC 6811), 
> 2. make use of other non-RPKI route-leak prevention mechanisms (such as RFC 9234), and 
> 3. apply best current practices described in BCP 194.

> While It is RECOMMENDED that under normal operating conditions Network Perspectives performing Multi-Perspective Issuance Corroboration forward all Internet traffic via a network or set of networks that filter all or some RPKI-invalid BGP routes as defined by RFC 6811, it is NOT REQUIRED.

This approach attempts to blend the elements of your proposal, while also aligning with the strong recommendation of relying on RPKI-invalid route filtering shared with us by the Princeton research team at F2F 57 and the corresponding [research](https://arxiv.org/pdf/2302.08000.pdf) that inspired this work.

Comments on this approach are welcome and appreciated!

job: Thank you @ryancdickson for the kind words.

Your re-framed version seems an improvement!

I would suggest to remove the words `all or some`.

job: To elaborate a bit:

The Princeton research team's observation that a proper issuance process execution has a dependency on proper network reachability seems on the nose to me. MPIC to me seems a fantastic mechanism to attempt to overcome various potential issues which may exist in the Internet's routing system substrate, I'm excited to see MPIC advance. Recognizing that proper network reachability is a key ingredient to the issuance process - regardless of MIPC - encouraging all forms of BGP routing hygiene seems very helpful.

romanf: Sorry if this question is a bit stupid... I'm not a network engineer. :)

How would a CA implement the MUST requirement
"Network Perspectives performing Multi-Perspective Issuance Corroboration MUST rely upon networks that implement measures to mitigate BGP routing incidents in the global Internet routing system." ?

Would this mean - in laymans-terms - that the networks where the remote perspecitves are located (or - if VPN are allowed - the VPN endpoint is) must implement this BGP mitigations? Or does this apply to all networks the CA is using (incl. their main internet access network as well as transit networks to the remote perspectives)?

ryancdickson: @romanf - not a "stupid" question. If there's opportunity to clarify these requirements either through discussion or updates to the ballot, they are worth addressing.

I would consider this a more flexible option than what we originally proposed (i.e., "_[Remote Network Perspectives performing Multi-Perspective Issuance Corroboration MUST]: under normal operating conditions, forward all Internet traffic via a network or set of networks that filter all or some RPKI-invalid BGP routes as defined by RFC 6811."_). 

This should be considered more flexible because rather than only focusing on RPKI-invalid route filtering to improve routing security relied upon by the Network Perspectives, it allows consideration for other routing security technologies as described by @job in [this](https://github.com/cabforum/servercert/pull/487#discussion_r1566070938) comment. That is to say, RPKI-invalid route filtering should not be the only acceptable approach adopted by network operators/ISPs to improve routing security that's described in the ballot. 

It's my understanding that if a CA Owner ensured its Network Perspectives relied upon service providers that filter RPKI-invalid routes (like many of those described on isbgpsafeyet.com) - they'd satisfy this expectation. 

Feedback from @job and other routing security experts welcome (as I am definitely not one!).

romanf: The list at isbgpsafeyet.com is awfully short... and just to double-check: this requirement is only for the remote perspectives, correct? Otherwise I can imagine some CAs getting into deep troubles if the DCs that host their primary perspective can't be connected to any ISP providing the required GBP filtering...

docs/BR.md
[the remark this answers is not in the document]
ryancdickson: Hi Doug!

I often find example scenarios helpful in cases such as this. Please find one below:

Note: for this example, assume DCV data reuse is set to 90 days by the CA.

**Today:**
- An applicant requests a certificate for the dNSname "ilovetls.com" using “Agreed-Upon Change to Website - ACME” (TLS BR 3.2.2.4.19).
- Primary DCV initiated: Their ACME client demonstrates control over the requested domain in accordance with TLS 3.2.2.4.19, and this is verified by the CA from its primary data center.
- MPIC DCV initiated: The demonstration of control over the requested domain would be corroborated by [N] remote network perspectives in accordance with the proposed TLS BR 3.2.2.9.
- MPIC CAA initiated: Successful MPIC CAA checks must take place for the requested domain in accordance with the proposed TLS BR 3.2.2.9.
- Primary CAA check initiated: The CA checks CAA for the the requested domains in accordance with TLS BR 3.2.2.8.
- Assuming no issues with the above, the certificate is issued.

**100 days from now:**
- The same applicant requests another certificate for the dNSname "ilovetls.com" using “Agreed-Upon Change to Website - ACME” (TLS BR 3.2.2.4.19).
- Primary DCV initiated: Their ACME client demonstrates control over the requested domain in accordance with TLS 3.2.2.4.19, and this is verified by the CA from its primary data center.
- MPIC DCV initiated: The demonstration of control over the requested domain would be corroborated by [N] remote network perspectives in accordance with the proposed TLS BR 3.2.2.9.
- MPIC CAA optional: Given it's been less than 398-days since the last successful MPIC CAA check and the same applicant is requesting the certificate, this may be skipped as described in the proposed 3.2.2.9.
- Primary CAA check initiated: The CA checks CAA for the the requested domains in accordance with TLS BR 3.2.2.8.
- Assuming no issues with the above, the certificate is issued.

Note: there's flexibility to the sequencing above (e.g., the MPIC lookups for DCV and CAA may take place at the same time), this is just one way of doing it.  

**To directly address your question:** Yes [but only for the remainder of the permitted 398-day reuse period for MPIC CAA data - and if the Applicant is determined to be the same as the original request.]

Does this address your question?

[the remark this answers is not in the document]
romanf: Thanks a lot Ryan for this very illustrative example! I understand where it comes from... it just seems a bit counter-intuitive. ;-) I probably missed the discussion that led to this wording. Any pointers where I can read up?
Thx!

ChristopherRC:
Closing in favor of https://github.com/cabforum/servercert/pull/507

PR #490, opened 2024-03-13
SC-72 - Delete except to policyQualifiers in EVGs; align with BRs by making them NOT RECOMMENDED

vanbroup:
This ballot updates the TLS Extended Validation Guidelines (EVGs) by removing the exceptions to `policyQualifiers` in section 9.7, to align them with the Baseline Requirements (BRs). 

As result, this ballot changes `policyQualifiers` from `MUST` to `NOT RECOMMENDED` as stated in the TLS Baseline Requirements, resolving a discrepancy introduced by [Ballot SC-62v2](https://cabforum.org/2023/03/17/ballot-sc62v2-certificate-profiles-update/) between section [7.1.2.7.9 Subscriber Certificate Policies](https://cabforum.org/working-groups/server/baseline-requirements/requirements/#71279-subscriber-certificate-certificate-policies) of the BRs and the [Additional Technical Requirements for EV Certificates](https://cabforum.org/working-groups/server/extended-validation/guidelines/#97-additional-technical-requirements-for-ev-certificates) in the EVGs.

The following motion has been proposed by Paul van Brouwershaven (Entrust) and endorsed by Dimitris Zacharopoulos (HARICA) and Iñigo Barreira (Sectigo).

### DIFF
--- a/docs/EVG.md
+++ b/docs/EVG.md
@@ -608,33 +608,13 @@ All provisions of the Baseline Requirements concerning Minimum Cryptographic Alg
 
    Otherwise, it MAY contain the anyPolicy identifier.
 
-2. The following fields MUST be present if the Subordinate CA is not controlled by the entity that controls the Root CA.
-
-   * `certificatePolicies:policyQualifiers:policyQualifierId`
-
-      `id-qt 1` [RFC 5280]
-
-   * `certificatePolicies:policyQualifiers:qualifier:cPSuri`
-
-      HTTP URL for the Root CA's Certification Practice Statement
-
-3. The `certificatePolicies` extension in EV Certificates issued to Subscribers MUST include the following:
+2. The `certificatePolicies` extension in EV Certificates issued to Subscribers MUST include the following:
 
    * `certificatePolicies:policyIdentifier` (Required)
 
       The Issuer's EV policy identifier
 
-4. Including `policyQualifiers` in the `certificatePolicies` extension in EV Certificates issued to Subscribers is NOT RECOMMENDED, but when included MUST include the following:
-
-   * `certificatePolicies:policyQualifiers:policyQualifierId` (Required)
-
-      `id-qt 1` [RFC 5280]
-
-   * `certificatePolicies:policyQualifiers:qualifier:cPSuri` (Required)
-
-      The HTTP or HTTPS URL for the Issuing CA’s Certificate Policies, Certification Practice Statement, Relying Party Agreement, or other pointer to online policy information provided by the Issuing CA.
-
-5. The `cRLDistributionPoints` extension MUST be present in Subscriber Certificates if the certificate does not specify OCSP responder locations in an `authorityInformationAccess` extension.
+3. The `cRLDistributionPoints` extension MUST be present in Subscriber Certificates if the certificate does not specify OCSP responder locations in an `authorityInformationAccess` extension.
 
 ## 9.8. Certificate Extensions
 
--- a/docs/EVG.md
+++ b/docs/EVG.md
@@ -624,15 +624,17 @@ All provisions of the Baseline Requirements concerning Minimum Cryptographic Alg
 
       The Issuer's EV policy identifier
 
+4. Including `policyQualifiers` in the `certificatePolicies` extension in EV Certificates issued to Subscribers is NOT RECOMMENDED, but when included MUST include the following:
+
    * `certificatePolicies:policyQualifiers:policyQualifierId` (Required)
 
       `id-qt 1` [RFC 5280]
 
    * `certificatePolicies:policyQualifiers:qualifier:cPSuri` (Required)
 
-      HTTP URL for the Subordinate CA's Certification Practice Statement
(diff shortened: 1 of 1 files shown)


robstradling:
@vanbroup: This section (9.7) of the EVGs begins with:
*"All provisions of the Baseline Requirements concerning Minimum Cryptographic Algorithms, Key
Sizes, and Certificate Extensions apply to EV Certificates with the following exceptions:"*

Am I right that what you're trying to do here is align the EVG requirements for Subscriber Certificate certificatePolicies extensions with the equivalent BR requirements?  If so, then wouldn't it make more sense for this draft ballot to remove the 3rd exception in section 9.7 entirely?

clintwilson:
As I recall, this discrepancy was discussed multiple times during the years of work on SC62. This is not (and never has been) an unintentional discrepancy, but rather an intentional difference between the TBRs and the EVGs - representing a consensus view and compromise based on numerous discussions in the CA/B Forum. 

Broadly in those discussions some CAs (Entrust included, iirc) indicated a strong preference to preserve allowing `policyQualifiers` values in certificates while others (Apple included) were supportive of `policyQualifiers` being documented as a MUST NOT as part of the Profiles work (which became SC62v2). While the Profiles work had set `policyQualifiers` as a MUST NOT early on, the later discussions (over more than a year) led to the text being updated to NOT RECOMMENDED later (not entirely because of, but still in no small part due to the requirement in the EVGs to include `policyQualifiers`).

Given this rather large and sudden reversal of position, it seems like the proposal here may be better represented by disallowing `policyQualifiers` in both the TBRs and EVGs, as was originally proposed (and which remains part of what we've expected to address with "Profiles V2", alongside a myriad of other things).

CBonnell:
I agree with @robstradling that the most straightforward way to align the EVGs and BRs is to merely remove the exception in the EVGs.

Also, could this ballot be updated to remove the "MUST add a cPSuri policyQualifier to externally operated EV Subordinate CA certificates" requirement in section 9.7 (2)? I think this also has the potential for being missed by readers of the BR and EVG.

vanbroup:
@clintwilson 
> As I recall, this discrepancy was discussed multiple times during the years of work on SC62. This is not (and never has been) an unintentional discrepancy, but rather an intentional difference between the TBRs and the EVGs - representing a consensus view and compromise based on numerous discussions in the CA/B Forum.

I don't recall that this was brought up as an intentional discrepancy and have not be able to find this in the minutes. The limited search in the mail archives is not very reliable so please point me to the right minutes if you can.

As you stated Entrust was strongly in favor to preserve the cPSuri in the policyQualifiers but we found consensus with the agreement that this information will remain in the issuing hierarchy.

dzacharo:
> As you stated Entrust was strongly in favor to preserve the cPSuri in the policyQualifiers but we found consensus with the agreement that this information will remain in the issuing hierarchy.

Paul,

In the TLS BRs, policyQualifiers are also NOT RECOMMENDED in the TLS CA Certificate Profile. See [Section 7.1.2.10.5](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712105-ca-certificate-certificate-policies)

vanbroup:
> In the TLS BRs, policyQualifiers are also NOT RECOMMENDED in the TLS CA Certificate Profile. See [Section 7.1.2.10.5](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712105-ca-certificate-certificate-policies)

Thanks for the correction, it would be great if we had a recording or good minutes of this discussion, maybe we where heading to that direction at some point but eventually decided to keep it NOT RECOMMENDED everywhere. 

For this ballot we only care about what the requirements currently state so lets focus on that for the text and justification of the ballot.

vanbroup:
@robstradling @CBonnell I removed the exceptions as it indeed clearly states "_All provisions of the Baseline Requirements concerning Minimum Cryptographic Algorithms, Key Sizes, and Certificate Extensions apply to EV Certificates with the following exceptions_".

@dzacharo are you ok with this as an endorser of the ballot?

vanbroup:
> Given this rather large and sudden reversal of position, it seems like the proposal here may be better represented by disallowing `policyQualifiers` in both the TBRs and EVGs, as was originally proposed (and which remains part of what we've expected to address with "Profiles V2", alongside a myriad of other things).

As we are removing the text from the EVGs, I think it would be better to have a separate ballot to propose changing the inclusion for policyQualifiers from NOT RECOMMENDED to MUST NOT in the BRs.

dzacharo:
> @robstradling @CBonnell I removed the exceptions as it indeed clearly states "_All provisions of the Baseline Requirements concerning Minimum Cryptographic Algorithms, Key Sizes, and Certificate Extensions apply to EV Certificates with the following exceptions_".
> 
> @dzacharo are you ok with this as an endorser of the ballot?

I agree to align with the current TLS BRs which does not cause any conflicts with other existing standards that might require the existence of the cPSUri.

HarshalSharma21:
[Get Assignment Done ](https://in.pinterest.com/getassignmenthelponline/)

PR #528, opened 2024-06-27
Fix lists format

hablutzel1:
The reason for this PR is the same that https://github.com/cabforum/forum/pull/51, which is basically that Markdown lists should be surrounded by blank lines for correct presentation.

Additionally, you can see a concrete presentation problem example for the TLS BRs in https://github.com/cabforum/build-guidelines-action/pull/51#issue-2378471003.

Finally, merging https://github.com/cabforum/build-guidelines-action/pull/51 could prevent the introduction of this type of Markdown problems in the future.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -2,6 +2,7 @@
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 subtitle: Version 2.0.5
 author:
+
   - CA/Browser Forum
 
 date: 7-June-2024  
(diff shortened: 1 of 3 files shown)


docs/BR.md
```diff
@@ -2,6 +2,7 @@
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 subtitle: Version 2.0.5
 author:
```
vanbroup, line 4: The front matter is in YAML, not Markdown.

It should be linted as YAML where this nestling would not be expected.

hablutzel1: I've removed the changes to the YAML frontmatter.

PR #534, opened 2024-07-12
Update BR.md

ENEN-DTR:
These changes address ambiguous descriptions in the sections 

- 1.6.1, 
- 7.1.2.7.7, 
- 7.1.2.11.2, 
- 7.1.2.3.2, 
- 7.1.2.7.9, 
- 7.1.2.8.8, 
- 7.1.2.10.5. 

All these are linguistic improvements to avoid ambiguities. It introduces the definition of scheme according to RFC 3986, section 3.1 to the BRs. It also clarifies that no methods other than those listed in the respective descriptions are permitted. 

D-Trust had an incident due to the current language of the BRs and its interpretation. The changes are intended to help avoid such misinterpretations in the future. Please see: https://bugzilla.mozilla.org/show_bug.cgi?id=1884714

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -466,6 +466,8 @@ The script outputs:
 
 **Root Certificate**: The self-signed Certificate issued by the Root CA to identify itself and to facilitate verification of Certificates issued to its Subordinate CAs.
 
+**Scheme**: A URI Scheme that is defined by RFC 3986, section 3.1. 
+
 **Short-lived Subscriber Certificate**: For Certificates issued on or after 15 March 2024 and prior to 15 March 2026, a Subscriber Certificate with a Validity Period less than or equal to 10 days (864,000 seconds). For Certificates issued on or after 15 March 2026, a Subscriber Certificate with a Validity Period less than or equal to 7 days (604,800 seconds).
 
 **Sovereign State**: A state or country that administers its own government, and is not dependent upon, or subject to, another power.
@@ -2089,10 +2091,10 @@ Table: Policy Restricted
 
 Table: Permitted `policyQualifiers`
 
-| __Qualifier ID__                     | __Presence__ | __Field Type__ |  __Contents__ |
+| __policyQualifier__                     | __Presence__ | __Field Type__ |  __Contents__ |
 | ---                                  | -            | -              | -----         |
-| `id-qt-cps` (OID: 1.3.6.1.5.5.7.2.1) | MAY          | `IA5String`    | The HTTP or HTTPS URL for the Issuing CA's Certificate Policies, Certification Practice Statement, Relying Party Agreement, or other pointer to online policy information provided by the Issuing CA. |
-| Any other qualifier                  | MUST NOT     | -              | -             |
+| `id-qt-cps` (OID: 1.3.6.1.5.5.7.2.1) | MAY          | `IA5String`    | When provided, every policyQualifier SHALL have the uniformResourceIdentifier SCHEME HTTP or HTTPS for the Issuing CA's Certificate Policies, Certification Practice Statement, Relying Party Agreement, or other pointer to online policy information provided by the Issuing CA. |
+| Any other policyQualifier                  | MUST NOT     | -              | Other policyQualifiers SHALL NOT be present. |
 
 
 ##### 7.1.2.3.3 Technically Constrained Non-TLS Subordinate CA Extended Key Usage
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -2425,11 +2427,11 @@ The `AuthorityInfoAccessSyntax` MUST contain one or more `AccessDescription`s. E
...
 
-| __Access Method__ | __OID__            | __Access Location__         | __Presence__ | __Maximum__ | __Description__ |
+| __Access Method__ | __OID__            | __Access Location__         | __Presence__ | __Maximum__ | __Access Description__ |
 | --                | --                 | ----                        | -            | -          | ---             |
-| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | MAY          | \*         | A HTTP URL of the Issuing CA's OCSP responder. |
-| `id-ad-caIssuers` | 1.3.6.1.5.5.7.48.2 | `uniformResourceIdentifier` | SHOULD       | \*         | A HTTP URL of the Issuing CA's certificate. |
-| Any other value   | -                  | -                           | MUST NOT     | -          | No other `accessMethod`s may be used. |
+| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | MAY          | \*         | When provided, every accessMethod SHALL have the uniformResourceIdentifier SCHEME HTTP. <br>Other schemes SHALL NOT be present. |
```
romanf, line 2432: Do you need the <br> here? I know it's valid markdown but some renderers (e.g. for PDF output) still don't like it. I would suggest to remove it.

ENEN-DTR: Hi Roman, "br" are removed now. Thanks for the heads up.

docs/BR.md
```diff
@@ -2089,10 +2091,10 @@ Table: Policy Restricted
 
 Table: Permitted `policyQualifiers`
 
-| __Qualifier ID__                     | __Presence__ | __Field Type__ |  __Contents__ |
+| __policyQualifier__                     | __Presence__ | __Field Type__ |  __Contents__ |
```
clintwilson, line 2094: This should likely be `policyQualifierId` as we're talking about `policyQualifierId` in `PolicyQualifierInfo` within `policyQualifiers`.

docs/BR.md
```diff
@@ -2089,10 +2091,10 @@ Table: Policy Restricted
...
 | ---                                  | -            | -              | -----         |
-| `id-qt-cps` (OID: 1.3.6.1.5.5.7.2.1) | MAY          | `IA5String`    | The HTTP or HTTPS URL for the Issuing CA's Certificate Policies, Certification Practice Statement, Relying Party Agreement, or other pointer to online policy information provided by the Issuing CA. |
-| Any other qualifier                  | MUST NOT     | -              | -             |
+| `id-qt-cps` (OID: 1.3.6.1.5.5.7.2.1) | MAY          | `IA5String`    | When provided, every policyQualifier SHALL have the uniformResourceIdentifier SCHEME HTTP or HTTPS for the Issuing CA's Certificate Policies, Certification Practice Statement, Relying Party Agreement, or other pointer to online policy information provided by the Issuing CA. |
```
clintwilson, line 2096: I don't think `SCHEME` needs to be capitalized here; it should be `Scheme` instead, as a defined term.
The phrasing here can also be simplified, imo, to `The uniformResourceIdentifier using the HTTP or HTTPS Scheme for the....`
If keeping the current wording, I think it should be updated to `When provided, the id-qt-cps policyQualifierId SHALL have the uniformResourceIdentifier Scheme HTTP or HTTPS for the....`

docs/BR.md
```diff
@@ -2425,11 +2427,11 @@ The `AuthorityInfoAccessSyntax` MUST contain one or more `AccessDescription`s. E
 
 The `AuthorityInfoAccessSyntax` MAY contain multiple `AccessDescription`s with the same `accessMethod`, if permitted for that `accessMethod`. When multiple `AccessDescription`s are present with the same `accessMethod`, each `accessLocation` MUST be unique, and each `AccessDescription` MUST be ordered in priority for that `accessMethod`, with the most-preferred `accessLocation` being the first `AccessDescription`. No ordering requirements are given for `AccessDescription`s that contain different `accessMethod`s, provided that previous requirement is satisfied.
 
-| __Access Method__ | __OID__            | __Access Location__         | __Presence__ | __Maximum__ | __Description__ |
+| __Access Method__ | __OID__            | __Access Location__         | __Presence__ | __Maximum__ | __Access Description__ |
```
clintwilson, line 2430: I'm not clear on why `Access` was added here. I think leaving it as just `Description` is clear and aligns with the many other tables that have that column.

docs/BR.md
```diff
@@ -2425,11 +2427,11 @@ The `AuthorityInfoAccessSyntax` MUST contain one or more `AccessDescription`s. E
...
 | --                | --                 | ----                        | -            | -          | ---             |
-| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | MAY          | \*         | A HTTP URL of the Issuing CA's OCSP responder. |
-| `id-ad-caIssuers` | 1.3.6.1.5.5.7.48.2 | `uniformResourceIdentifier` | SHOULD       | \*         | A HTTP URL of the Issuing CA's certificate. |
-| Any other value   | -                  | -                           | MUST NOT     | -          | No other `accessMethod`s may be used. |
+| `id-ad-ocsp`      | 1.3.6.1.5.5.7.48.1 | `uniformResourceIdentifier` | MAY          | \*         | When provided, every accessMethod SHALL have the uniformResourceIdentifier SCHEME HTTP. Other schemes SHALL NOT be present. |
```
clintwilson, line 2432: Similar to above regarding rewording, simplifying, and SCHEME, but also the second sentence needs to have `schemes` bet capitalized: `Other Schemes SHALL NOT be present.`

docs/BR.md
```diff
@@ -2953,7 +2956,7 @@ Table: `DistributionPoint` profile
 | `reasons`           | MUST NOT        |                 |
 | `cRLIssuer`         | MUST NOT        |                 |
 
-A `fullName` MUST contain at least one `GeneralName`; it MAY contain more than one. All `GeneralName`s MUST be of type `uniformResourceIdentifier`, and the scheme of each MUST be "http". The first `GeneralName` must contain the HTTP URL of the Issuing CA's CRL service for this certificate.
+A `fullName` MUST contain at least one `GeneralName`; it MAY contain more than one. All `GeneralName`s MUST be of type `uniformResourceIdentifier`, and have the uniformResourceIdentifier SCHEME HTTP. Every `GeneralName` must contain the HTTP URL of the Issuing CA's CRL service for this certificate.
```
clintwilson, line 2959: I think the original wording is clearer, but can still be updated now that we have a defined term to reference:
> `GeneralName`s MUST be of type `uniformResourceIdentifier`, and the Scheme of each MUST be HTTP. Each `GeneralName` MUST contain the....

docs/BR.md
ENEN-DTR: I agree with this change. Thanks.

docs/BR.md
ENEN-DTR: Sounds good.

docs/BR.md
ENEN-DTR: Agreed.

docs/BR.md
ENEN-DTR: Sounds good.

docs/BR.md
ENEN-DTR: Sounds good.

clintwilson:
> @clintwilson; @CBonnell: I'm not happy with the term "Scheme" as it is also used also in different contexts. Would it be better to use "URI Scheme" instead so that it's not confused with the general term "scheme"?

Agreed, I think "URI Scheme" as a defined term is a good idea for this update as well.

Issue #422, opened 2023-02-15
Section 4.9.10: Untangle "assigned" vs "reserved" serials, precertificates, and OCSP


aarongable:
The Profiles ballot updates [Section 7.1.2.9](https://github.com/cabforum/servercert/blob/profiles/docs/BR.md#7129-precertificate-profile) to say:

> Once a Precertificate is signed, relying parties are permitted to treat this as a binding committment from the CA of the intent to issue a corresponding Certificate, or more commonly, that a corresponding Certificate exists.

This language is, to my understanding, just a reification of the common position that has been taken by root programs for a number of years now. For example, the MRSP says:

> The logging of a precertificate in a Certificate Transparency log is considered by Mozilla to be a binding intent to issue a final certificate[.]

There is a slight difference between the current Mozilla policy and the proposed Profiles language: Mozilla says that "logging to CT" is a binding intent to issue, while the Profiles ballot says that just "signing" is a binding intent to issue. To be honest, I prefer the new language, but it introduces some weirdness regarding OCSP.

Section [4.9.10](https://github.com/cabforum/servercert/blob/2c63814/docs/BR.md#4910-on-line-revocation-checking-requirements) of the BRs says:

> A certificate serial number within an OCSP request is...
> - "assigned" if a Certificate with that serial number has been issued by the Issuing CA
> - "reserved" if a Precertificate [RFC6962] with that serial number has been issued[...].

It also says:

> The OCSP responder MAY provide definitive responses about "reserved" certificate serial numbers, as if there was a corresponding Certificate that matches the Precertificate [RFC6962].

But if, as the Profiles ballot text says, a "Precertificate is... a binding commitment... that a corresponding Certificate exists", then there's no actual difference between "reserved" and "assigned". All serials which are "reserved" may be treated by relying parties as actually "assigned", and therefore the OCSP responder MUST provide a definitive response for them.

This isn't a huge deal: as noted above, the new Profiles language largely just carries forward existing Mozilla language, so most publicly trusted CAs are already in this state today. But I think it would be worthwhile to clean up Section 4.9.10 to only list two categories of serials, to make the requirements clearer.

timfromdigicert:
[...]

PR #535, opened 2024-07-18
Ballot SC-76: Clarify and improve OCSP requirements

aarongable:
This ballot attempts to address three concerns:
- The confusion around "reserved" serials, which do not actually exist because all Precertificate serials are assumed to also exist in corresponding Certificates and are therefore actually "assigned";
- Confusion around whether, and how quickly, OCSP responders must begin providing authoritative responses for Certificates and Precertificates; and
- Confusion around whether and how the OCSP requirements apply to Certificates which do not contain an AIA OCSP URL, but for which the CA's OCSP responder is still willing to provide responses.

It addresses these concerns by:
- Stating that OCSP responses must be available within 15 minutes of signing a certificate containing an AIA OCSP URL;
- Removing the concept of a "reserved" serial entirely;
- Moving all OCSP requirements into Section 4.9.9, leaving Section 4.9.10 (which RFC 3647 says is meant to place requirements on relying parties, not on CAs) empty; and
- Organizing the requirements in Section 4.9.9 into three clusters:
  - Definitions of "validity interval", "assigned", and "unassigned";
  - Requirements on OCSP Responders, which apply only to responses from AIA OCSP URLs found in issued certs; and
  - Requirements on OCSP Responses, which apply to all responses regardless of whether the certificate in question has an AIA OCSP URL.

Addresses https://github.com/mozilla/pkipolicy/issues/280
Addresses https://github.com/cabforum/servercert/issues/422
Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1905419


docs/BR.md
```diff
@@ -1326,21 +1326,32 @@ No stipulation.
...
+
 OCSP responses MUST conform to RFC6960 and/or RFC5019. OCSP responses MUST either:
 
 1. Be signed by the CA that issued the Certificates whose revocation status is being checked, or
 2. Be signed by an OCSP Responder whose Certificate is signed by the CA that issued the Certificate whose
 revocation status is being checked.
 
-In the latter case, the OCSP signing Certificate MUST contain an extension of type `id-pkix-ocsp-nocheck`, as
-defined by RFC6960.
+In the latter case, the OCSP signing Certificate MUST contain an extension of type `id-pkix-ocsp-nocheck`, as defined by RFC6960.
```
CBonnell, line 1339: Consider requiring the delegated OCSP responder cert to comply with the OCSP responder cert profile in section 7.

This may require that some OCSP responder certifcates be replaced if issued prior to 2023-09-15, but that's probably not the worst thing since a OCSP responder key shouldn't be used for more than a year anyway.

docs/BR.md
```diff
@@ -1326,21 +1326,32 @@ No stipulation.
 
 The following SHALL apply for communicating the status of Certificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
 
+Authoritative OCSP responses MUST be available (i.e. the responder MUST NOT respond with the "unknown" status) starting no more than 15 minutes after the Certificate's `notAfter` timestamp.
```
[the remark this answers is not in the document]
CBonnell: Perhaps the obligation to provide OCSP responses should be tied to when the Certificate's serial number transitions from "unassigned" to "assigned". This would accommodate any backdating and yet ensure that responses are distributed timely.

docs/BR.md
```diff
@@ -1326,21 +1326,32 @@ No stipulation.
...
-defined by RFC6960.
+In the latter case, the OCSP signing Certificate MUST contain an extension of type `id-pkix-ocsp-nocheck`, as defined by RFC6960.
+
+OCSP responders operated by the CA SHALL support the HTTP GET method, as described in RFC 6960 and/or RFC 5019. The CA MAY process the Nonce extension (`1.3.6.1.5.5.7.48.1.2`) in accordance with RFC 8954.
+
+A certificate serial is "assigned" if:
+- a Certificate or Precertificate with that serial number has been issued by the Issuing CA, using any current or previous key associated with that CA subject; or
+- a Precertificate with that serial number has been issued by a Precertificate Signing Certificate, as defined in [Section 7.1.2.4](#7124-technically-constrained-precertificate-signing-ca-certificate-profile), associated with the Issuing CA.
+
+A certificate serial is "unused" if it is not "assigned".
```
CBonnell, line 1347: Since we're reworking terminology: consider changing "unused" to "unassigned" so that there is consistency in naming with the "assigned" serial number state.

docs/BR.md
```diff
@@ -1326,21 +1326,32 @@ No stipulation.
 
 The following SHALL apply for communicating the status of Certificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
 
+Authoritative OCSP responses MUST be available (i.e. the responder MUST NOT respond with the "unknown" status) starting no more than 15 minutes after the Certificate's `notAfter` timestamp.
+
+The following SHALL apply for communicating the status of *all* Certificates for which an OCSP responder is willing to respond.
```
XolphinMartijn, line 1331: What does "willing" mean? Could an OCSP response of always "unknown" be considered non-willing, and thus always be compliant?
```suggestion
The following SHALL apply for communicating the status of any certificate containing an Authority Information Access extension with an id-ad-ocsp accessMethod.
```

docs/BR.md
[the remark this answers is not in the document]
aarongable: > s/notAfter/notBefore/ perhaps?

Heh oops, thanks for catching this typo.

> Backdating the notBefore timestamp by more than 15 minutes would mean that the CA would have had to make available an authoritative OCSP response some number of minutes earlier than it possibly could have done.

This is a good point. I've always wondered why the BRs allow backdating by up to 48 hours (I've never seen a need to backdate more than 1hr, and even that feels generous).

> I think the "15 minutes" countdown should begin when the "certificate signing operation" occurs.

Yeah, I don't love this solution, but I think you're right that it's likely the best path forward. And at least the trail has been blazed somewhat by the existing backdating requirements referencing this same timestamp. I'll go this direction, thanks.

docs/BR.md
aarongable: Yeah, that's a good idea. This PR hasn't actually changed this line yet (it only appears here due to a line-wrapping thing) but I think that makes sense to fit with this same ballot.

docs/BR.md
aarongable: Good idea.

docs/BR.md
aarongable: Yes, that's the idea.

There are four possible situations:
- The cert contains an OCSP URL, and the responder provides responses. This is good.
- The cert contains an OCSP URL, but the responder does not provide responses (i.e. responds "unknown"). This is bad, and is covered by the clause above which says that responses MUST be available.
- The cert does not contain an OCSP URL, and the responder does not provide responses. This is fine, and is the whole point of Ballot SC-063 making CRLs mandatory and OCSP optional.
- The cert does not contain an OCSP URL, but the responder provides responses anyway. This _should be_ fine. But the current BRs have no provision for it. A responder which provides OCSP responses for certs which do not contain OCSP URLs is, per the BRs, allowed to make those OCSP responses wildly malformed. This is what this ballot is attempting to resolve, by stating that these requirements apply even to those responses.

aarongable: I've now completely rearranged this section, and I believe it should be much clearer now. Let me know what you think!

docs/BR.md
```diff
@@ -1324,50 +1324,41 @@ No stipulation.
 
 ### 4.9.9 On-line revocation/status checking availability
 
-The following SHALL apply for communicating the status of Certificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
+The validity interval of an OCSP response is the difference in time between the `thisUpdate` and `nextUpdate` field, inclusive. For purposes of computing differences, a difference of 3,600 seconds shall be equal to one hour, and a difference of 86,400 seconds shall be equal to one day, ignoring leap-seconds.
 
-OCSP responses MUST conform to RFC6960 and/or RFC5019. OCSP responses MUST either:
+A certificate serial is "assigned" if:
```
romanf, line 1329: I suggest to add an empty line before the bullet list to make it more compatible with renderers.

aarongable: Good catch, done, thanks!

docs/BR.md
```diff
@@ -1324,50 +1324,42 @@ No stipulation.
...
-3. For OCSP responses with validity intervals less than sixteen hours, then the CA SHALL update the information provided via an Online Certificate Status Protocol prior to one-half of the validity period before the nextUpdate.
-4. For OCSP responses with validity intervals greater than or equal to sixteen hours, then the CA SHALL update the information provided via an Online Certificate Status Protocol at least eight hours prior to the nextUpdate, and no later than four days after the thisUpdate.
+- For OCSP responses with validity intervals less than sixteen hours, the CA SHALL provide an updated OCSP response prior to one-half of the validity period before the nextUpdate.
+- For OCSP responses with validity intervals greater than or equal to sixteen hours, the CA SHALL provide an updated OCSP response at least eight hours prior to the nextUpdate, and no later than four days after the thisUpdate.
 
-For the status of Subordinate CA Certificates:
+For the status of Subordinate CA Certificates, the CA SHALL provide an updated OCSP response at least every twelve months, and within 24 hours after revoking the covered Subordinate CA Certificate.
```
wthayer, line 1347: I prefer "a" over "the covered" here. "the" implies one, and "covered" may be ambiguous. Otherwise LGTM. Thanks Aaron!

docs/BR.md
```diff
@@ -1324,50 +1324,42 @@ No stipulation.
 
 ### 4.9.9 On-line revocation/status checking availability
 
-The following SHALL apply for communicating the status of Certificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
+The validity interval of an OCSP response is the difference in time between the `thisUpdate` and `nextUpdate` field, inclusive. For purposes of computing differences, a difference of 3,600 seconds shall be equal to one hour, and a difference of 86,400 seconds shall be equal to one day, ignoring leap-seconds.
```
AEtherC0r3, line 1327: `nextUpdate` should not be counted in the validity interval of an OCSP response, because in section 3.2 of RFC6960 it is stated that `nextUpdate` must be in the future for the OCSP response to be valid

docs/BR.md
```diff
@@ -1324,50 +1324,42 @@ No stipulation.
...
 
-In the latter case, the OCSP signing Certificate MUST contain an extension of type `id-pkix-ocsp-nocheck`, as
-defined by RFC6960.
+A certificate serial is "unassigned" if it is not "assigned".
 
-### 4.9.10 On-line revocation checking requirements
+The following SHALL apply for communicating the status of Certificates and Precertificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
 
-The following SHALL apply for communicating the status of Certificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
+Authoritative OCSP responses MUST be available (i.e. the responder MUST NOT respond with the "unknown" status) starting no more than 15 minutes after the certificate signing operation occurs.
```
AEtherC0r3, line 1338: This requirement should be scoped to certificates issued by online CAs. It is not reasonable to expect that an OCSP response will be available for certificates issued by Root or other offline CAs within 15 minutes from signing.

docs/BR.md
aarongable: Changing the way the validity interval is computed is outside the scope of this ballot. In this ballot, this line is simply moved from its previous location, its contents are not changed at all.

docs/BR.md
aarongable: Yeah, I don't love the "the covered" language here either. I went that way because I was concerned that "a" could be interpreted as meaning that new OCSP responses for _all_ Subordinate CA Certificates must be available within 24 hours after _any_ of them is revoked. But I'm happy to change this back to "a" and let the official discussion period talk about it.

docs/BR.md
aarongable: Good point. I've scoped this to "subscriber certificates" since I don't think I want to get into the intricacies of online/offline in this section.

aarongable:
Thanks for the feedback, all! I've incorporated updates, as well as an effective date thanks to a request for such on the mailing list.

docs/BR.md
AEtherC0r3: This ballot could be a good opportunity to fix this as well

docs/BR.md
timfromdigicert: I think 'corresponding certificate' is perhaps more obvious. I also think these requirements read cleaner as singular statements about one instance, instead of plural collective statements, as the extension to a collection is obvious.

"For the status of a Subordinate CA Certificate, the CA SHALL provide an updated OCSP response at least every twelve months, and within 24 hours after revoking the corresponding Subordinate CA Certificate."

aarongable: Oh I like that singular version a lot. What do you think of simplifying it even further:
> For the status of a Subordinate CA Certificate, the CA SHALL provide an updated OCSP response at least every twelve months, and within 24 hours after revoking the Certificate.

docs/BR.md
```diff
@@ -1357,50 +1358,42 @@ No stipulation.
...
 
-In the latter case, the OCSP signing Certificate MUST contain an extension of type `id-pkix-ocsp-nocheck`, as
-defined by RFC6960.
+A certificate serial is "unassigned" if it is not "assigned".
 
-### 4.9.10 On-line revocation checking requirements
+The following SHALL apply for communicating the status of Certificates and Precertificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
 
-The following SHALL apply for communicating the status of Certificates which include an Authority Information Access extension with an id-ad-ocsp accessMethod.
+Effective 2025-01-15, authoritative OCSP responses for Subscriber Certificates MUST be available (i.e. the responder MUST NOT respond with the "unknown" status) starting no more than 15 minutes after the certificate signing operation occurs.
```
jdkasten, line 1372: > starting no more than 15 minutes after the certificate signing operation occurs.

Could this be modified to focus on the publication of the certificate as opposed to the certificate signing operation? 

1. It is difficult to know precisely when the certificate signing operation occurred from an external auditing standpoint (e.g. backdating the notBefore by a small amount is common).
2. There may be hiccups in the OCSP pipeline which don't affect the initial certificate signing operation. A CA which signs a precertificate, but is unable to push new OCSP for 15 minutes, shouldn't be penalized if the precertificate hasn't been sent to CT yet or relayed to the subscriber.

Proposed text: "starting no more than 15 minutes after the certificate is published or is made available to the subscriber.”

docs/BR.md
aarongable: I think it's unclear whether there's actually something to be fixed here, so I'd prefer not to bog this ballot down with that separate debate.

RFC 6960 has this to say:

> #### 2.4. Semantics of thisUpdate, nextUpdate, and producedAt
> nextUpdate: The time at or before which newer information will be available about the status of the certificate.

> #### 4.2.2.1 Time
> The thisUpdate and nextUpdate fields define a recommended validity interval.  This interval corresponds to the {thisUpdate, nextUpdate} interval in CRLs. Responses whose nextUpdate value is earlier than the local system time value SHOULD be considered unreliable.

Unfortunately, RFC 5280 has nothing to say about how clients should consume the `nextUpdate` field of CRLs, so there's no additional clarity to be found there.

So given the quotes above, we have the following problems:
1) The requirement in 4.2.2.1 is only a SHOULD, not a MUST.
2) The precedent set by the certificate's Validity Period being inclusive of the `notAfter` timestamp is that timestamps in X.509 documents represent the _whole_ time interval which could possibly be referenced by that timestamp, not the instant at the beginning of that interval. A `notAfter` timestamp of `2024-08-13 09:04:00` means that the certificate can be trusted throughout the whole `09:04:00` second, and only becomes untrusted when the clock rolls over to `09:04:01`. There's no indication that CRL and OCSP `nextUpdate` timestamps are to be treated any differently. If an OCSP response has a `nextUpdate` of `2024-08-13 09:04:00`, and that timestamp represents that _whole_ second and not just the instant at the beginning of that second, then a client whose clock says it is `09:04:00.5` is not looking at a "response whose nextUpdate value is earlier than the local system time value". It's looking at a response whose nextUpdate _contains_ the local system time, and is therefore under no obligation to treat the response as unreliable.

Look, I strongly dislike the interpretation of X.509 timestamps as being inclusive of their whole interval. I think it is unnecessarily pedantic and highly unintuitive. But the fact is that precedent for this interpretation is so well established that people decided to encode it directly into the BRs to prevent any future confusion. While I would be personally in favor of a ballot that does away with inclusive timestamps, that ballot should tackle OCSP, CRLs, and Certificates together. This ballot is not the right place to make that change.

AEtherC0r3: It is in section 3.2 of RFC 6960:
> [3.2](https://datatracker.ietf.org/doc/html/rfc6960#section-3.2).  Signed Response Acceptance Requirements
> 
>    Prior to accepting a signed response for a particular certificate as
>    valid, OCSP clients SHALL confirm that:
>
>    6. When available, the time at or before which newer information will
>       be available about the status of the certificate (nextUpdate) is
>       greater than the current time.

i.e. For an OCSP response to be valid the following inequality must hold: `nextUpdate > currentTime`
Unfortunately, this only applies to OCSP responses.

I don't mind if we don't change it. Just thought it would be nice to make this sentence more accurate.

docs/BR.md
timfromdigicert: I like it

docs/BR.md
aarongable: Intriguing. I like this idea, but have some hesitation. Two questions:

1. While the publication date can be taken to be "SCT + MMD", not all certificates are published to CT logs, and it's similarly impossible for an external auditor to determine when the certificate is "made available to the subscriber". Does this phrasing truly solve the difficulty, or just move it to a different location?
2. If we were to make this change, how do you think it should interact with the limitation of this requirement to just Subscriber Certificates? If we change this to be within 15 minutes of publication, then it could potentially be applied to CA certificates as well.

XolphinMartijn: Do we need additional clarification that this also applies for the moment a precertificate is signed?

aarongable: I'm going to go with 
> ... starting no more than 15 minutes after the Certificate or corresponding Precertificate is published or otherwise made available."

aarongable:
Could someone add the `ballot` tag to this PR, so that it gets reflected on https://cabforum.org/working-groups/server/ballots/ ? Thanks!

barrini (approved):

PR #545, opened 2024-09-10
SC-078 - Subject organizationName alignment for DBA / Assumed Name

XolphinMartijn:


XolphinMartijn:
Thanks @CBonnell. I copied that part from the SBRs which do have the definition. Your suggestion looks good!

PR #548, opened 2024-09-16
Sunset WHOIS [Version .1]

ryancdickson:
All,

In light of recent events where research from WatchTowr Labs demonstrated how threat actors could exploit WHOIS to obtain fraudulently issued TLS certificates [1] and follow-on discussions in MDSP [2][3], we drafted an introductory proposal [4] to sunset the use of WHOIS for identifying Domain Contacts.

The proposal sets a prohibition against relying on WHOIS to identify Domain Contacts beginning 11/1/2024.

While publicly-trusted CA Owners are required to disclose and maintain in-use DCV methods to the CCADB [5], the collected data lacks specificity, hindering our ability to assess the extent of reliance on WHOIS and the potential impact of transitioning away from it.

Feedback on the proposal (preferably using comments or suggestions on this Pull Request via GitHub) along with volunteers for endorsers would be appreciated.

Thanks,
Ryan

P.S., I apologize if this effort is redundant to discussions already taking place in the Forum, I was traveling last week and am catching up on email.

[1] https://labs.watchtowr.com/we-spent-20-to-achieve-rce-and-accidentally-became-the-admins-of-mobi/
[2] https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/FuOi_uhQB6U
[3] https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/mAl9XjieSkA
[4] https://github.com/cabforum/servercert/compare/main...ryancdickson:staging:sunset-whois?expand=1
[5] https://docs.google.com/spreadsheets/d/1IXL8Yk12gPQs8GXiosXCPLPgATJilaiVy-f9SbsMA28/edit?gid=268412787#gid=268412787

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1,11 +1,11 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 
-subtitle: Version 2.0.7
+subtitle: Version 2.0.X
 author:
   - CA/Browser Forum
 
-date: 6-September-2024  
+date: DD-MONTH-2024  
 
 
 
@@ -144,6 +144,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.0.5 | SC73 | Compromised and weak keys | 3-May-2024 | 1-July-2024 |
 | 2.0.6 | SC75 | Pre-sign linting | 28-June-2024 | 6-August-2024 |
 | 2.0.7 | SC67 | Require Multi-Perspective Issuance Corroboration | 2-August-2024 | 6-September-2024 |
+| 2.0.X | SCXX | Sunset use of WHOIS to identify Domain Contacts | TBD | TBD |
 
 \* Effective Date and Additionally Relevant Compliance Date(s)
 
@@ -197,10 +198,11 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2023-09-15     | Section 7 (and others)    | CAs MUST use the updated Certificate Profiles passed in Version 2.0.0                                                                                                                                                                                                                                                                                                                                                                                    |
 | 2024-03-15     | 4.9.7                     | CAs MUST generate and publish CRLs.                                                                                                                                                                                                                                                                                                                                                                                                                      |
 | 2024-09-15     | 4.3.1.2                   | The CA SHOULD implement a Linting process to test the technical conformity of the to-be-issued Certificate with these Requirements.                                                                                                                                                                                                                                                                                                                      |
+| 2024-11-1      | 3.2.2.4                   | CAs MUST NOT rely on WHOIS to identify Domain Contacts. |
+
 | 2025-03-15     | 4.3.1.2                   | The CA SHALL implement a Linting process to test the technical conformity of the to-be-issued Certificate with these Requirements.                                                                                                                                                                                                                                                                                                                       |
 | 2025-03-15     | 8.7                       | The CA SHOULD use a Linting process to test the technical accuracy of already issued Certificates against the sample set chosen for Self-Audits.                                                                                                                                                                                                                                                                                                         |
 | 2025-03-15     | 3.2.2.9                   | CAs MUST corroborate the results of domain validation and CAA checks from multiple Network Perspectives where specified. |
-
 
 ## 1.3 PKI Participants
 
@@ -749,6 +751,8 @@ The Random Value SHALL remain valid for use in a confirming response for no more
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
+Effective November 1, 2024, validations using this method MUST NOT rely on WHOIS to identify Domain Contact information.
+
 ##### 3.2.2.4.3 Phone Contact with Domain Contact
 
 This method has been retired and MUST NOT be used. Prior validations using this method and validation data gathered according to this method SHALL NOT be used to issue certificates.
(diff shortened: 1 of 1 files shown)


shaver:
This is nice and succinct. Question: should it include guidance for what CAs should do in terms of audit or revalidation, if they have previously issued under whois DCV? Should renewal require revalidation with a permitted method? Something more aggressive like requiring revalidation by Jan 1, 2025? Unless the subscriber is unable to revalidate under an approved method, I don't think it would necessitate revocation and reissuance; if the subscriber *isn't* able to complete DCV with an approved method, then revocation is probably appropriate!

(I don't see general language in the BRs about what to do when renewing certs that were issued under now-prohibited validation methods; maybe this is a more general concern that deserves a separate PR; happy to author that if so.)

docs/BR.md
```diff
@@ -749,6 +751,8 @@ The Random Value SHALL remain valid for use in a confirming response for no more
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
+Effective November 1, 2024, validations using this method MUST NOT rely on WHOIS to identify Domain Contact information.
```
CBonnell, line 754: The language needs to state whether reuse of previous validation is allowed or not. Otherwise, it is unclear.

docs/BR.md
```diff
@@ -771,6 +775,8 @@ The Random Value SHALL remain valid for use in a confirming response for no more
...
```
CBonnell, line 778: This method has already been banned, so this text is not needed.

dougbeattie:
I have a similar question:  Does this statement mean that no new domains may be validated starting on that date or that all existing domains validated via the methods you've listed must be disabled on that date?:

- Effective November 1, 2024, validations using this method MUST NOT rely on WHOIS to identify Domain Contact information.


And I don't think this method uses WHOIS data at all, so I recommend we remove the statement:

- 3.2.2.4.4 Constructed Email to Domain Contact

CBonnell (commented):
Cleaning up Domain Contact and Domain Name Registrant terms would also be helpful as part of this initiative.

Happy to open a PR against this branch if there's agreement that this would be useful.

XolphinMartijn:
@dougbeattie 
>And I don't think this method uses WHOIS data at all, so I recommend we remove the statement:
> 3.2.2.4.4 Constructed Email to Domain Contact

It does. WHOIS is part of the "Domain Contact" definition, which calls in the allowance. 

Would recommend removing WHOIS from that definition, as Corey is suggesting, but that makes the Effective Date callout a bit harder

docs/BR.md
XolphinMartijn: "validations" strongly suggests "validation" events, not "issuance" events ;)

ryancdickson: This seems to intersect with comments from @shaver (thank you!) and @dougbeattie (thank you!), responding here as its the latest.

**From:** Effective November 1, 2024, validations using this method MUST NOT rely on WHOIS to identify Domain Contact information.

**To:** Effective November 1, 2024, Subscriber Certificates MUST NOT be issued that rely on WHOIS to identify Domain Contact information. Prior validations using WHOIS to identify Domain Contact information MUST NOT be used to issue new Subscriber Certificates, regardless of permitted data reuse periods.

Thoughts? (would be happy to see this less verbose, but also in a way that's not subject to misinterpretation)

shaver: I don't think you need the "regardless" clause, because MUST NOT is absolute. But will this be clear enough that it affects "renewals" as well as issuance for new domains?

XolphinMartijn: @ryancdickson This does indeed clarify that it's based on "at issuance" rather than "at DCV".

@shaver "renewal" of a certificate is considered an "issuance" event, which would invoke this prohibition

barrini: Just a minor comment regarding the effective dates. Have in mind that we need at least 7 days for discussion when the ballot is ready, another 7 for voting and then 1 month for IPR review. It´s almost impossible to have the version published by that date and be effective.

ryancdickson: @barrini  - Today + 44 days = October 30, 2024?

barrini: If everything goes smooth, there are no discussions/changes, the voting is perfect, there´re no weekends when need to release the versions, etc. yes, you´re right but I think it´s too tight.

CBonnell: +up to 3 business days for the Chair to publish the IPR "all clear" and final version of the document(s). To avoid a repeat of the confusion highlighted during the [discussions of SC-46](https://lists.cabforum.org/pipermail/servercert-wg/2021-May/002733.html), an immediate effective date (i.e., don't specify any effective date) would be clearer.

ryancdickson: Thanks for the additional context, @CBonnell.

To confirm my understanding: If I were to change...

**From:** "Effective November 1, 2024, Subscriber Certificates MUST NOT be issued that 1) rely on WHOIS to identify Domain Contact information or 2) reuse validation data where WHOIS was used to identify the Domain Contact."

**To:** "Subscriber Certificates MUST NOT be issued that 1) rely on WHOIS to identify Domain Contact information or 2) reuse validation data where WHOIS was used to identify the Domain Contact."

While also...
- updating the Ballot preamble to call out the new requirements would be effective immediately upon publication.
- and, relying on the SCWG Chair to set the effective date during publication.

That'd be an acceptable approach from your perspective?

CBonnell: If the desire is to propose a ballot with a November 1st (or thereabouts) effective date, the steps you outlined are how I would recommend writing the ballot. Doing so avoids any SC46-esque confusion about effective dates in the event the BR is published on (changes come into effect midnight UTC) or after the stated effective date.

barrini: Correct, this way, when the new version is published, in the effectiveness column of the new version is set, usually, the date of the publication, so once is published is effective. But, if you want to set a specific date and start counting from there for statistics for example, or just to know which CAs are still using that method, then set a date with enough time, let´s say Dec 1st.

PR #549, opened 2024-09-16
SC-080: Sunset WHOIS (Version 1)

ryancdickson:
**Purpose of Ballot SC-080 V1:**
 
This Ballot proposes updates to the Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates (i.e., TLS BRs) related to sunsetting the use of WHOIS when identifying Domain Contacts.

**Background:**

In light of recent events where research from WatchTowr Labs demonstrated how threat actors could exploit WHOIS to obtain fraudulently issued TLS certificates [1] and follow-on discussions in MDSP [2][3], we drafted an introductory proposal [4] to sunset the use of WHOIS for identifying Domain Contacts.

The proposal sets a prohibition against relying on WHOIS to identify Domain Contacts beginning 11/1/2024.

**Proposal Revision History:**
- Pre-Ballot Version 1 [4]
 
**Previous Versions of this Ballot:**
- N/A 

**References**:
[1] https://labs.watchtowr.com/we-spent-20-to-achieve-rce-and-accidentally-became-the-admins-of-mobi/
[2] https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/FuOi_uhQB6U
[3] https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/mAl9XjieSkA
[4] https://github.com/cabforum/servercert/pull/548
[5] https://docs.google.com/spreadsheets/d/1IXL8Yk12gPQs8GXiosXCPLPgATJilaiVy-f9SbsMA28/edit?gid=268412787#gid=268412787

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -751,7 +751,7 @@ The Random Value SHALL remain valid for use in a confirming response for no more
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
-Effective November 1, 2024, Subscriber Certificates MUST NOT be issued that rely on WHOIS to identify Domain Contact information. Prior validations using WHOIS to identify Domain Contact information MUST NOT be used to issue new Subscriber Certificates, regardless of permitted data reuse periods.
+Effective November 1, 2024, Subscriber Certificates MUST NOT be issued that 1) rely on WHOIS to identify Domain Contact information or 2) reuse validation data where WHOIS was used to identify the Domain Contact.  
 
 ##### 3.2.2.4.3 Phone Contact with Domain Contact
 
@@ -822,7 +822,7 @@ Confirming the Applicant's control over the FQDN by validating the Applicant is 
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
-Effective November 1, 2024, Subscriber Certificates MUST NOT be issued that rely on WHOIS to identify Domain Contact information. Prior validations using WHOIS to identify Domain Contact information MUST NOT be used to issue new Subscriber Certificates, regardless of permitted data reuse periods.
+Effective November 1, 2024, Subscriber Certificates MUST NOT be issued that 1) rely on WHOIS to identify Domain Contact information or 2) reuse validation data where WHOIS was used to identify the Domain Contact.
 
 ##### 3.2.2.4.13 Email to DNS CAA Contact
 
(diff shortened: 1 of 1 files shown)


wthayer:
Based on the comments from Francisco Arias on the list and my knowledge of this method and the vulnerability being addressed, I believe this ballot's scope should be reduced to only deprecate information retrieved in an RFC 3912 query.

timfromdigicert:
Yes, we’ve been talking to ICANN about this as well and I agree.

 

If we can get Andy added as a member representative for ICANN so he can participate in these discussions that would be really helpful.

 

-Tim

ryancdickson:
@wthayer / @timfromdigicert - Do you have an opinion on continuing to permit the use of an "HTTPS Website" as a source of WHOIS data?

> WHOIS: Information retrieved directly from the Domain Name Registrar or registry operator via the protocol defined in RFC 3912, the Registry Data Access Protocol defined in RFC 7482, **or an HTTPS website**.

My understanding of the WatchTowr report was that online WHOIS tools (i.e., HTTPS websites hosting tools intended to perform WHOIS lookups) were also susceptible to relying on stale information.

XolphinMartijn:
> Do you have an opinion on continuing to permit the use of an "HTTPS Website" as a source of WHOIS data?

@ryancdickson While not tagged, I'll inject my opinion here. These sites quiet often are behind a captcha, and have (severe) rate limits. Our experience tells us these in general are not viable for automated lookups. 

When it's not automated, it means manual, so a CA employee would need to lookup if an email address is within the WHOIS data, and confirm back. This is what we generally see as manual DCV, and opens the door to human errors.

wthayer:
I think Andrew's point about the likelihood of the "HTTPS Website" using an insecure RFC 3912 mechanism on the back end is valid, and I suspect that CAs could easliy switch to performing manual lookups with an RDAP command-line client instead of a browser, so I'd lean toward killing this option.

shaver:
I think that the "Domain Contact" and "Domain Name Registrant" entries in 1.6.1 should also be updated to not promote WHOIS, if not also to say how it *is* acceptable to validate those.

(Do we really feel that RDAP APIs aren't going to be backed by RFC-3912 queries just like the web sites are? Why would RDAP not have rate limits like the HTTPS ones do? WHOIS is a broken wire protocol, to be sure, but it would be nice to be confident that accessible registrar data is reliable for DCV purposes before we push people to migrate to another way of accessing that data over the wire.)

defacto64:
I have a doubt. Let's look at the definition of Domain Contact in BRs:

> **Domain Contact**: The Domain Name Registrant, technical contact, or administrative contact
> (or the equivalent under a ccTLD) as listed in the WHOIS record of the Base Domain Name or
> in a DNS SOA record, or as obtained through direct contact with the Domain Name Registrar.

Since the changes proposed in this pull request do not modify the definition above, I assume that - while "CAs MUST NOT rely on WHOIS to identify Domain Contacts" (that I am fine with)  - a CA is still allowed to rely on other ways to identify Domain Contacts, such as e.g. "through direct contact with the Domain Name Registrar" (from the definition above).

If my interpretation is correct, and there is consensus, then there is no need to discuss this any further.

If, however, my interpretation is incorrect, and you/people actually want to also deprecate the case "through direct contact with the Domain Name Registrar", then I think it is necessary to clarify this, and this pull request should also include a change to the definition mentioned above.

ryancdickson:
Closing this pull request in favor of an updated version (coming soon) given feedback via this Pull Request and https://github.com/ryancdickson/staging/pull/9 .

ryancdickson:
Now superseded by https://github.com/cabforum/servercert/pull/551

PR #561, opened 2024-11-14
SC-083: Winter 2024-2025 Cleanup Ballot

XolphinMartijn:
This cleanup ballot / PR cleans up a large number of pending issues, for which I will let the Commit history show the details below.

Additionally, it removes the State/Province and Country border requirements for MPIC as proposal, which has been discussed with a few CABF members


docs/BR.md
```diff
@@ -784,24 +782,26 @@ This method has been retired and MUST NOT be used. Prior validations using this
...
+  1. 30 days; or
+  2. if the Applicant submitted the Certificate request, the time frame permitted for reuse of validated information relevant to the Certificate (such as in [Section 4.2.1](#421-performing-identification-and-authentication-functions) of these Guidelines or Section 3.2.2.14.3 of the EV Guidelines).
 
-CAs using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e. Random Value or Request Token) as the Primary Network Perspective.
+Validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e. Random Value or Request Token) as the Primary Network Perspective.
```
CBonnell, line 794: This comment is applicable to all methods where this language was changed: I think we need tweak this language, because validations do not implement anything. How about:

"CAs performing validations using this method MUST implement..."

docs/BR.md
```diff
@@ -841,7 +841,7 @@ Each email MAY confirm control of multiple FQDNs, provided that each email addre
 
 The Random Value SHALL be unique in each email. The email MAY be re-sent in its entirety, including the re-use of the Random Value, provided that its entire contents and recipient(s) SHALL remain unchanged. The Random Value SHALL remain valid for use in a confirming response for no more than 30 days from its creation. The CPS MAY specify a shorter validity period for Random Values.
 
-CAs using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the selected contact address used for domain validation observed by the Primary Network Perspective.
+Validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e. Random Value or Request Token) as the Primary Network Perspective.
```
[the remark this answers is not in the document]
CBonnell: My comment is regarding "challenge information (i.e. Random Value or Request Token)". This information is not found in the DNS (or on any webpage) for this method.

docs/BR.md
```diff
@@ -1124,10 +1124,10 @@ If any of the above considerations are performed by a Delegated Third Party, the
...
-- *Effective December 15, 2026*, the CA MUST implement Multi-Perspective Issuance Corroboration using at least five (5) remote Network Perspectives. The CA MUST NOT proceed with certificate issuance if the number of non-corroborations is greater than allowed in the Quorum Requirements table and if the remote Network Perspectives that do corroborate the determinations made by the Primary Network Perspective do not fall within the service regions of at least two (2) distinct Regional Internet Registries.
+- *Effective September 15, 2025*, the CA MUST implement Multi-Perspective Issuance Corroboration using at least two (2) remote Network Perspectives. The CA MUST only proceed with certificate issuance if the requirements defined in Quorum Requirements Table are satisfied.
```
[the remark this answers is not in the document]
CBonnell: How about:

```
The CA MUST ensure that the requirements defined in Quorum Requirements Table are satisfied. If the requirements are not satisfied, then the CA MUST NOT proceed with issuance of the Certificate.
```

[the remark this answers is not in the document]
XolphinMartijn: I like that. Committed to all timeline items

docs/BR.md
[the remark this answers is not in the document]
XolphinMartijn: Gotcha. Thanks. Looks like I got a bit too happy with the copy-paste there. All items updates

XolphinMartijn:
Added clarifying language based on earlier discussions that IDNA2003 and IDNA2008 are allowed. Thanks @CBonnell for the proposal

docs/BR.md
```diff
@@ -2681,6 +2681,8 @@ Table: `GeneralName` within a `subjectAltName` extension
 | `iPAddress`                 | Y             | The entry MUST contain the IPv4 or IPv6 address that the CA has confirmed the Applicant controls or has been granted the right to use through a method specified in [Section 3.2.2.5](#3225-authentication-for-an-ip-address). The entry MUST NOT contain a Reserved IP Address. |
 | `registeredID`              | N             | -              |
 
+**Note**: The allowance for including P-Labels that do not conform to IDNA 2003 is an exception to best practice as defined in RFC 5280. These Requirements allow for the inclusion of P-Labels that do not conform with IDNA 2003 to support newer versions of the Unicode character repertoire, among other improvements to the various IDNA standards.
```
robstradling, line 2684: @XolphinMartijn @CBonnell: Framing this as "an exception to best practice" seems wrong to me.  At IETF, "best practice" has evolved - see RFC8399, which updates RFC5280 and permits IDNA2008.
CABForum documents don't automatically recognise IETF document updates or obsoletions (and I think that's a feature, not a bug), but this doesn't mean we can never choose to recognise new IETF documents going forwards!
In this case, ISTM that the appropriate way forward is for the BRs to be updated to recognise RFC8399.  i.e., references to RFC5280 should become "RFC5280, as updated by RFC8399".

CBonnell: Hi @robstradling, so if I'm understanding correctly, you're suggesting changing:

> The allowance for including P-Labels that do not conform to IDNA 2003 is an exception to best practice as defined in RFC 5280.  These Requirements allow for the inclusion of P-Labels that do not conform with IDNA 2003 to support newer versions of the Unicode character repertoire, among other improvements to the various IDNA standards.

to:

> The allowance for including P-Labels that do not conform to IDNA 2003 and IDNA 2008 is an exception to best practice as defined in RFC 5280 and RFC 8399.  These Requirements allow for the inclusion of P-Labels that do not conform with IDNA 2003 and IDNA 2008 to support Unicode Technical Standard #46 and its variants, among other changes to the various IDNA standards.

robstradling: Hi @CBonnell.  Actually I was trying to suggest that (1) the current best practice, as defined by "RFC 5280, as updated by RFC8399", is to permit P-Labels that conform to IDNA 2003 and/or IDNA 2008, and to reject all other Reserved LDH Labels; and that (2) the BRs should be updated to regard "RFC 5280, as updated by RFC8399" as best practice, rather than as an exception to best practice.
Does that make sense?

CBonnell: I cribbed the "exception to best practice..." language from section 7.2.2:

> Note: The CA SHOULD update the revocation date in a CRL entry when it is determined that the
private key of the Certificate was compromised prior to the revocation date that is indicated in the
CRL entry for that Certificate. Backdating the revocationDate field is an exception to best practice
described in RFC 5280 (Section 5.3.2); however, these requirements specify the use of the
revocationDate field to support TLS implementations that process the revocationDate field as the
date when the Certificate is first considered to be compromised.

I think it's rather clear that the text is establishing the RFC(s) as best practice and some allowance within the BRs to be an exception to such best practice and not the other way around. But if you have improved concrete language that resolves the concern, perhaps that be added.

robstradling: Hi @CBonnell.  I think we're still talking past each other a bit.  Let me try to restate fully where I'm coming from...

Sectigo's view of the current situation is that:
- RFC5280 only permits IDNA 2003 to be used.
- The BRs currently require conformance to RFC5280.
- At IETF, RFC5280 has been updated by RFC8399, which adds permission for IDNA 2008 to be used.
- However, at CABForum, IETF document updates are not automatically incorporated into the BRs.
- The BRs were updated a while ago such that the LDH Label, Non-Reserved LDH Label, and XN-Label definitions now incorporate quotes from RFC5890, which is one of the documents that constitute IDNA 2008.  However, we don't believe that merely quoting some definitions from an RFC means that that entire RFC is assumed to be incorporated into the BRs, especially when that would conflict with other BR requirements (e.g., the requirement to conform to RFC5280, which only permits IDNA 2003).
- We can accept that the intention of the SC48v2 author/endorsers was to permit IDNA 2008, and so we feel that it's appropriate for a clean-up ballot to improve the language in line with that intent.

The gist of how I'd like to see the BRs updated:
- Explicitly permit IDNA 2008, by explicitly incorporating RFC8399 in the BRs.
- Allow the definition of "best practice" to move with the times, and not get stuck with updates to RFC5280 being labelled as "exceptions" to best practice.

I think the best way to achieve this is to update the BRs as follows:

1. Update the "P-Label" definition along these lines...
```
P‐Label: A XN‐Label that contains valid output of the Punycode algorithm (as defined in RFC 3492,
Section 6.3) from the fifth and subsequent positions.  These Requirements permit P-Labels that conform to RFC 5280 as updated by RFC 8399, meaning that both IDNA 2003 and IDNA 2008 may be used.
```

2. Add RFC 8399 to section 1.6.3 (References).

Then, when section 7.1.2 says "Except as explicitly noted, all normative requirements imposed by RFC 5280 shall apply, in addition to the normative requirements imposed by this document", this updated definition of P-Label would be instance of "Except as explicitly noted", and so the "normative requirements imposed by RFC 5280" that only permit IDNA 2003 would be overriden.

CBonnell: Hi @robstradling, I believe that the current language is sufficiently clear from a compliance standpoint that IDNA 2008 (and others) are allowed (the definition of P-Label and associated requirements are quite explicit on this point), but I agree that the language can be improved to aid readers.

> P‐Label: ...  These Requirements permit P-Labels that conform to RFC 5280 as updated by RFC 8399, meaning that both IDNA 2003 and IDNA 2008 may be used.

This language is problematic, as it is now unclear whether XN-labels that conform with UTS#46 (or other variants) but not IDNA 2008 are allowed, whereas it is clear that they are allowed in the current language. This, coupled with RFC 8399's SHOULD-level requirement for adherence to IDNA 2008, especially muddies the waters.

This is why I suggested the language in my previous post: it gives a nod to RFC 5280 and 8399 but does not establish them as normative in this area, because doing so adds confusion. I think the clearest course is to clearly establish what is allowed directly in the BRs and add informative references to the RFCs.

robstradling: Hi @CBonnell.  I still don't agree that the current language explicitly permits IDNA 2008 (and others), but it looks like we'll have to agree to disagree on that point.
TBH, I'm not familiar with IDNs that are neither IDNA 2003 nor IDNA 2008 but are instead "UTF#46 (or other variants)", so I can't comment on whether or not the BRs should permit them.  You make a good point that "RFC 8399's SHOULD-level requirement for adherence to IDNA 2008" would impact whether or not they're permitted. 
 I suppose it would also impact IDNA 2003 IDNs that don't conform to IDNA 2008.
So I understand now that you don't consider post-RFC5280 IETF "best practice" to be a suitable baseline for the BRs to use, which is why you want to frame the language as "an exception to best practice as defined in RFC 5280".  I guess I could live with that.
I do agree though that "the clearest course is to clearly establish what is allowed directly in the BRs and add informative references to the RFCs".

docs/BR.md
```diff
@@ -1079,11 +1079,11 @@ The set of responses from the relied upon Network Perspectives MUST provide the
...
+A Network Perspective MAY use a recursive DNS resolver that is NOT co-located with the Network Perspective. However, the DNS resolver used by the Network Perspective MUST fall within the same Regional Internet Registry service region as the Network Perspective relying upon it. Furthermore, for any pair of DNS resolvers used on a Multi-Perspective Issuance Corroboration attempt, the straight-line distance between the two DNS resolvers reside in MUST be at least 500 km. The location of a DNS resolver is determined by the point where unencapsulated outbound DNS queries are typically first handed off to the network infrastructure providing Internet connectivity to that DNS resolver.
```
ryancdickson, line 1082:
```suggestion
A Network Perspective MAY use a recursive DNS resolver that is NOT co-located with the Network Perspective. However, the DNS resolver used by the Network Perspective MUST fall within the same Regional Internet Registry service region as the Network Perspective relying upon it. Furthermore, for any pair of DNS resolvers used on a Multi-Perspective Issuance Corroboration attempt, the straight-line distance between the two DNS resolvers MUST be at least 500 km. The location of a DNS resolver is determined by the point where unencapsulated outbound DNS queries are typically first handed off to the network infrastructure providing Internet connectivity to that DNS resolver.
```

docs/BR.md
XolphinMartijn:
```suggestion
**Note**: As an explicit exception from RFC 5280, P-Labels are permitted to not conform to IDNA 2003. These Requirements allow for the inclusion of P-Labels that do not conform with IDNA 2003 to support newer versions of the Unicode character repertoire, among other improvements to the various IDNA standards.
```

docs/BR.md
```diff
@@ -784,24 +782,26 @@ This method has been retired and MUST NOT be used. Prior validations using this
...
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
 ##### 3.2.2.4.8 IP Address
 
 Confirming the Applicant's control over the FQDN by confirming that the Applicant controls an IP address returned from a DNS lookup for A or AAAA records for the FQDN in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
 
-CAs using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same IP address as the Primary Network Perspective.
+CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e. Random Value or Request Token) as the Primary Network Perspective.
```
XolphinMartijn, line 802:
```suggestion
CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same IP address as the Primary Network Perspective.
```

dzacharo (approved):
IPR review period was completed without notices.

Issue #423, opened 2023-02-20
Remove specific version in the WebTrust reference in section 1.6.3


dzacharo:
Section 1.6.3 contains references to specific versions of documents such as the NetSec and WebTrust. We could probably keep the specific version number reference to NetSec but it would make sense to remove the version from the WebTrust reference to be consistent with the ETSI reference. Section 8.4 covers the version issue for WebTrust and ETSI anyway.

PR #565, opened 2024-12-17
Remove NCSSR Version

clintwilson:
This was initially completed in https://github.com/cabforum/servercert/pull/460, to address #423, but a Version Number reference remains.


dzacharo:
@clintwilson reading #423 again, I don't think it was supposed to remove the NetSec version, just the WebTrust in order to align with the ETSI reference which didn't have a specific version.

PR #567, opened 2025-01-02
SC-098: Process RFC 8657 CAA Parameters

### LINKED ISSUE #353: Define standard CAA semantics for limiting cert issuance

wthayer:
Update 3.2.2.8 to require that CAs process CAA accounturi and validationmethod parameters defined in RFC 8657

Fixes https://github.com/cabforum/servercert/issues/353


docs/BR.md
```diff
@@ -1082,6 +1082,11 @@ CAs MAY check CAA records at any other time.
 
 When processing CAA records, CAs MUST process the issue, issuewild, and iodef property tags as specified in RFC 8659, although they are not required to act on the contents of the iodef property tag. Additional property tags MAY be supported, but MUST NOT conflict with or supersede the mandatory property tags set out in this document. CAs MUST respect the critical flag and not issue a certificate if they encounter an unrecognized property tag with this flag set.
 
+EFFECTIVE DD-MM-YYY:
+When processing CAA records, CAs MUST process the accounturi and validationmethods parameters as specified in RFC 8657. In addition:
+* When the certificate request does not use the ACME protocol defined in RFC 8555, the CA MUST define the recognized format of the accounturi in their CPS.
+* The CA MUST only recognize validationmethods parameters in the format "tlsbr-" + the 3.2.2.4 subsection number that defines the permitted validation method, e.g. "tlsbr-18" represents Agreed-Upon Change to Website v2.
```
robstradling, line 1088: Your proposed "tlsbr-" prefixes can only conform to [RFC 8657 section 4](https://www.rfc-editor.org/rfc/rfc8657#name-extensions-to-the-caa-record) if each "tlsbr-" validation method label is registered with the [ACME Validation Methods IANA registry](https://www.iana.org/assignments/acme/acme.xhtml#acme-validation-methods).

However, [RFC 8555 section 9.7.8](https://datatracker.ietf.org/doc/html/rfc8555#section-9.7.8) requires that _"When evaluating a request for an assignment in this registry, the designated expert should ensure that the method being registered has a clear, interoperable definition"_.  The TLS BR DCV methods are "clear" insofar as they specify baseline requirements for DCV, but I wouldn't say that they're "interoperable".  For example, ACME dns-01 and Sectigo's non-ACME DNS-based DCV mechanisms are not interoperable, and yet these are all covered by the TLS BR "3.2.2.4.7 DNS Change" method.  Also, since dns-01 is already present in that IANA registry, ISTM that the _"does not overlap with existing validation methods"_ requirement would be a roadblock to registering a "tlsbr-7" label.

Furthermore, requiring that only "tlsbr-" labels be recognised by the CA would mean that the primary ACME use cases described by RFC 8657 (e.g., "validationmethods=dns-01") would no longer be permitted!

Note that to avoid this IANA registry requirement, [RFC 8657 section 4](https://www.rfc-editor.org/rfc/rfc8657#name-extensions-to-the-caa-record) requires that _'the CA MUST use labels beginning with the string "ca-", which are defined to have CA-specific meaning.'_  I think "CA-specific" is entirely appropriate for the use case we're discussing in this PR, because each CA takes the baseline requirements for DCV (described in the 3.2.2.4 subsections) and creates their own specific implementation that is interoperable (between the CA and their Subscribers).

I propose changing this line as follows:
```suggestion
* The CA MUST define each recognized validationmethods label, along with the corresponding 3.2.2.4 subsection number, in their CPS.
```

robstradling (dismissed):

docs/BR.md
```diff
@@ -1082,6 +1082,11 @@ CAs MAY check CAA records at any other time.
 
 When processing CAA records, CAs MUST process the issue, issuewild, and iodef property tags as specified in RFC 8659, although they are not required to act on the contents of the iodef property tag. Additional property tags MAY be supported, but MUST NOT conflict with or supersede the mandatory property tags set out in this document. CAs MUST respect the critical flag and not issue a certificate if they encounter an unrecognized property tag with this flag set.
 
+EFFECTIVE DD-MM-YYY:
+When processing CAA records, CAs MUST process the accounturi and validationmethods parameters as specified in RFC 8657. In addition:
+* When the certificate request does not use the ACME protocol defined in RFC 8555, the CA MUST define the recognized format of the accounturi in their CPS.
```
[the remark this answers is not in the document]
wthayer: Agreed, your wording is better!

docs/BR.md
wthayer: Agreed that whatever we do has to begin with 'ca-' to comply with 8657. Your proposal simplifies this requirement at the expense of making the semantics CA-specific, potentially creating confusion and hindering adoption. Since I have yet to devise a solution to the dns-01 vs ca-tlsbr-7 confusion, I will propose your approach. Thank you!

docs/BR.md
```diff
@@ -1082,6 +1086,10 @@ CAs MAY check CAA records at any other time.
 
 When processing CAA records, CAs MUST process the issue, issuewild, and iodef property tags as specified in RFC 8659, although they are not required to act on the contents of the iodef property tag. Additional property tags MAY be supported, but MUST NOT conflict with or supersede the mandatory property tags set out in this document. CAs MUST respect the critical flag and not issue a certificate if they encounter an unrecognized property tag with this flag set.
 
+*Effective September 15, 2025*, when processing CAA records, CAs MUST process the accounturi and validationmethods parameters as specified in RFC 8657. In addition:
```
geegeea, line 1089: RFC 8657 section 5.6 requires DNSSEC to be validated while performing CAA checks; under the current text, CAs would need to start validating DNSSEC for all CAA DNS queries; is this intentional?

We have seen that in practice many domains that are able to validate under current rules would begin to fail with strict DNSSEC enforcement. GTS would love to see DNSSEC used more widely and configured properly more often, but more time is likely needed before strict DNSSEC checks cause minimal issues for legitimate requesters.

wthayer: Thank you, this is an excellent point. We discussed this today on the Validation SC call and decided that the cleanest solution is to merge this ballot with @clintwilson's proposal to require DNSSEC as drafted at https://github.com/cabforum/servercert/compare/main..Require-DNSSEC-for-CAA-Lookups

docs/BR.md
wthayer: We discussed this on today's Validation SC call and consensus was that we should define the format for the validationmethods label, e.g. 'ca-tlsbr-<method #>' One argument for this was that a standardized format would be more likely to result in automation support.

In addition, @CBonnell suggested that we define the URI format for the accounturi parameter.

I plan to draft changes to this PR to address this feedback.

docs/BR.md
wthayer: Clint's DNSSEC ballot has been incorporated.

wthayer:
Updated based on 24-Jan Validation meeting:
- still specifying the CA-specific label format. consensus was that this does not violate the RFC
- adopted Ben's wording
- rearranged 3.2.2.8 and added subsections
- Changed MUST date to 2027 for parameters. Left the 2026 date for DNSSEC since it's arguably a clarification
- Drafted a recommendation that CAs accept validationmethods labels from ACME or the BRs

shuvosarker22 (approved):

docs/BR.md
```diff
@@ -1236,6 +1219,43 @@ If a Delegated Third Party fulfills any of the CA's obligations under this secti
...
+
+###### 4.2.2.1.2 CAA Parameters
+
+When processing CAA records, CAs SHOULD process the accounturi and validationmethods parameters as specified in RFC 8657. *Effective March 15, 2027*, when processing CAA records, CAs MUST process the accounturi and validationmethods parameters as specified in RFC 8657.
+
+In addition, *Effective March 15, 2026*, if the CA processes the accounturi and validationmethods parameters:
+* If the CA accepts certificate requests via any protocol other than the ACME protocol defined in RFC 8555, the CA MUST define the supported format(s) of the accounturi in Section 4.2 of their CP and/or CPS.
+* If the CA accepts certificate requests via any protocol other than the ACME protocol defined in RFC 8555, the CA MUST interpret and process validationmethods labels formed by concatenating the string ‘ca-tbr-’ with the BR 3.2.2.4 subsection number, e.g. ‘ca-tbr-7’ represents the DNS method described in TLS BR 3.2.2.4.7. If a CA performs domain validation using a mechanism that can be represented by multiple labels (e.g. 'dns-01' and 'ca-tbr-7'), the CA SHOULD accept any of the labels as granting permission to issue.
```
aarongable, line 1253: Echoing my comment from the call this morning: I think this sentence comes a little close to the line of "well, CA Foo doesn't _quite_ implement RFC 8555 (e.g. they don't provide a list of all orders for an account), so they qualify here and need to support these validationmethods labels". I think it's clear that this is not the _intent_ of this sentence, so I'm wondering if there's any minor change we can make to make this clear.

docs/BR.md
```diff
@@ -1236,6 +1219,45 @@ If a Delegated Third Party fulfills any of the CA's obligations under this secti
...
+
+###### 4.2.2.1.2 CAA Parameters
+
+When processing CAA records, CAs SHOULD process the accounturi and validationmethods parameters as specified in RFC 8657. *Effective March 15, 2027*, when processing CAA records, CAs MUST process the accounturi and validationmethods parameters as specified in RFC 8657.
+
+In addition, *Effective March 15, 2026*, if the CA processes the accounturi and validationmethods parameters:
+* If the CA accepts certificate requests via any protocol other than the ACME protocol, the CA MUST define the supported format(s) of the accounturi in Section 4.2 of their CP and/or CPS.
+* If the CA accepts certificate requests via any protocol other than the ACME protocol, the CA MUST interpret and process validationmethods labels formed by concatenating the string ‘ca-tbr-’ with the BR 3.2.2.4 subsection number, e.g. ‘ca-tbr-7’ represents the DNS method described in TLS BR 3.2.2.4.7. If a CA performs domain validation using a mechanism that can be represented by multiple labels (e.g. 'dns-01' and 'ca-tbr-7'), the CA SHOULD accept any of the labels as granting permission to issue.
```
orangepizza, line 1255: 1. Can 'http-01' considered to represent 3.2.2.4.18? while it'd strictly mean 3.2.2.4.19, it's reasonable to assume user meant any http update challenges.
2. Can ACME running CA use 'ca-tbr-19' as permission to use http-01 challenge? this paragraph doesn't apply if a CA doesn't offer something else, isn't it?

aarongable: > 1. Can 'http-01' considered to represent 3.2.2.4.18? while it'd strictly mean 3.2.2.4.19, it's reasonable to assume user meant any http update challenges.

No, 3.2.2.4.18 and 3.2.2.4.19 are in fact different validation methods (the expect the token to be in a different location and to look different) so permission to use one method cannot be interpreted as permission to use the other.

>     2. Can ACME running CA use 'ca-tbr-19' as permission to use http-01 challenge? this paragraph doesn't apply if a CA doesn't offer something else, isn't it?

Yes, this paragraph simply says that non-ACME CAs MUST respect those designations; therefore an ACME CA MAY also respect them, as long as it appropriately documents that fact in its CPS.

docs/BR.md
```diff
@@ -1402,6 +1371,59 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+
+###### 4.2.2.1.3 CAA Parameters
+
+When processing CAA records, CAs SHOULD process the `accounturi` and `validationmethods` parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
+*Effective 2027-03-15*, when processing CAA records, CAs MUST process the `accounturi` and `validationmethod`s parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
+
+In addition, *Effective 2027-03-15*:
+* If the CA accepts certificate requests via any protocol other than the ACME protocol defined in [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555), the CA MUST define the supported format of the `accounturi` in Section 4.2 of their CP and/or CPS.
```
robstradling: A URL must _'provide a means of locating the resource by describing its primary access mechanism (e.g., its network "location")'_ (RFC3986 section 1.1.3).

If there is no unique URL at which an individual Subscriber accesses their account (e.g., all Subscribers access their accounts via a shared login page), then the CA _cannot_ use a URL-based accountURI format.  Constructing a "fake" URL that doesn't actually "work" would be syntactically valid but would fail semantically because it's neither _"a means of locating the resource"_ nor _"its primary access mechanism"_.  In such cases, the CA must use a URN-based accountURI format instead.

Please consider suggesting, recommending, or perhaps even requiring, RFC 7565 (The 'acct' URI Scheme) as the standard format for URN-based accountURIs.

wthayer: Thank you @robstradling, this makes sense to me. I've added "SHOULD comply with the 'acct' URI scheme defined in [RFC 7565](https://datatracker.ietf.org/doc/html/rfc7565)".

docs/BR.md
```diff
@@ -1367,6 +1337,61 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+*Effective 2027-03-15*, when processing CAA records, CAs MUST process the `accounturi` and `validationmethod`s parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
+
+In addition, *Effective 2027-03-15*:
+- If the CA accepts certificate requests via any protocol other than the ACME protocol defined in [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555), the CA MUST define the supported format of the `accounturi` in Section 4.2 of their CP and/or CPS, and SHOULD comply with the 'acct' URI scheme defined in [RFC 7565](https://datatracker.ietf.org/doc/html/rfc7565).
+- For certificate requests made using the ACME protocol, the CA MAY also define an 'accounturi' format following the requirements in the prior bullet. If the CA accepts a self-defined 'accounturi' format for ACME requests, the CA MUST maintain a mapping between the self-defined label and the ACME accounts for which the self-defined label authorizes issuance. Note: this provision overrides RFC 8657 and is intended to allow Subscribers to link many ACME accounts to a single CA account.
```
ChristopherRC: What about something like this?

```suggestion
- For certificate requests made using the ACME protocol, the CA MAY permit the 'accounturi' parameter to identify a primary organizational account (the "Parent Account"). As an explicit exception to Section 3 of RFC 8657, the CA MAY issue a certificate requested by a different account (the "Subordinate ACME Account") if and only if the CA ensures all of the following:
1. The CA maintains an internal, auditable mapping that binds the Subordinate ACME Account to the Parent Account identified by the 'accounturi'.
2. The CA has cryptographically or administratively verified that the Parent Account explicitly authorized the Subordinate ACME Account to obtain certificates under this mapping.
3. The CA retains audit logs demonstrating this authorization and mapping for the standard data retention period required by these Requirements.
```

The intent being:

- Explicitly defining the relationship as a "Parent Account" and a "Subordinate ACME Account" in an attempt to remove ambiguity about what a "self-defined label" is.
- Point 2 attempts to ensure that a CA's automated system cannot arbitrarily map two accounts together. It requires the CA to prove that the Parent Account owner actually authorized the creation of the child account.
- Point 3 attempts to ensure that an auditor reviewing a certificate where the requested ACME account doesn't strictly match the DNS accounturi can ask the CA for the logs to prove the issuance was valid.
- Removing the "Note:" and referencing the specific exception to Section 3 of RFC 8657 makes the deviation more of a normative requirement.

docs/BR.md
```diff
@@ -1367,6 +1337,61 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+
+###### 4.2.2.1.3 CAA Parameters
+
+When processing CAA records, CAs SHOULD process the `accounturi` and `validationmethods` parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
+*Effective 2027-03-15*, when processing CAA records, CAs MUST process the `accounturi` and `validationmethod`s parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
+
+In addition, *Effective 2027-03-15*:
+- If the CA accepts certificate requests via any protocol other than the ACME protocol defined in [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555), the CA MUST define the supported format of the `accounturi` in Section 4.2 of their CP and/or CPS, and SHOULD comply with the 'acct' URI scheme defined in [RFC 7565](https://datatracker.ietf.org/doc/html/rfc7565).
```
geegeea:
```suggestion
- If the CA accepts certificate requests via any protocol other than a protocol functionally equivalent to the ACME protocol defined in [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555), the CA MUST define the supported format of the `accounturi` in Section 4.2 of their CP and/or CPS, and SHOULD comply with the 'acct' URI scheme defined in [RFC 7565](https://datatracker.ietf.org/doc/html/rfc7565).
```

Echoing @aarongable 's older [comment](https://github.com/cabforum/servercert/pull/567#discussion_r1993768921), the suggested change is intended to provide necessary flexibility for CAs whose implementations are based on the ACME protocol but may contain minor deviations from the strict requirements of RFC 8555. For example, a CA may not accept the optional notBefore and notAfter fields of a newOrder request payload. That shouldn't disqualify them from being considered an ACME CA. Similar suggestion on line 1375 below.

docs/BR.md
```diff
@@ -1367,6 +1337,61 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+- For certificate requests made using the ACME protocol, the CA MAY also define an 'accounturi' format following the requirements in the prior bullet. If the CA accepts a self-defined 'accounturi' format for ACME requests, the CA MUST maintain a mapping between the self-defined label and the ACME accounts for which the self-defined label authorizes issuance. Note: this provision overrides RFC 8657 and is intended to allow Subscribers to link many ACME accounts to a single CA account.
+- If the CA accepts certificate requests via any protocol other than the ACME protocol, the CA MUST interpret and process `validationmethods` labels formed by concatenating the string ‘ca-tbr-’ with the BR 3.2.2.4 subsection number, e.g. ‘ca-tbr-7’ represents the DNS method described in TLS BR 3.2.2.4.7. If a CA performs domain validation using a mechanism that can be represented by multiple labels (e.g. 'http-01' and 'ca-tbr-19'), the CA SHOULD accept any of the labels as granting permission to issue.
```
geegeea:
```suggestion
- If the CA supports domain validation methods that are not registered in the [IANA ACME Validation Methods registry](https://www.iana.org/assignments/acme/acme.xhtml#acme-validation-methods), the CA MUST interpret and process `validationmethods` labels formed by concatenating the string ‘ca-tbr-’ with the BR 3.2.2.4 subsection number, e.g. ‘ca-tbr-7’ represents the DNS method described in TLS BR 3.2.2.4.7. If a CA performs domain validation using a mechanism that can be represented by multiple labels (e.g. 'http-01' and 'ca-tbr-19'), the CA SHOULD accept any of the labels as granting permission to issue.
```

Same suggestion as above on line 1373 to avoid a situation where a functionally equivalent ACME-based service is unintentionally forced into the 'any protocol other than the ACME protocol' category.
Edit: Updated suggestion to address @XolphinMartijn 's comment above.

docs/BR.md
XolphinMartijn: @geegeea  I get the thought here, but "functionally equivalent to" is very much open to interpretation. Is a regular REST API functionally equivalent to ACME? Cause the end result can be the same..

Also a minor nit, but:
```suggestion
- If the CA accepts certificate requests via any method other than the ACME protocol defined in [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555), the CA MUST define the supported format of the `accounturi` in Section 4.2 of their CP and/or CPS, and SHOULD comply with the 'acct' URI scheme defined in [RFC 7565](https://datatracker.ietf.org/doc/html/rfc7565).
```

I would suggest the above option, rather than "protocol"

geegeea: I see your point, but the current wording is ambiguous for CAs with minor deviations from the spec. As an example, see the list of Boulder's divergences from ACME: https://github.com/letsencrypt/boulder/blob/main/docs/acme-divergences.md. I would consider Boulder very much an ACME server but the above wording makes it unclear by implying strict RFC 8555 compliance. Since the core of this requirement is how the account is identified, would it be better to say: 'If the CA does not identify the Subscriber account via an ACME Account URL as described in RFC 8555'? This focuses on the URI format rather than full RFC 8555 compliance.
```suggestion
- If the CA does not identify the Subscriber account via an ACME Account URL as described in [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555), the CA MUST define the supported format of the `accounturi` in Section 4.2 of their CP and/or CPS, and SHOULD comply with the 'acct' URI scheme defined in [RFC 7565](https://datatracker.ietf.org/doc/html/rfc7565).
```

docs/BR.md
```diff
@@ -1151,56 +1157,20 @@ Databases maintained by the CA, its owner, or its affiliated companies do not qu
...
 
 #### 3.2.2.9 Multi-Perspective Issuance Corroboration
 
 Multi-Perspective Issuance Corroboration attempts to corroborate the determinations (i.e., domain validation pass/fail, CAA permission/prohibition) made by the Primary Network Perspective from multiple remote Network Perspectives before Certificate issuance. This process can improve protection against equally-specific prefix Border Gateway Protocol (BGP) attacks or hijacks.
 
 The CA MAY use either the same set, or different sets of Network Perspectives when performing Multi-Perspective Issuance Corroboration for the required 1) Domain Authorization or Control and 2) CAA Record checks.
 
-The set of responses from the relied upon Network Perspectives MUST provide the CA with the necessary information to allow it to affirmatively assess:
-
 a. the presence of the expected 1) Random Value, 2) Request Token, 3) IP Address, 4) Contact Address, or 5) Persistent DCV TXT Record, as required by the relied upon validation method specified in [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) and [Section 3.2.2.5](#3225-authentication-for-an-ip-address); and
```
sheurich, line 1175: The reorganization drops the sentence introducing items `a.` and `b.`:

```diff
- The set of responses from the relied upon Network Perspectives MUST provide the CA with the necessary information to allow it to affirmatively assess:
-
  a. the presence of the expected 1) Random Value, ...
  b. the CA's authority to issue to the requested domain(s), ...
```

Without the preamble, items `a.` and `b.` are grammatically incomplete fragments, and the normative MUST governing MPIC assessment is lost.

Suggested fix — restore the preamble above item `a.`:

```suggestion
The set of responses from the relied upon Network Perspectives MUST provide the CA with the necessary information to allow it to affirmatively assess:

a. the presence of the expected 1) Random Value, 2) Request Token, 3) IP Address, 4) Contact Address, or 5) Persistent DCV TXT Record, as required by the relied upon validation method specified in [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) and [Section 3.2.2.5](#3225-authentication-for-an-ip-address); and
```

docs/BR.md
```diff
@@ -1151,56 +1157,20 @@ Databases maintained by the CA, its owner, or its affiliated companies do not qu
...
-b. the CA's authority to issue to the requested domain(s), as specified in [Section 3.2.2.8](#3228-caa-records).
+b. the CA's authority to issue to the requested domain(s), as specified in [Section 4.2.2.1](#4221-caa-records).
```
sheurich: The anchor target `#4221-caa-records` doesn't match the generated slug for the new heading (`#### 4.2.2.1 CAA record processing` → `#4221-caa-record-processing`). The §3.2.2.8 stub above gets this right; this link does not.

```suggestion
b. the CA's authority to issue to the requested domain(s), as specified in [Section 4.2.2.1](#4221-caa-record-processing).
```

docs/BR.md
```diff
@@ -1367,6 +1337,61 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+
+###### 4.2.2.1.3 CAA Parameters
+
+When processing CAA records, CAs SHOULD process the `accounturi` and `validationmethods` parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
+*Effective 2027-03-15*, when processing CAA records, CAs MUST process the `accounturi` and `validationmethod`s parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
```
sheurich: Backtick misplaced — the trailing `s` falls outside the code span, so this renders as `validationmethod`s in body font. The SHOULD line above (line 1369) has it right.

```suggestion
*Effective 2027-03-15*, when processing CAA records, CAs MUST process the `accounturi` and `validationmethods` parameters as specified in [RFC 8657](https://datatracker.ietf.org/doc/html/rfc8657).
```

docs/BR.md
```diff
@@ -1367,6 +1337,61 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+
+###### 4.2.2.1.2 CAA Multi-Perspective Issuance Corroboration
+
+Some methods relied upon for validating the Applicant's ownership or control of the subject domain(s) (see [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control)) or IP address(es) (see [Section 3.2.2.5](#3225-authentication-for-an-ip-address)) to be listed in a certificate require CAA records to be retrieved and processed from additional remote Network Perspectives before Certificate issuance (see [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration)). To corroborate the Primary Network Perspective, a remote Network Perspective's CAA check response MUST be interpreted as permission to issue, regardless of whether the responses from both Perspectives are byte-for-byte identical. Additionally, a CA MAY consider the response from a remote Network Perspective as corroborating if one or both of the Perspectives experience an acceptable CAA record lookup failure, as defined in this section.
```
sheurich: The phrase "as defined in this section" used to unambiguously refer to §3.2.2.8, which contained both this paragraph and the acceptable-lookup-failure definition. After the split, this paragraph lives in §4.2.2.1.2 but the definition is in the parent §4.2.2.1 — an explicit cross-reference would remove the ambiguity.

```suggestion
Some methods relied upon for validating the Applicant's ownership or control of the subject domain(s) (see [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control)) or IP address(es) (see [Section 3.2.2.5](#3225-authentication-for-an-ip-address)) to be listed in a certificate require CAA records to be retrieved and processed from additional remote Network Perspectives before Certificate issuance (see [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration)). To corroborate the Primary Network Perspective, a remote Network Perspective's CAA check response MUST be interpreted as permission to issue, regardless of whether the responses from both Perspectives are byte-for-byte identical. Additionally, a CA MAY consider the response from a remote Network Perspective as corroborating if one or both of the Perspectives experience an acceptable CAA record lookup failure, as defined in [Section 4.2.2.1](#4221-caa-record-processing).
```

docs/BR.md
```diff
@@ -1367,6 +1337,61 @@ CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addres
...
+
+CAs are permitted to treat a record lookup failure as permission to issue if:
+
+- the failure is outside the CA's infrastructure; and
+- the lookup has been retried at least once; and
+- the CA has confirmed that the domain is "Insecure" as defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3).
+
+CAs MUST document potential issuances that were prevented by a CAA record in sufficient detail to provide feedback to the CA/Browser Forum on the circumstances, and SHOULD dispatch reports of such issuance requests to the contact(s) stipulated in the CAA iodef record(s), if present. CAs are not expected to support URL schemes in the iodef record other than mailto: or https:.
+
+###### 4.2.2.1.2 CAA Multi-Perspective Issuance Corroboration
```
sheurich: All three subsections (`4.2.2.1.2`, `4.2.2.1.3`, `4.2.2.1.4`) use `######` (h6), skipping h5. The existing BR convention at this depth is h5 — see `##### 3.2.2.8.1 DNSSEC Validation of CAA Records`. Applying the same fix to lines 1367 and 1378 will restore consistency with the rest of §3.2.2.x.

```suggestion
##### 4.2.2.1.2 CAA Multi-Perspective Issuance Corroboration
```

docs/BR.md
sheurich: +1 on the wording change from "this provision overrides RFC 8657" to an explicit exception to Section 3. The current "overrides" contradicts the "also" earlier in the same bullet and leaves room to read the text as permission to stop accepting the ACME account URL, which I don't believe is the intent. "Explicit exception to Section 3 of RFC 8657" scopes the deviation precisely.

(Narrow endorsement of the framing only — no opinion on the Parent/Subordinate scaffolding.)

docs/BR.md
wthayer: Accepted Gurleen's second suggestion. Thank you!

dzacharo (approved):

PR #573, opened 2025-03-03
SC-XX: Sunset the Inclusion of Address and Routing Parameter Area Names

### LINKED ISSUE #153: Clarify validation requirements for .arpa
In https://github.com/zmap/zlint/issues/343, there were questions raised about whether the special use `in.arpa` and `in6.arpa` domains may make use of wildcards.

BRs section 3.2.2.6 has special provisions for when the wildcard appears to the left of particular domain labels, and it’s ambiguous whether or not that section is triggered, based on how IANA administers that root zone and delegates to the RIRs.

CBonnell:
Resolves #153.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -148,7 +148,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.1.2    | SC80       | Strengthen WHOIS lookups and Sunset Methods 3.2.2.4.2 and 3.2.2.4.15                   | 7-Nov-2024  | 16-Dec-2024                       |
 | 2.1.3    | SC83       | Winter 2024-2025 Cleanup Ballot                                                        | 23-Jan-2025 | 24-Feb-2025                       |
 | 2.1.4    | SC84       | DNS Labeled with ACME Account ID Validation Method                                     | 28-Jan-2025 | 1-Mar-2025                       |
-| 2.1.X    | SCXX       | Sunset inclusion of Address and Routing Parameter Area Names                           | XX-YY-2025 | XX-YY-2025                       |
+| 2.1.X    | SCXX       | Sunset the Inclusion of Address and Routing Parameter Area Names                       | XX-YYY-2025 | XX-YY-2025                       |
 
 \* Effective Date and Additionally Relevant Compliance Date(s)
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1,11 +1,11 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 
-subtitle: Version 2.1.4
+subtitle: Version 2.1.X
 author:
   - CA/Browser Forum
 
-date: 1-March-2025
+date: X-Y-2025
 
 copyright: |
   Copyright 2025 CA/Browser Forum
@@ -148,6 +148,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.1.2    | SC80       | Strengthen WHOIS lookups and Sunset Methods 3.2.2.4.2 and 3.2.2.4.15                   | 7-Nov-2024  | 16-Dec-2024                       |
 | 2.1.3    | SC83       | Winter 2024-2025 Cleanup Ballot                                                        | 23-Jan-2025 | 24-Feb-2025                       |
 | 2.1.4    | SC84       | DNS Labeled with ACME Account ID Validation Method                                     | 28-Jan-2025 | 1-Mar-2025                       |
+| 2.1.X    | SCXX       | Sunset inclusion of Address and Routing Parameter Area Names                           | XX-YY-2025 | XX-YY-2025                       |
 
 \* Effective Date and Additionally Relevant Compliance Date(s)
 
@@ -207,6 +208,8 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2025-03-15     | 8.7                       | The CA SHOULD use a Linting process to test the technical accuracy of already issued Certificates against the sample set chosen for Self-Audits.                                                                                                                                                                                                                                                                                                         |
 | 2025-03-15     | 3.2.2.9                   | CAs MUST corroborate the results of domain validation and CAA checks from multiple Network Perspectives where specified.                                                                                                                                                                                                                                                                                                                                 |
 | 2025-07-15     | 3.2.2.4                   | CAs MUST NOT rely on Methods 3.2.2.4.2 and 3.2.2.4.15 to issue Subscriber Certificates.                                                                                                                                                                                                                                                                                                                                                                  |
+
+| 2025-09-15     | 4.2.2                     | CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names.                                                                                                                                                                                                                                                                                                                                                                         |
 
 ## 1.3 PKI Participants
 
@@ -301,6 +304,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **Application Software Supplier**: A supplier of Internet browser software or other relying-party application software that displays or uses Certificates and incorporates Root Certificates.
 
+**Address and Routing Parameter Area Name**: A Domain Name whose Top-Level Domain is "arpa". Example: `1.1.168.192.in-addr.arpa`.
+
 **Attestation Letter**: A letter attesting that Subject Information is correct written by an accountant, lawyer, government official, or other reliable third party customarily relied upon for such information.
 
 **Audit Period**: In a period-of-time audit, the period between the first day (start) and the last day of operations (end) covered by the auditors in their engagement. (This is not the same as the period of time when the auditors are on-site at the CA.) The coverage rules and maximum length of audit periods are defined in [Section 8.1](#81-frequency-or-circumstances-of-assessment).
@@ -379,7 +384,7 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **High Risk Certificate Request**: A Request that the CA flags for additional scrutiny by reference to internal criteria and databases maintained by the CA, which may include names at higher risk for phishing or other fraudulent usage, names contained in previously rejected certificate requests or revoked Certificates, names listed on the Miller Smiles phishing list or the Google Safe Browsing list, or names that the CA identifies using its own risk-mitigation criteria.
 
-**Internal Name**: A string of characters (not an IP address) in a Common Name or Subject Alternative Name field of a Certificate that cannot be verified as globally unique within the public DNS at the time of certificate issuance because it does not end with a Top Level Domain registered in IANA's Root Zone Database.
+**Internal Name**: A string of characters (not an IP address) in a Common Name or Subject Alternative Name field of a Certificate that cannot be verified as globally unique within the public DNS at the time of certificate issuance because it does not end with a Top-Level Domain registered in IANA's Root Zone Database.
 
 **IP Address**: A 32-bit or 128-bit number assigned to a device that uses the Internet Protocol for communication.
 
@@ -511,6 +516,8 @@ The script outputs:
 **Terms of Use**: Provisions regarding the safekeeping and acceptable uses of a Certificate issued in accordance with these Requirements when the Applicant/Subscriber is an Affiliate of the CA or is the CA.
 
 **Test Certificate**: This term is no longer used in these Baseline Requirements.
+
+**Top-Level Domain**: From RFC 8499 (https://tools.ietf.org/html/rfc8499): "A Top-Level Domain is a zone that is one layer below the root, such as "com" or "jp"."
 
 **Trustworthy System**: Computer hardware, software, and procedures that are: reasonably secure from intrusion and misuse; provide a reasonable level of availability, reliability, and correct operation; are reasonably suited to performing their intended functions; and enforce the applicable security policy.
 
@@ -1246,7 +1253,9 @@ If a Delegated Third Party fulfills any of the CA's obligations under this secti
 
 ### 4.2.2 Approval or rejection of certificate applications
 
-CAs SHALL NOT issue certificates containing Internal Names or Reserved IP Addresses, as such names cannot be validated according to [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) or [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
+CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addresses, as such names cannot be validated according to [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) or [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
+
+Effective 2025-09-15, CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names.
 
 ### 4.2.3 Time to process certificate applications
 
@@ -2749,7 +2758,7 @@ Table: `GeneralName` within a `subjectAltName` extension
 | ---                         | -             | ------         |
 | `otherName`                 | N             | -              |
 | `rfc822Name`                | N             | -              |
-| `dNSName`                   | Y             | The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT contain an Internal Name. The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character. The zero-length Domain Label representing the root zone of the Internet Domain Name System MUST NOT be included (e.g. "example.com" MUST be encoded as "example.com" and MUST NOT be encoded as "example.com."). |
+| `dNSName`                   | Y             | The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT contain an Internal Name. Effective 2025-09-15, the entry MUST NOT contain an Address and Routing Parameter Area Name. The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character. The zero-length Domain Label representing the root zone of the Internet Domain Name System MUST NOT be included (e.g. "example.com" MUST be encoded as "example.com" and MUST NOT be encoded as "example.com."). |
 | `x400Address`               | N             | -              |
 | `directoryName`             | N             | -              |
 | `ediPartyName`              | N             | -              |


docs/BR.md
```diff
@@ -301,6 +304,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **Application Software Supplier**: A supplier of Internet browser software or other relying-party application software that displays or uses Certificates and incorporates Root Certificates.
 
+**Address and Routing Parameter Area Name**: A Domain Name whose Top-Level Domain is "arpa". Example: `1.1.168.192.in-addr.arpa`.
```
dzacharo, line 307: The language is clear but perhaps the example makes the reader believe we are trying to block only non-routable IP addresses. If this is not a big concern by others, you may ignore this comment :)

CBonnell: That's a good point. Maybe we can switch to an IP address from one of these blocks: https://datatracker.ietf.org/doc/html/rfc5737#section-3

docs/BR.md
```diff
@@ -289,6 +292,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 ### 1.6.1 Definitions
 
+**Address and Routing Parameter Area Name**: A Domain Name whose Top-Level Domain is "arpa". Example: `1.2.0.192.in-addr.arpa`.
```
dzacharo, line 295: Shall we also add an example for ip6.arpa?

dzacharo: 6.4.0.0.1.0.0.0.7.0.2.0.5.5.1.0.1.0.0.0.0.0.8.2.8.4.6.0.1.0.0.2.ip6.arpa. :-)

CBonnell: Added :)

kidmin: It would be better to use a prefix for documentation like as IPv4 (`192.0.2.1`). The prefix is `2001:db8::/32` ([RFC 3849](https://datatracker.ietf.org/doc/html/rfc3849)).
Example: `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.d.e.e.f.8.b.d.0.1.0.0.2.ip6.arpa.`

peterthomassen:
I've been following this discussion for a while. It appears to have originated in https://github.com/cabforum/servercert/issues/153 (via https://github.com/zmap/zlint/issues/343) where the problem of wildcard certificates under RIR-controlled rDNS names appeared.

In that thread, it seemed like prohibiting certificates with a wildcard SAN covering an rDNS name would be a good thing to do. It could be easily implemented by looking whether there is a wildcard at any of the labels that is supposed to represent an IPv4 address octet / an IPv6 address nibble.

For some reason, the proposal has shifted to prohibiting certificates under .arpa altogether. I seem to have missed how we went from wildcard to general prohibition!

While the thread linked above only talks explicitly about wildcards, the [pre-ballot discussion](https://groups.google.com/a/groups.cabforum.org/g/servercert-wg/c/LnOC7dpfQqQ) only talks explicitly about the general prohibition.

It would be nice to learn what triggered that shift, especially considering the following points:

1. The [pre-ballot discussion](https://groups.google.com/a/groups.cabforum.org/g/servercert-wg/c/LnOC7dpfQqQ/m/v8WUh32pCwAJ)'s "Purpose of Ballot" says that .arpa "is not intended to include hostnames".
    - This is in contradiction to RFC 3596 which states (in Section 2.5):
      > The intent of this domain is to provide a way of mapping an IPv6 address to a host name, although it may be used for other purposes as well.

      Note that this **both declares those names hostnames**, and **also endorses other purposes**.
    - It is also in contradiction to the fact that there **are** official hostnames under .arpa, such as `blackhole.as112.arpa` (see RFC 7535) or `{a,b,c,...}.ns.arpa` (RFC 9120, as recent as 2021).
    - Even the nameservers of the rDNS subtrees are hostnames under .arpa, such as `a.in-addr-servers.arpa` and `b.ip6-servers.arpa` etc. (see RFC 5855).
    - It's possible that in the future, **those hostnames may need certificates** for DoH/DoT/DoQ.

It's not clear why the (valid) concern for wildcard certificates covering subnet rDNS would need to be addressed in a way that shuts that door unnecessarily.

(Of course, in theory, the rules could be changed again. In practice, people would think "DoX can't be done under .arpa because you can't have a certificate", and drop the idea.)

2. In https://github.com/cabforum/servercert/issues/153#issuecomment-1196011767, @enygren noted
    > that RFC 8738 overloads .arpa as a way to validate IP address certificates, including sending an {in-addr,in6}.arpa name as the SNI value.

    - While that is correct, it would be sufficiently addressed by outlawing certificate issuance for valid IP address rDNS FQDNs.
    - It is certainly not necessary to rule out certificates under other names, such as 37.41.45.in-addr.arpa (which corresponds to a /24 network, not an IP address -- and [actually exists](https://37.41.45.in-addr.arpa/)). Why disallow IP block owners to run an info page there, for example?
    - Other harmless scenarios are possible, e.g., turning an IP address into a hostname like `mail.0.237.26.194.in-addr.arpa` (note the subdomain), which can then be used in an MX record (which requires a hostname) for TLS-secured SMTP without purchasing a domain. Again, why disallow?

These use cases do not collide with the IP address validation SNI ambiguity, and they also have nothing to do with the wildcard problem that this discussion originally started with.

---

All in all, it seems that a general prohibition of certificate issuance under .arpa would be overly broad, and in particular not motivated by any technical or security reasons. Arbitrary carve-outs add needless complexity (e.g., for the nameservers under .arpa), and should be avoided where possible.

The concerns of wildcard certificates and RFC 8738 SNI overloading are addressed by making only the following two changes:
1. CAs SHALL NOT issue Certificates containing names which correspond to the reverse DNS name of an IPv4 address (i.e., exactly 4 labels followed by `in-addr.arpa`) or of an IPv6 address (i.e., exactly 32 labels followed by `ip6.arpa`).
2. CAs SHALL NOT issue Certificates containing names which correspond to the wildcard reverse DNS name of an IPv4 address (i.e., the wildcard label, followed by at most 3 labels and by `in-addr.arpa`) or of an IPv6 address (i.e., the wildcard label followed by at most 31 labels and by `ip6.arpa`).

I'd like to encourage an exchange about whether the proposal could be narrowed down this way, and if it cannot, what the reaons would be.

(Note that (1) breaks sites like https://160.176.40.77.in-addr.arpa/ (which actually exists), and could alternatively be addressed by deprecating validation methods 3.2.2.5.3 and 3.2.2.5.7, or by amending RFC 6066 Section 3 to allow literal IP addresses in the SNI HostName. -- In any case, though, it is not necessary to rule out certificates for subnames of rDNS FQDNs, or for nameservers under .arpa.)

---

Disclosure: I've shared these concerns with the ICANN Security and Stability Advisory Committee of which I am a member, but am not speaking for the committee, nor has the committee developed a consensus position. I am also not speaking for the IETF DNSOP group (of which I am a Secretary).

peterthomassen:
> Hi @peterthomassen, thank you for your insights and feedback.

Thank you for getting back, @CBonnell! :-)

> > For some reason, the proposal has shifted to prohibiting certificates under .arpa altogether. I seem to have missed how we went from wildcard to general prohibition!
> 
> In the course of discussion several years ago, we discussed this particular passage from RFC 3172:
> 
> > This domain is termed an "infrastructure domain", as its role is to support the operating infrastructure of the Internet.  In particular, the "arpa" domain is not to be used in the same manner (e.g., for naming hosts) as other generic Top Level Domains are commonly used.
> 
> and this passage:
> 
> > More generally, this domain name undertakes a role as  a limited use domain for Internet infrastructure applications, by providing a name root for the mapping of particular protocol values to names of service entities.
> 
> Given these statements, it was determined that issuance for any hostname under .arpa would run counter to this guidance, so it was agreed at the time to sunset this issuance.

While this limitation is indeed stated in RFC 3172, that doesn't mean that "you" (CA/B Forum) has to enforce it :-)

That's particularly true if a certain "lapse" causes no issues, and may actually be useful to some. In fact, not even the IETF sees itself as the "protocol police", except on specific days (see the publication date of RFC 8962). :) (... true, RFC 3172 is from the IAB, not the IETF.)

> > The intent of this domain is to provide a way of mapping an IPv6 address to a host name, although it may be used for other purposes as well.
> 
> This section is speaking to PTR records, not to naming hosts under ip6.arpa. As for "other purposes", naming hosts would conflict with the guidance in RFC 3172.

Fair.

> I discussed your concerns with the endorsers of the ballot and we will not move forward with this ballot as-is, as there is value in allowing issuance for DoH/DoT nameservers under .arpa and do not want to hamper those efforts. Our current thinking is that ballot will be scaled down to prohibiting wildcards under .arpa, but we can certainly discuss.

Great! Thank you for considering the feedback with the others.

> As I hinted at above, I think it would very helpful that RFC 3172 be updated to explicitly permit the naming of hosts such as the previously mentioned nameservers. Doing so would eliminate confusion on the intended purpose of .arpa.

I agree with you, yes!

docs/BR.md
```diff
@@ -208,6 +209,8 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
...
 | 2025-07-15     | 3.2.2.4                   | CAs MUST NOT rely on Methods 3.2.2.4.2 and 3.2.2.4.15 to issue Subscriber Certificates.                                                                                                                                                                                                                                                                                                                                                                  |
 
+| 2025-09-15     | 4.2.2                     | CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names.                                                                                                                                                                                                                                                                                                                                                                         |
```
dzacharo:
```suggestion
| 2026-03-15     | 4.2.2                     | CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names.                                                                                                                                                                                                                                                                                                                                                                         |
```

Suggest future effective date

docs/BR.md
```diff
@@ -1246,7 +1253,9 @@ If a Delegated Third Party fulfills any of the CA's obligations under this secti
 
 ### 4.2.2 Approval or rejection of certificate applications
 
-CAs SHALL NOT issue certificates containing Internal Names or Reserved IP Addresses, as such names cannot be validated according to [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) or [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
+CAs SHALL NOT issue Certificates containing Internal Names or Reserved IP Addresses, as such names cannot be validated according to [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) or [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
+
+Effective 2025-09-15, CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names.
```
dzacharo:
```suggestion
Effective 2026-03-15, CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names.
```

docs/BR.md
```diff
@@ -289,6 +292,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 ### 1.6.1 Definitions
 
+**Address and Routing Parameter Area Name**: A Domain Name whose Top-Level Domain is "arpa". Examples: `1.2.0.192.in-addr.arpa` (IP version 4) and `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa` (IP version 6).
```
dzacharo:
```suggestion
**Address and Routing Parameter Area Name**: A Domain Name whose Top-Level Domain ends with the labels "in-addr.arpa" or "ip6.arpa". Examples: `1.2.0.192.in-addr.arpa` (IP version 4) and `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa` (IP version 6).
```

docs/BR.md
```diff
@@ -2749,7 +2758,7 @@ Table: `GeneralName` within a `subjectAltName` extension
...
+| `dNSName`                   | Y             | The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT contain an Internal Name. Effective 2025-09-15, the entry MUST NOT contain an Address and Routing Parameter Area Name. The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character. The zero-length Domain Label representing the root zone of the Internet Domain Name System MUST NOT be included (e.g. "example.com" MUST be encoded as "example.com" and MUST NOT be encoded as "example.com."). |
```
dzacharo:
```suggestion
| `dNSName`                   | Y             | The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT contain an Internal Name. Effective 2026-03-15, the entry MUST NOT contain an Address and Routing Parameter Area Name. The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character. The zero-length Domain Label representing the root zone of the Internet Domain Name System MUST NOT be included (e.g. "example.com" MUST be encoded as "example.com" and MUST NOT be encoded as "example.com."). |
```

dzacharo (commented):
Updates based on SCWG Teleconference of 2025-08-28.

@CBonnell please review and share with external reviewers as needed.

docs/BR.md
peterthomassen:
```suggestion
**Address and Routing Parameter Area Name**: A Domain Name which ends with the labels "in-addr.arpa" or "ip6.arpa". Examples: `1.2.0.192.in-addr.arpa` (IP version 4) and `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa` (IP version 6).
```

CBonnell: We should probably rename the definition since it's now scoped to only `in-addr.arpa` and `ip6.arpa`. I'll dig around a few RFCs to see if there's a term we can (re)use to describe a Domain Name that ends in `in-addr.arpa` or `ip6.arpa`.

peterthomassen: Ah, I now realize that the new text is limited to the reverse DNS second-level names, and no longer encompasses all of the `.arpa` TLD. That seems like a reasonable resolution of the problem! Thank you for making this change (and sorry for not realizing this when I sent my recent message on the list).

It's a bit of a pity that sites like https://37.41.45.in-addr.arpa/ won't work any longer, especially as they are currently working and will suddenly break. But I understand the position that this is not viewed as a legitimate use of the namespace by everyone.

> We should probably rename the definition since it's now scoped to only `in-addr.arpa` and `ip6.arpa`. I'll dig around a few RFCs to see if there's a term we can (re)use to describe a Domain Name that ends in `in-addr.arpa` or `ip6.arpa`.

RFC 3172 Section 4 talks about `in-addr.arpa` as "[T]he IPv4 reverse address domain". So perhaps the definition could just be called "Reverse Address Domains" (and associated changes, like "CAs SHALL NOT issue Certificates containing Address and Routing Parameter Area Names" --> "CAs SHALL NOT issue Certificates under Reverse Address Domains").

docs/BR.md
```diff
@@ -305,6 +307,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 ### 1.6.1 Definitions
 
+**Address and Routing Parameter Area Name**: A Domain Name whose Top-Level Domain ends with the labels "in-addr.arpa" or "ip6.arpa". Examples: `1.2.0.192.in-addr.arpa` (IP version 4) and `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa` (IP version 6).
```
peterthomassen: While the definition correctly captures the scope for reverse DNS names, the name "Address and Routing Parameter Area Name" sounds like it's intended to apply to all of `.arpa`. There's a risk of confusion.

Suggest renaming this definition to "IP reverse address domains".

(The change would affect other places in this PR as well.)

clintwilson (approved):
This looks clean and straight-forward to me.

docs/BR.md
```diff
@@ -219,6 +219,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
...
 | 2026-03-15     | 3.2.2.8.1                 | DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.                                                                                                                                                                                                                                                                                                                        |
+| 2026-03-15     | 4.2.2                     | CAs SHALL NOT issue Certificates containing IP Reverse Address Domain Names.                                                                                                                                                                                                                                                                                                                                                                         |
```
geegeea:
```suggestion
| 2026-03-15     | 4.2.2                     | CAs SHALL NOT issue Certificates containing Domain Names ending in an IP Reverse Zone Suffix.                                                                                                                                                                                                                                                                                                                                                                         |
```

docs/BR.md
```diff
@@ -2863,7 +2870,7 @@ Table: `GeneralName` within a `subjectAltName` extension
...
+| `dNSName`                   | Y             | The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT contain an Internal Name. Effective 2026-03-15, the entry MUST NOT contain an IP Reverse Address Domain Name. The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character. The zero-length Domain Label representing the root zone of the Internet Domain Name System MUST NOT be included (e.g. "example.com" MUST be encoded as "example.com" and MUST NOT be encoded as "example.com."). |
```
geegeea:
```suggestion
| `dNSName`                   | Y             | The entry MUST contain either a Fully-Qualified Domain Name or Wildcard Domain Name that the CA has validated in accordance with [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). Wildcard Domain Names MUST be validated for consistency with [Section 3.2.2.6](#3226-wildcard-domain-validation). The entry MUST NOT contain an Internal Name. Effective 2026-03-15, the entry MUST NOT contain a Domain Name ending in an IP Reverse Zone Suffix. The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character. The zero-length Domain Label representing the root zone of the Internet Domain Name System MUST NOT be included (e.g. "example.com" MUST be encoded as "example.com" and MUST NOT be encoded as "example.com."). |
```

docs/BR.md
```diff
@@ -404,7 +404,7 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **IP Address Registration Authority**: The Internet Assigned Numbers Authority (IANA) or a Regional Internet Registry (RIPE, APNIC, ARIN, AfriNIC, LACNIC).
 
-**IP Reverse Address Domain Name**: A Domain Name that ends with the Domain Labels "in-addr.arpa" or "ip6.arpa". Examples: `1.2.0.192.in-addr.arpa` (IP version 4) and `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa` (IP version 6).
+**IP Reverse Zone Suffix**: One of the two FQDNs that consist of the Domain Labels "in-addr.arpa" or "ip6.arpa". These two FQDNs serve as the root of the IP version 4 and IP version 6 reverse mapping space. "in-addr.arpa" is the root of the IP version 4 reverse mapping space and "ip6.arpa" is the root of the IP version 6 reverse mapping space.
```
tobij, line 410: I am not sure we really have to explain which is which; it does not seem relevant to begin with in this definition and the BRs are not Wikipedia, but maybe I would have formed a different opinion if I looked at this from the perspective of SC-91 instead?

In any case, I do not think it is a problem to leave it in, but it stood out, so I thought I would mention it.

clintwilson (approved):

PR #579, opened 2025-03-26
SC-085: Require Validation of DNSSEC (when present) for CAA and DCV Lookups

birgelee:
(updating PR #571 ) Require DNSSEC for CAA and DCV lookups using an RFC reference. Obsoletes PR #571 and branches off of the latest commits on main instead of Clint's branch since Clint's branch (as Clint's old DNSSEC branch is now several commits behind).

Identical text as PR #571 except i.e., statement was removed per discussion to avoid confusion.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -741,13 +741,14 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
-Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective MUST:
 
-* MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) with the exception that the resolver does not need to distinguish between the Bogus and Indeterminate states defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3); and
-* MUST support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
-* MUST support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
-* MUST properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4)
-* SHOULD have the Extension Mechanisms for DNS (EDNS) buffer size set to 1232 bytes.
+* be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) with the exception that the resolver does not need to distinguish between the Bogus and Indeterminate states defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3); and
+* support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
+* support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
+* properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4).
+
+In addition, the DNS resolver SHOULD have the Extension Mechanisms for DNS (EDNS) buffer size set to 1232 bytes.
 
 Effective November 15th, 2025: CAs MUST NOT use local policy to disable DNSSEC validation on any DNS query associated with the validation of domain authorization or control.
 
@@ -1133,13 +1134,15 @@ CAs MUST document potential issuances that were prevented by a CAA record in suf
 
 ##### 3.2.2.8.1 DNSSEC Validation of CAA Records
 
-Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective:
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective MUST:
 
-* MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) with the exception that the resolver does not need to distinguish between the Bogus and Indeterminate states defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3); and
-* MUST support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
-* MUST support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
-* MUST properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4)
-* SHOULD have the Extension Mechanisms for DNS (EDNS) buffer size set to 1232 bytes.
+* be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) with the exception that the resolver does not need to distinguish between the Bogus and Indeterminate states defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3); and
+* support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
+* support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
+* properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4).
+
+
+In addition, the DNS resolver SHOULD have the Extension Mechanisms for DNS (EDNS) buffer size set to 1232 bytes.
 
 Effective November 15th, 2025: CAs MUST NOT use local policy to disable DNSSEC validation on any DNS query associated CAA record lookups.
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -743,7 +743,7 @@ Completed validations of Applicant authority may be valid for the issuance of mu
 
 Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective MUST:
 
-* validate DNSSEC using the algorithm defined in [RFC 4035 Section 5](https://datatracker.ietf.org/doc/html/rfc4035#section-5); and
+* perform DNSSEC validation using the algorithm defined in [RFC 4035 Section 5](https://datatracker.ietf.org/doc/html/rfc4035#section-5); and
 * support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
 * support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
 * properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4).
@@ -1134,7 +1134,7 @@ CAs MUST document potential issuances that were prevented by a CAA record in suf
 
 Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective MUST:
 
-* validate DNSSEC using the algorithm defined in [RFC 4035 Section 5](https://datatracker.ietf.org/doc/html/rfc4035#section-5); and
+* perform DNSSEC validation using the algorithm defined in [RFC 4035 Section 5](https://datatracker.ietf.org/doc/html/rfc4035#section-5); and
 * support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
 * support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
 * properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4).
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1145,6 +1145,8 @@ Effective March 15th, 2026: DNSSEC-validation errors observed by the Primary Net
 
 DNSSEC validation back to the IANA DNSSEC root trust anchor MAY be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration.
 
+DNSSEC validation back to the IANA DNSSEC root trust anchor is considered outside the scope of self-audits performed to fulfill the requirements in [Section 8.7](#87-self-audits).
+
 #### 3.2.2.9 Multi-Perspective Issuance Corroboration
 
 Multi-Perspective Issuance Corroboration attempts to corroborate the determinations (i.e., domain validation pass/fail, CAA permission/prohibition) made by the Primary Network Perspective from multiple remote Network Perspectives before Certificate issuance. This process can improve protection against equally-specific prefix Border Gateway Protocol (BGP) attacks or hijacks.
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -731,6 +731,10 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) and MUST be compliant with [RFC 6840](https://datatracker.ietf.org/doc/html/rfc6840).
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor SHOULD be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration, but it is not required.
+
 CAs SHALL maintain a record of which domain validation method, including relevant BR version number, they used to validate every domain.
 
 **Note**: FQDNs may be listed in Subscriber Certificates using `dNSName`s in the `subjectAltName` extension or in Subordinate CA Certificates via `dNSName`s in `permittedSubtrees` within the Name Constraints extension.
@@ -1105,9 +1109,17 @@ CAs are permitted to treat a record lookup failure as permission to issue if:
 
 * the failure is outside the CA's infrastructure; and
 * the lookup has been retried at least once; and
-* the domain's zone does not have a DNSSEC validation chain to the ICANN root.
+* the CA has confirmed that the domain is "Insecure" as defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3).
 
 CAs MUST document potential issuances that were prevented by a CAA record in sufficient detail to provide feedback to the CA/Browser Forum on the circumstances, and SHOULD dispatch reports of such issuance requests to the contact(s) stipulated in the CAA iodef record(s), if present. CAs are not expected to support URL schemes in the iodef record other than mailto: or https:.
+
+##### 3.2.2.8.1 DNSSEC Validation of CAA Records
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) and MUST be compliant with [RFC 6840](https://datatracker.ietf.org/doc/html/rfc6840).
+
+DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor SHOULD be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration, but it is not required.
 
 #### 3.2.2.9 Multi-Perspective Issuance Corroboration
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -735,13 +735,13 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
-DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
+Effective November 15th, 2025: DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
 
 * MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
 * MUST support NSEC3 and SHA-2 as defined in [RFC 6840 Section 2](https://datatracker.ietf.org/doc/html/rfc6840#section-2); and
 * MUST properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4)
 
-DNSSEC validation back to the ICANN DNSSEC root trust anchor SHOULD be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration, but it is not required.
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MAY be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration.
 
 CAs SHALL maintain a record of which domain validation method, including relevant BR version number, they used to validate every domain.
 
@@ -1123,13 +1123,13 @@ CAs MUST document potential issuances that were prevented by a CAA record in suf
 
 ##### 3.2.2.8.1 DNSSEC Validation of CAA Records
 
-DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective:
+Effective November 15th, 2025: DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective:
 
 * MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
 * MUST support NSEC3 and SHA-2 as defined in [RFC 6840 Section 2](https://datatracker.ietf.org/doc/html/rfc6840#section-2); and
 * MUST properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4)
 
-DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
+Effective November 15th, 2025: DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
 
 DNSSEC validation back to the ICANN DNSSEC root trust anchor MAY be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration.
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1119,7 +1119,7 @@ DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed o
 
 DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
 
-DNSSEC validation back to the ICANN DNSSEC root trust anchor SHOULD be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration, but it is not required.
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MAY be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration.
 
 #### 3.2.2.9 Multi-Perspective Issuance Corroboration
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -735,13 +735,13 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
-Effective November 15th, 2025: DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
 
 * MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
 * MUST support NSEC3 and SHA-2 as defined in [RFC 6840 Section 2](https://datatracker.ietf.org/doc/html/rfc6840#section-2); and
 * MUST properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4)
 
-DNSSEC validation back to the ICANN DNSSEC root trust anchor MAY be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration.
+DNSSEC validation back to the IANA DNSSEC root trust anchor MAY be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration.
 
 CAs SHALL maintain a record of which domain validation method, including relevant BR version number, they used to validate every domain.
 
@@ -1123,7 +1123,7 @@ CAs MUST document potential issuances that were prevented by a CAA record in suf
 
 ##### 3.2.2.8.1 DNSSEC Validation of CAA Records
 
-Effective November 15th, 2025: DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective:
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective:
 
 * MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
 * MUST support NSEC3 and SHA-2 as defined in [RFC 6840 Section 2](https://datatracker.ietf.org/doc/html/rfc6840#section-2); and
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -1105,10 +1109,18 @@ CAs are permitted to treat a record lookup failure as permission to issue if:
...
 
+##### 3.2.2.8.1 DNSSEC Validation of CAA Records
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) and MUST be compliant with [RFC 6840](https://datatracker.ietf.org/doc/html/rfc6840).
+
+DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor SHOULD be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration, but it is not required.
```
dzacharo, line 1122: I think it would be better to change the SHOULD, combined with "not required" in the same sentence, to MAY.

```suggestion
DNSSEC validation back to the ICANN DNSSEC root trust anchor MAY be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration.
```

dzacharo:
Please don't forget to add the referenced RFCs in section 1.6.3.

docs/BR.md
ChristopherRC: Is it better to keep the SHOULD (and its RFC 2119 meaning) as is, and instead remove the trailing “[...], but it is not required.” at the end of the sentence? It seems that the proposed change of the existing SHOULD to a MAY would make this practice not recommended and instead optional.

dzacharo:
It should be optional. It doesn't provide any meaningful security improvement if the DNS zone is already signed. This was confirmed in past discussions and repeated by Henry in today's validation subcommittee meeting.

DZ. 
 
  On Thu, Mar 27, 2025 at 10:18, Chris ***@***.***> wrote:   



@ChristopherRC commented on this pull request.

In docs/BR.md:
>  
 CAs MUST document potential issuances that were prevented by a CAA record in sufficient detail to provide feedback to the CA/Browser Forum on the circumstances, and SHOULD dispatch reports of such issuance requests to the contact(s) stipulated in the CAA iodef record(s), if present. CAs are not expected to support URL schemes in the iodef record other than mailto: or https:.
 
+##### 3.2.2.8.1 DNSSEC Validation of CAA Records
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with CAA record lookups performed by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with CAA record lookups performed by the Primary Network Perspective MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) and MUST be compliant with [RFC 6840](https://datatracker.ietf.org/doc/html/rfc6840).
+
+DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor SHOULD be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration, but it is not required.

Is it better to keep the SHOULD (and its RFC 2119 meaning) as is, and instead remove the trailing “[...], but it is not required.” at the end of the sentence? It seems that the proposed change of the existing SHOULD to a MAY would make this practice not recommended and instead optional.

docs/BR.md
dzacharo: It should be optional. It doesn't provide any meaningful security improvement if the DNS zone is already signed. This was confirmed in past discussions and repeated by Henry in today's validation subcommittee meeting.

birgelee: After thinking over this some more I am slightly in favor of "MAY". I know some auditors take "SHOULD" very seriously and having this as a SHOULD would make not validating DNSSEC at remote perspectives less than optimal.

The advantage of "MAY" in this sentence is that MPIC does best when perspectives are placed in the most diverse locations possible including spread across different cloud providers (this is not required or typically something I have seen in existing MPIC deployments, but it can offer further improved security). If a CA were considering deploying in a new cloud provider that did not support DNSSEC, I argue it would be better for the CA to deploy there and expand to a different cloud provider than to stay with a more conservative approach of only staying in cloud providers that support DNSSEC.

I think with MPIC requirements in general, language has been chosen to encourage ease of deployment in diverse regions over strict security mandates for those regions (which is sometimes at odds of diversity). This, I feel here going against the "more is better" attitude of security to favor MPIC deployment diversity is preferable.

That said, for those who can deploy DNSSEC validation in your MPIC infrastructure, I do encourage it. Open MPIC even provides a sample unbound container (which we use in our docker-based deployment examples) to assist with DNSSEC validation in MPIC deployments ( https://github.com/open-mpic/open-mpic-containers/tree/main/unbound ). On a side note: Open MPIC does test against the CAA test suite and a non-DNSSEC MPIC deployment will fail *.caatestsuite-dnssec.com tests.

birgelee:
My recent commits added a reference to the appropriate RFCs in Sec 1.6.3

I also did a full review of RFC 6840. Several Forum members advised either moving the RFC 6840 reference to a SHOULD or bringing in specific sections. Upon reviewing the RFC I feel the only sections that need to be referenced in the Baseline Requirements are Sections 2 and Sections 4 and it would be best to keep them as a MUST statement.

Section 2 defines the critical additional algorithms NSEC3 and SHA-2. To quote RFC 6840:

Validators that do not support
   validation of responses using NSEC3 will be hampered in validating
   large portions of the DNS space.

I personally feel SHA-2 is in a similar boat given it has become a very popular digest algorithm and not supporting it significantly hampers the effectiveness of DNSSEC validation.

RFC 6840 Section 4 describes several security concerns. Some of these are merely clarifications of previous RFC text. As RFC 6840 states:

This section provides clarifications that, if overlooked, could lead
   to security issues.

I similarly think these are important to include as a MUST statement.

The other text in the RFC discusses other concerns like interoperability and are not critical to the secure operations of DNSSEC.

If someone is aware of a modern DNS resolver that does not comply with sections 2 and 4 of RFC6840 please let me know. This RFC is now 12 years old and largely presents clarifications on the proper way to operate DNSSEC. I has been included in every RFC compliance list I have found.

docs/BR.md
```diff
@@ -731,6 +735,14 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
...
 
+Effective November 15th, 2025: DNSSEC validation back to the ICANN DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
+
+* MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
+* MUST support NSEC3 and SHA-2 as defined in [RFC 6840 Section 2](https://datatracker.ietf.org/doc/html/rfc6840#section-2); and
+* MUST properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4)
+
+DNSSEC validation back to the ICANN DNSSEC root trust anchor MAY be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration.
```
CBonnell, line 744: I received a comment that IANA maintains the DNSSEC root trust anchor, not ICANN.

birgelee: IANA is more appropriate here. I will update this. The old text in the BRs referencing the DNSSEC root trust anchor said ICANN, but IANA is actually tasked with managing it.

SherfeyInv (approved):

CoutoKona822 (approved):

docs/BR.md
```diff
@@ -1105,10 +1117,22 @@ CAs are permitted to treat a record lookup failure as permission to issue if:
 
 * the failure is outside the CA's infrastructure; and
 * the lookup has been retried at least once; and
-* the domain's zone does not have a DNSSEC validation chain to the ICANN root.
+* the CA has confirmed that the domain is "Insecure" as defined in [RFC 4035 Section 4.3](https://datatracker.ietf.org/doc/html/rfc4035#section-4.3).
```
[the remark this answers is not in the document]
birgelee: Thanks for bringing this up. I did review the thread and I think the main issue was unbound-host marking several domains as "insecure" where in fact it had lookup errors so the insecure status of the domain could not be determined.

It appears this issue has been fixed an updated version of unbound-host. Rerunning the queries in question I now get a status of "2(SERVFAIL). (error)" which no longer states the domain is insecure, thus complying with the RFC. This ballot language also has an explicit reference to the "SERVFAIL" error (see line 1132 of the new changes) as not being permission to issue which I think is helpful.

Even in light of this, I still think the use of RFC language is superior to prior language in the BRs.

1. The RFC 4035 definition of "Insecure" properly defines the state of a domain which is appropriate for this CAA check failure language. Given there is an existing and appropriate definition in a well established DNSSEC standard doc, I think the correct thing is to use it.
2. I think these problems and bugs discussed exist whether the old or new language is employed. The linked MDSP thread is in fact debating the applicability of signing certificates in these states under the old baseline requirement language (which is in my opinion a less accurate and more ambiguous description of the RFC 4035 insecure state). The root of the problem is that if a DNS resolver encounters certain types of lookup errors, it can be very difficult to determine the DNSSEC state of a domain. I think this challenge is helped by specifically tying the requirement to the RFC 4035 insecure state which is understood by DNS software. Without this I personally don't see a straightforward way for a CA to deduce the presence of absence of a DNSSEC chain given the CA's DNS resolver is throwing an error on the domain in question.
3. As I stated previously, the mentioned bug has since been fixed.

A longer-term initiative along this front could be to conduct a survey of CAs and find out how many still issue on CAA lookup failures. The cleanest approach down the road would probably be to remove this language (i.e., require all CAA lookups to be successful) because this clause requires DNS resolvers that return errors to additionally indicate if the domain in question was DNSSEC signed or not. If the DNS resolver had an error, then something went wrong, and it is likely the DNS resolver will not be able to make an accurate attestation of the domain's DNSSEC status.

In the interest of minimal changes in each ballot, I felt proposing removal of this language would simply complicate things and is not fully related to DNSSEC. However, given that we now have an RFC 4035 reference, I feel using it here is appropriate and offers more clarity.

Below are the outputs of my unbound-host runs using the same flags as the original MDSP post:
```
henry@MacBook-Pro-259 sbin % ./unbound-host -v -t CAA -D -f /usr/local/etc/unbound/root.key blackhole.caatestsuite-dnssec.com.
Host blackhole.caatestsuite-dnssec.com. not found: 2(SERVFAIL). (error)

henry@MacBook-Pro-259 sbin % ./unbound-host -v -t CAA -D -f /usr/local/etc/unbound/root.key servfail.caatestsuite-dnssec.com. 
Host servfail.caatestsuite-dnssec.com. not found: 2(SERVFAIL). (error)

henry@MacBook-Pro-259 sbin % ./unbound-host -v -t CAA -D -f /usr/local/etc/unbound/root.key refused.caatestsuite-dnssec.com. 
Host refused.caatestsuite-dnssec.com. not found: 2(SERVFAIL). (error)

henry@MacBook-Pro-259 sbin % ./unbound-host -v -t CAA -D -f /usr/local/etc/unbound/root.key expired.caatestsuite-dnssec.com. 
expired.caatestsuite-dnssec.com. has no CAA record (BOGUS (security failure))
validation failure <expired.caatestsuite-dnssec.com. CAA IN>: signature expired from 45.79.179.40 and 45.79.179.40 for key expired.caatestsuite-dnssec.com. while building chain of trust

henry@MacBook-Pro-259 sbin % ./unbound-host -v -t CAA -D -f /usr/local/etc/unbound/root.key missing.caatestsuite-dnssec.com. 
missing.caatestsuite-dnssec.com. has no CAA record (BOGUS (security failure))
validation failure <missing.caatestsuite-dnssec.com. CAA IN>: no signatures from 45.79.179.40 for key missing.caatestsuite-dnssec.com. while building chain of trust

```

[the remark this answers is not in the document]
CBonnell: Thanks for checking this. What version of unbound-host are you using?

docs/BR.md
```diff
@@ -731,6 +735,14 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
+
+* MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
+* MUST support NSEC3 and SHA-2 as defined in [RFC 6840 Section 2](https://datatracker.ietf.org/doc/html/rfc6840#section-2); and
```
CBonnell, line 741: I don't see any MUST-level requirements in https://datatracker.ietf.org/doc/html/rfc6840#section-2.1 (NSEC3) or https://datatracker.ietf.org/doc/html/rfc6840#section-2.2 (SHA-2) for validators to implement these two features. They are "strongly encouraged", which I interpret to be a "SHOULD" as opposed to being mandatory.

docs/BR.md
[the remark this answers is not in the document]
birgelee: Unbound and unbound-host are Version 1.22.0

docs/BR.md
```diff
@@ -731,6 +735,14 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective:
+
+* MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4); and
```
CBonnell, line 740: A few thoughts:

1. This ballot should explicitly disallow CAs from using the "get out of jail free card" where DNSSEC validation is disabled if defined by local policy (RFC 4035, section 4.2)
2. I think the fragmentation guidance in section 4.1 regarding proper handling of fragmentation is outdated. See https://www.dnsflagday.net/2020/. Specifically, the SHOULD in that section is outdated.
3. The requirement to discern the four different states is not strictly required. Specifically, there is no need to distinguish between Bogus and Indeterminate for the purposes of domain validation.
4. Add a reference to RFC 8198, section 7? https://www.rfc-editor.org/rfc/rfc8198.html#section-7

docs/BR.md
birgelee: In this situation, I think it is important to consider context. Particularly 1) RFC 6840 is now 13 years old and 2) the RFC is titled "Clarifications and Implementation Notes for DNS Security (DNSSEC)." Given this RFC is intended to be clarification and security notes, I suspect the IETF would have deemed it inappropriate to use MUST to dictate the use of new algorithms. Furthermore, this type of language would also instantaneously made any existing DNSSEC implementation that had not yet implemented those algorithms non-compliant which I do not think was the goal of this RFC. In today's DNSSEC ecosystem I feel a MUST is appropriate for both of these. I will address the case for SHA-2 and NSEC3 separately.

SHA-2: IANA lists the SHA-2 algorithm as mandatory ( https://www.iana.org/assignments/ds-rr-types/ds-rr-types.xhtml ). I feel SHA-2 being mandatory is particularly critical as the behavior of a DNSSEC-validating resolver when it encounters an unknown digest algorithm is to continue processing the lookup without performing validation. The delegations to .com, .net and .org all use the SHA-2 digest so any domain under these TLDs cannot be validated if a resolver does not support the SHA-2 digest. Measuring the top DNSSEC domains on the Tranco list ( https://tranco-list.eu ) shows 960 out of the 1058 DS records found use the SHA-2 algorithm ( https://github.com/birgelee/dnssec-measurements ). Furthermore, there exist no viable alternatives to the SHA-2 digest. The SHA-1 digest originally introduced in the first DNSSEC documents is now broken and all other digests are optional with limited support. Thus, if SHA-2 is not supported there is no option for a secure digest algorithm. Given that unknown digest algorithms cause a fail-open behavior, not supporting SHA-2 is similar to not validating at all. I personally cannot see mandating DNSSEC without requiring support of the SHA-2 digest.

NSEC3: Unlike SHA-2 support, resolvers lacking NSEC3 support will fail closed in the presence of NSEC3 records. If a recursive resolver queries a record that does not exist and cannot understand an NSEC3 record, it will be unable to validate the proof of nonexistence and return an error code for the query. Technically this is an availability not a security concern. However, NSEC3 is quite prominent with 296 / 707 top domains using it ( https://github.com/birgelee/dnssec-measurements ). A CA not supporting NSEC3 will likely encounter significant errors particularly since CAA record behavior treats the absence of the record as permission to issue. This could harm DNSSEC deployment as many properly configured domains will encounter DNSSEC errors which are not the domain's fault but due to the CA's use of an outdated DNSSEC validator.

It is important to remember that RFC 6840 is 13 years old. 13 years ago, it may not have seemed fitting to strictly mandate the use of these new algorithms and it reasonable that the algorithms be recommended for a significant period of time before being mandated.  RFC 6840 does still contain strong language as it "define[s] NSEC3 and SHA-2 (RFC 4509 and RFC 5702) as core parts of the DNSSEC specification." (RFC 6840). Given that 1) NSEC3 and SHA-2 are core parts of the DNSSEC specification and 2) these algorithms have been deployed significantly leading to high prevalence and even dominance (in the case of SHA-2) in the DNSSEC ecosystem, I think mandating their support is called for. Additionally, I would ask the question of what operational advantage is obtained by listing these as SHOULD? Are there any contemporary production DNS resolvers that do not support these algorithms? I would even argue there is an operational advantage to having these as a MUST given a resolver that does not support these methods is likely to be 13+ years out of date and subject to ample other security vulnerabilities.

docs/BR.md
birgelee: ### regarding local policy
 I agree. Possibly language like the following could be added:

CAs MUST NOT use local policy to disable DNSSEC validation on any DNS query associated with the validation of domain authorization or control.

with a corresponding line in the CAA section.

### regarding fragmentation guidance
That is correct. The SHOULD in that section is outdated and the DNS Flag Day recommendation is the current best practice (e.g., https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html ). Perhaps just adding a clarifying SHOULD in the bullet list

* SHOULD have the Extension Mechanisms for DNS (EDNS) buffer size set to 1232 bytes.

### regarding distinguishing between Bogus and Indeterminate

I agree that distinguishing between Bogus and Indeterminate is not required. I feel both should be treated as a lookup failure that does not give permission to issue. Maybe the first bullet could be changed to:

* MUST be a "security-aware resolver" as defined in [RFC 4035 Section 4](https://datatracker.ietf.org/doc/html/rfc4035#section-4) with the exception that the resolver does not need to distinguish between the Bogus and Indeterminate sates; and

### RFC 8198, section 7

Sure. Potentially in the DNS resolver bullet list stating the properties, a property could be added stating:

* SHOULD follow the guidance in RFC 8198 Section 7.

I will note that I don't think this update is strictly needed from a security or compliance standpoint.

docs/BR.md
birgelee: I was reviewing RFCs and I came across [RFC 4034 Appendix A](https://datatracker.ietf.org/doc/html/rfc4034#appendix-A) which reads:

A DNSSEC aware resolver or name server MUST implement all MANDATORY
   algorithms.

RFC 4034 is referenced by RFC 4035 as being the authoritative reference for all DNSSEC resource records. Given that the [current IANA registry](https://www.iana.org/assignments/ds-rr-types/ds-rr-types.xhtml) lists SHA-2 as MANDATORY and 4034 requires support for all MANDATORY algorithms, I feel not supporting SHA-2 would be an RFC violation.

CBonnell: To be clear, I don't have any objection to requiring NSEC3 or SHA-2 algorithm support for the reasons you noted. My point is that the current draft language does not establish any MUST-level requirements, because the referenced sections in RFC 6840 all contain SHOULD-level language. SHOULD-level "strong encouragements" do not become MUST-level requirements solely due to the passage of time. In a similar vein, updates to IANA registries (or any other RFC publication action) do not translate to BR normative changes absent BR language to require adherence to the MTI elements in the registry. 

Instead, I think this ballot should instead directly reference the appropriate section(s) of the appropriate RFCs that define NSEC3 and the use of SHA-2. This will make the requirement much easier to understand.

birgelee: @CBonnell that makes sense. Thanks for clarifying. I will push a change to have the language directly reference the SHA-2 and NSEC3 RFCs.

xiaohuilam:
Hi all,  

I wonder what this ballot is supposed to require CAs to do in the following situations:  

- If a TLD has misconfigured DNSSEC (e.g., the [.CN domain's DNSSEC](https://dnsviz.net/d/cn/dnssec/));  
- If a domain’s DNS server rejects DNSSEC typed queries, yet the domain itself is DNSSEC unsigned.

birgelee:
@xiaohuilam 

First and foremost I want to clarify that from the purposes of this ballot (and I would argue for DNSSEC in general), ".cn" is **not** misconfigured. The notices presented by dnsviz.net on this TLD are intended for DNS zone operators to improve their DNSSEC configuration. ".cn" does not lead to a lookup error on any DNSSEC validating resolver. In fact, the inclusion of RFC5155 in this ballot text ensures that CAs will resolve this zone correctly with a NOERROR response code. The proper configuration of .cn can be confirmed with the following command:

```
./unbound-host -v -D -t A cn
cn has no address (secure)
```

Notice the unbound host tool properly accepted the non-existence proof showing the domain has no IPv4 A address and marked the domain as “secure” (i.e., having a valid DNSSEC signature chain).

DNSSEC and non-DNSSEC domains under “.cn” also show NOERROR response codes and valid signatures. For example:

```
./unbound-host -v -D -t A sjtu.edu.cn
sjtu.edu.cn has address 202.120.2.119 (secure)
```

sjtu.edu.cn is DNSSEC signed under .cn and resolves fine.

At the end of this response, I have a more detailed discussion of the DNSSEC configuration on .cn, but it will not cause a lookup error on proper resolver implementations.

Regarding these specific points:

- If a TLD truly has a bogus DNSSEC configuration: This will cause a lookup error for domains under this TLD as the DNSSEC signature chain cannot be established. As I discussed earlier, .cn is not an example of this situation, and I know of no such contemporary instances of DNSSEC resolution errors at the TLD level. This ballot is also written in a way to allow resolvers to return NOERROR wherever possible without degrading security (e.g., BCPs and updates to DNSSEC mandating more specific behaviors have been omitted).

- If a domain’s DNS authoritative server rejects DNSSEC queries, and the domain itself is DNSSEC unsigned: This case is perfectly fine and will lead to normal certificate issuance with no lookup errors. A domain’s DNSSSEC state is established by the parent zone via the inclusion of a DS record at that domain name. Presuming the parent zone is a properly DNSSEC-configured TLD (or even just a non-DNSSEC TLD), the recursive resolver will know not to expect any DNSSEC data at the domain because there is no DS record. Conventional non-DNSSEC resolution behavior is used when talking to this domain’s servers. The authoritative servers do not need to answer any DNSSEC-related queries since they did not opt in. If a domain was configured with authoritative servers that had no knowledge of DNSSEC, that domain will still resolve fine even with DNSSEC validation because the recursive resolver knows not to expect any signature on that domain. 

--------------------

Regarding the .cn DNSSEC configuration:

As I stated above, this DNSSEC configuration is perfectly valid. DNSVis’s complaint with the domain ([seen when the non-existence setting is turned on](https://dnsviz.net/d/cn/dnssec/?rr=all&a=all&ds=all&doe=on&ta=.&tk=)) is that it does not follow RFC 9276 which is Best Current Practice 236 (BCP 236). Specifically, .cn uses the hash-iterations field of its NSEC3 record (which is well defined and standardized in RFC 5155 and would be supported by this ballot). BCP 236 states that domains MUST use an iteration count of 0 (i.e., not using hash iterations). .cn uses an iteration count of 10, but this is actually pretty standard and I have even seen [10 recommended elsewhere](https://dnsinstitute.com/documentation/dnssec-guide/ch06s02.html). Ironically, increasing the hash iteration count actually provides additional security against zone enumeration attacks for the .cn zone although presents a slightly higher load to DNSSEC-validating resolvers which is why it was set to zero in BCP 236.

Importantly:
- RFC 9276 is not Standards Track. It offers configuration guidance and is not redefining how DNSSEC validation is performed. 
- RFC 9276 Section 3.1 (cited by DNSVis) is providing guidance to domain owners, not DNSSEC validators. Even under RFC 9276, a compliant DNSSEC validator will still accept non-zero hash iterations.
- RFC 9276 is not cited in this ballot. This ballot is intentionally written in a minimally-prescriptive manner to avoid complications related to the publications of new BCPs and updates to the standard. 

Below is a full DNSSEC-enabled query for .cn and a DNSSEC-signed domain under .cn. Both return NOERROR codes and would not cause a validation problem by CAs.

```
henry@MacBook-Pro-259 ~ % dig +dnssec @8.8.8.8 cn IN A

; <<>> DiG 9.10.6 <<>> +dnssec @8.8.8.8 cn IN A
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 34587
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 0, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 512
;; QUESTION SECTION:
;cn.            	IN	A

;; AUTHORITY SECTION:
cn.        	1800	IN	SOA	a.dns.cn. root.cnnic.cn. 2033333274 7200 3600 2419200 21600

cn.        	1800	IN	RRSIG	SOA 8 1 86400 20250524025426 20250424015426 38388 cn. p3A7pKZ8v10ImFb5ybHEzUWI8YBnZllWgXd+UMGWKcY4CCeGBNrH2L4D dgqmKCeUDwTqliQmtCa8rGzOYE893C/mzHhmA8YQ59iHEmR/nUrUfj5/ x9589ltnTN3OAMdz79WekWpg1Iq6rv+euWQMRdFas9OBwxRBCiy6srCE DM8=

3QDAQA092EE5BELP64A74EBNB8J53D7E.cn. 21600 IN NSEC3 1 1 10 AEF123AB 3QM14FQ32F1CJFTP8D3J5BCTNP5BIELO  NS SOA RRSIG DNSKEY NSEC3PARAM

3QDAQA092EE5BELP64A74EBNB8J53D7E.cn. 21600 IN RRSIG NSEC3 8 2 21600 20250517012154 20250417004141 38388 cn. dfJK2fip7I07CNCsBLpqGrVEggXoGtPtk+/ILHrR320pMdQopfWxChJn IwfPYcvZIuMevjoRDIvGq3fWE3yCj4zjF9t1or42sT7UHrIx5Y8r++gP YVgxVO4RCEUpcSkQvTT6fIOLiFoSjhPUUKtg2kIR/jBxoXITcOBlANr4 uHw=

;; Query time: 63 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Wed Apr 23 22:55:30 EDT 2025
;; MSG SIZE  rcvd: 492

henry@MacBook-Pro-259 ~ %
```







```
henry@MacBook-Pro-259 ~ % dig +dnssec @8.8.8.8 sjtu.edu.cn IN A

; <<>> DiG 9.10.6 <<>> +dnssec @8.8.8.8 sjtu.edu.cn IN A
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 50607
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 512
;; QUESTION SECTION:
;sjtu.edu.cn.        	IN	A

;; ANSWER SECTION:
sjtu.edu.cn.    	300	IN	A	202.120.2.119
sjtu.edu.cn.    	300	IN	RRSIG	A 8 3 300 20250523085516 20250423085516 9528 sjtu.edu.cn. Zsr8Aw3ursIiAE0TX8mryThHIWz7rCHhGfi+bRuAJNwZQtctK3hyTg9R aL4RGFPZrHs+T9dBGbGXawv83Vr26wY6C+FQI3Ah3t8uiQelWJ4yWefJ bwZPcN/ot6QnaBIcbQZjQNY0ObrCOPAjaYZq3W5708J5aSRJumwrM9pI Z2k=
```

geegeea:
Henry, thank you for putting together this proposal. GTS fully supports requiring DNSSEC validation for the additional security properties it provides. It has been mentioned that using an open source resolver should allow CAs to meet the requirements of this proposal. Coincidentally, GTS went through this exercise several months ago when we started the process of moving to a new DNS resolution solution using an open source resolver.

GTS has performed DNSSEC validation for several years now, so DNSSEC support was an important consideration when evaluating the available resolvers. I will share our analysis below to highlight some implementation considerations that may come up.

After eliminating resolvers which were not options for us due to license restrictions (this includes PowerDNS which uses the GNU GPLv2 license), we evaluated the following list:

Bind
Unbound
DJBDNS - Didn’t support DNSSEC out of the box
Knot - Had open DNSSEC related issues at the time of evaluation: https://gitlab.nic.cz/knot/knot-resolver/-/issues/910, this issue is now closed.
Hickory - At the time of evaluation, their site said it was not production ready. 
CoreDNS - Does not function as a recursive resolver without a plugin such as libunbound.

We found Bind and Unbound to be the only feasible choices. Unbound has some level of security audit: https://ostif.org/our-audit-of-unbound-dns-by-x41-d-sec-full-results/ and Bind has extra functionality which we didn’t need and we didn’t want the additional attack surface. So we chose Unbound. One consideration to be aware of with this, is if a lot of CAs land on the same open source resolver, then a vulnerability or a bug in that resolver would render all of those CAs unable to operate, and possibly require mass revocations.

CBonnell:
Hickory-DNS still carries the warning that it is not production ready: https://github.com/hickory-dns/hickory-dns/blob/main/bin/README.md?plain=1#L48

birgelee:
@geegeea
Thanks for sharing GTS's open source DNS analysis. I wanted to add some commentary regarding, testing an arbitrary deployment for DNSSEC support and the impact of the DNSSEC support requirement on selection of DNSSEC software.

## DNSSEC Validation Testing

A purpose of this ballot is to implement DNSSEC in a way that is easy to verify and test. There are several services for testing the DNSSEC validation behavior of a resolver. One such service is [Check My DNS by DNS OARC](https://cmdns.dev.dns-oarc.net/) which can be run in a browser and does check for DNSSEC support (as well as other good security features of DNS resolvers).

Below I list several commands to test for specific DNSSEC features. In the below commands, replace `<candidate-DNSSEC-validating-resolver>` with the IP address of the DNS resolver being tested.

1. CAA Test suite's DNSSEC tests. CAA test suite (caatestsuite.com) already contains DNSSEC tests to test the DNSSEC validation behavior of DNS resolvers used to retrieve CAA records. `expired.caatestsuite-dnssec.com`, 
`missing.caatestsuite-dnssec.com`, `blackhole.caatestsuite-dnssec.com`, `servfail.caatestsuite-dnssec.com`, and `refused.caatestsuite-dnssec.com` should all return resolution failures that are not treated as permission to issue when queried with a DNSSEC-validating resolver. These tests can be run using `dig <caa-test-suite-domain> IN CAA @<candidate-DNSSEC-validating-resolver>`. A DNSSEC validating resolver will return an error response with `status: SERVFAIL` in the DIG response for all of these queries.

2. Simple successful queries of DNSSEC-signed domains. The CAA test suite examples include DNSSEC failure modes where a non-DNSSEC-validating resolver may return a successful query, but proper DNSSEC validation makes the result a SERVFAIL due to invalid signatures. It is also important to confirm the positive capabilities required of the DNSSEC validator. Simply querying any DNSSEC-signed domain can largely confirm this. One specific example is `cloudflare.com` which is DNSSEC signed. The command `dig cloudflare.com IN CAA @<candidate-DNSSEC-validating-resolver>` should return 1) CloudFlare's CAA records, 2) `status: NOERROR` for the status code and, 3) the AD flag shows by dig as `flags: ... ad ...`. There can be other flags present in the flags list, but the `ad` flag should be included to indicate the resolver successfully performed validation and determined that domain was secure. The DNSSEC chain to CloudFlare's domain contains DS records with digest type 2 (thus confirming the DS-record SHA-2 support required by this ballot) and RRSIG records with algorithm type 8 from gtld-servers.net (thus confirming the algorithm 8 support required by this ballot). CloudFlare also uses algorithm type 13 which is very popular and good to be supported although not required by this  ballot. Seeing the `status: NOERROR` and `ad` flag implies support for all three of these techniques.

3. Successful queries of non-existent record sets using NSEC and NSEC3. The above commands do not explicitly test handling of NSEC or NSEC3 records because in all cases either the requested records existed or there was a DNSSEC validation error. CloudFlare uses NSEC and the proper handling of this can be tested using `dig xyz.cloudflare.com in CAA +dnssec @<candidate-DNSSEC-validating-resolver>` (where xyz is some non-existent CloudFlare subdomain). Here the appropriate response from a DNSSEC-validating resolver is `status: NOERROR`, `flags: ... ad ...`, and `ANSWER: 0` in the header section of the dig response. The DNS-validating resolver should also the relevant RRSIG and NSEC records proving non-existence, but the header checks alone ensure these records were validated successfully. To test NSEC3 behavior,  `dig com IN CAA @<candidate-DNSSEC-validating-resolver>` will produce a negative result with NSEC3 records. For those who are interested, `cn` can be used in place of `com` to get a response with the same header, reaffirming `cn`'s valid DNSSEC configuration (i.e., `dig cn IN CAA @<candidate-DNSSEC-validating-resolver>`). This response should similarly show `status: NOERROR`, `flags: ... ad ...`, and `ANSWER: 0` in the header section of the dig response.


To make the above-mentioned testing process automated, you can see the script at [https://github.com/birgelee/dnssec-validator-test](https://github.com/birgelee/dnssec-validator-test) for an example. To be clear, the best way to verify the required DNSSEC support is by documentation and confirmation with the project maintainers. These tests are not exhaustive.


## Impact of the DNSSEC requirement on the available options of DNS software

Regarding the analysis presented by @geegeea , I want to mention that I **do not** think Knot Resolver should be excluded based on the now patched DNSSEC bug @geegeea mentioned. Knot resolver powered CloudFlare's 1.1.1.1 for a long time (https://blog.cloudflare.com/big-pineapple-intro/), is deployed by many large organizations, and offers an SLA for enterprises (https://www.knot-resolver.cz/support/pro/). Additionally, the maintainers of Knot Resolver (CZ.NIC) also provide one of the most widely used linux software routers (Bird) and have a fairly good reputation. Even reputable projects occasionally see bugs.

Unbound and Bind, mentioned by @geegeea are also good alternatives. The Open MPIC project maintains an Unbound container (https://github.com/open-mpic/open-mpic-containers/tree/main/unbound), and Bind is probably the oldest and most established DNS resolver option. As @geegeea mentioned, people sometimes prefer lighter-weight resolvers like Unbound or Knot due to bloat concerns with Bind which 1) packages its recursive resolver with its authoritative DNS implementation and 2) has an extensive feature set which may not be needed in all cases.

Hickory DNS is not production ready but will likely be the chosen option by Let's Encrypt once its development is complete (it is funded by the same parent organization ISRG). One advantage of Hickory DNS is that it is written in Rust, a memory safe language, whereas many of the other options are C or C++ codebases (primarily for efficiency). 

At a higher level, I think the most relevant consideration for this ballot is: to what extent does the DNSSEC requirement limit the number of DNS resolvers available to CAs? I would personally argue that requiring DNSSEC in fact assists with good DNS resolver choice and DNS resolvers that have not implemented DNSSEC are likely poorly maintained or not intended for enterprise use. E.g., DJBDNS does not natively support DNSSEC and the project is [often considered no longer maintained (although many patches exist)](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=256450).

While not an authoritative source, Wikipedia provides a very rough overview and lists [14 options for recursive DNS resolvers](https://en.wikipedia.org/wiki/Comparison_of_DNS_server_software) (13 if you count dbndns and djbdns as the same). Of these 13 options, only 4 lack DNSSEC support including options like djbdns (discussed earlier), Posadis ([which has not had a new release since 2013](https://sourceforge.net/projects/posadis/)) and Mara DNS ([which seems to have one maintainer](https://github.com/samboy/MaraDNS)). Obviously, choice of DNS resolver is dependent on the needs of a CA, but I do think there are several options for DNS resolvers that support DNSSEC, and I feel DNSSEC support does not significantly limit DNS resolver options. Furthermore, many DNS resolvers specifically list RFCs they comply with ([[Hickory](https://github.com/hickory-dns/hickory-dns)] [[Bind](https://bind9.readthedocs.io/en/v9.18.15/general.html)] [[Unbound](https://unbound.docs.nlnetlabs.nl/en/latest/reference/rfc-compliance.html)] [[Power DNS](https://www.powerdns.com/compliance)], [[Knot](https://www.knot-resolver.cz/documentation/latest/rfc-list.html)]) or make explicit statements about their support for DNSSEC (e.g., [[Technitium](https://technitium.com/dns/)], [[MS DNS](https://learn.microsoft.com/en-us/windows-server/networking/dns/validate-dnssec-responses)], [[Secure64 Cache](https://secure64.com/wp-content/uploads/2023/06/Secure64-DNS-Cache-Product-Data-Sheet-1.pdf)])). DNS resolver project maintainers can verify the required feature set for a CA. To clarify, the features mentioned in this ballot are quite standard on modern DNS resolvers.

While I do agree that there is some centralization in the DNS ecosystem particularly when constraints like licensing terms are imposed, I do not feel the DNSSEC requirement significantly reduces the available options of DNS resolvers and I think DNSSEC support is a good litmus test for whether a DNS resolver is well maintained and suitable for enterprise use.

geegeea:
Henry, thank you for the commentary. Is it correct to summarize from the comment that resolvers like Unbound, Bind, and Knot are all good candidates for meeting the requirements of this ballot, but to definitively say so we would need to refer to documentation or confirm with maintainers?

Regarding resolver choice, we agree with the assessment that:
The most relevant consideration for this ballot is: to what extent does the DNSSEC requirement limit the number of DNS resolvers available to CAs?
Requiring DNSSEC assists with good resolver choice

There is an assumption in the commentary that CAs all use third party resolvers (third party meaning a resolver which is developed outside of the CA’s org, not the BR term “delegated third party”). This may not be true, some CAs could have in-house developed resolvers, and depending on the enforcement timeline, this ballot could push most CAs to use third party resolvers. This may not be a significant risk for this ballot, but I want to point it out so the choice is intentional.

docs/BR.md
```diff
@@ -731,6 +741,17 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
 
 Completed validations of Applicant authority may be valid for the issuance of multiple Certificates over time. In all cases, the validation must have been initiated within the time period specified in the relevant requirement (such as [Section 4.2.1](#421-performing-identification-and-authentication-functions) of this document) prior to Certificate issuance. For purposes of domain validation, the term Applicant includes the Applicant's Parent Company, Subsidiary Company, or Affiliate.
 
+Effective November 15th, 2025: DNSSEC validation back to the IANA DNSSEC root trust anchor MUST be performed on all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective. The DNS resolver used for all DNS queries associated with the validation of domain authorization or control by the Primary Network Perspective MUST:
```
geegeea, line 744: Section 8.7 of the BRs requires the CA to monitor adherence to these requirements by performing self-audits on 3% of certificate issuances. Under the current ballot text, will a CA be required to also verify DNSSEC verification as part of these self-audits? Doing so will likely require saving all DNSSEC lookup records, and re-validating them during the self-audit. Can it be made explicit in the ballot text whether this will be expected or not?

birgelee: Hi @geegeea 

Thanks for following up. To briefly clarify the comment I made about checking with documentation or maintainers. In the case of [[Hickory](https://github.com/hickory-dns/hickory-dns)] [[Bind](https://bind9.readthedocs.io/en/v9.18.15/general.html)] [[Unbound](https://unbound.docs.nlnetlabs.nl/en/latest/reference/rfc-compliance.html)] [[Power DNS](https://www.powerdns.com/compliance)], [[Knot](https://www.knot-resolver.cz/documentation/latest/rfc-list.html)], all the RFCs listed in this ballot appear on their compliance lists and I feel this is sufficient to show these choices achieve compliance. Many other DNS resolvers state DNSSEC support but don't provide the same level of detail regarding their compliance. This is where I would recommend contact with maintainers. The RFCs required by this ballot are quite standard and I think it is unlikely a well maintained DNS resolver with DNSSEC support would have issue with them.

Regarding first-party-maintained resolvers: I encourage any CA in such a situation to speak up. I will say I think in-house developed resolvers do expose a CA to some risk of misissuance given the frequency of vulnerabilities and Best Current Practice updates regarding the DNS protocol. Some practices that were once commonplace for DNS resolvers are now serious security vulnerabilities (e.g., fixed source ports). If this ballot does encourage a push to externally-maintained DNS resolvers, that may even be beneficial to the ecosystem. Also, to clarify, no text in this ballot precludes the use of an in-house DNS resolver, but I do understand that CAs in this position may need longer to invest the required engineering resources.

I do not think DNS records should be saved for the purpose of the self audit. Many DNS resolvers do not produce a validation attestation and the volume of saving all DNS records is extremely high. In DNSSEC, my understanding is that DNS resolvers have the right to cache DNSSEC signed results without storing the underlying signatures used to validate those results (presuming the signatures verified correctly). Thus in many implementations it is non-trivial to simply ask the resolver to produce an entire validation chain particularly if intermediate results were not stored at the time the lookups were performed.

docs/BR.md
```diff
@@ -1132,16 +1132,16 @@ CAs MUST document potential issuances that were prevented by a CAA record in suf
...
 * support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
 * properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4).
 
-Effective November 15th, 2025: CAs MUST NOT use local policy to disable DNSSEC validation on any DNS query associated CAA record lookups.
+Effective March 15th, 2026: CAs MUST NOT use local policy to disable DNSSEC validation on any DNS query associated CAA record lookups.
 
-Effective November 15th, 2025: DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
+Effective March 15th, 2026: DNSSEC-validation errors observed by the Primary Network Perspective (e.g., SERVFAIL) MUST NOT be treated as permission to issue.
 
 DNSSEC validation back to the IANA DNSSEC root trust anchor MAY be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration.
```
geegeea, line 1146:
```suggestion
DNSSEC validation back to the IANA DNSSEC root trust anchor MAY be performed on all DNS queries associated with CAA record lookups performed by Remote Network Perspectives as part of Multi-Perspective Issuance Corroboration.

DNSSEC validation back to the IANA DNSSEC root trust anchor is considered outside the scope of self-audits performed to fulfill the requirements in [Section 8.7](#87-self-audits).
```

docs/BR.md
```diff
@@ -731,6 +741,17 @@ The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qua
...
+
+* perform DNSSEC validation using the algorithm defined in [RFC 4035 Section 5](https://datatracker.ietf.org/doc/html/rfc4035#section-5); and
+* support NSEC3 as defined in [RFC 5155](https://datatracker.ietf.org/doc/html/rfc5155); and 
+* support SHA-2 as defined in [RFC 4509](https://datatracker.ietf.org/doc/html/rfc4509) and [RFC 5702](https://datatracker.ietf.org/doc/html/rfc5702); and
+* properly handle the security concerns enumerated in [RFC 6840 Section 4](https://datatracker.ietf.org/doc/html/rfc6840#section-4).
+
+Effective March 15th, 2026: CAs MUST NOT use local policy to disable DNSSEC validation on any DNS query associated with the validation of domain authorization or control.
+
+DNSSEC validation back to the IANA DNSSEC root trust anchor MAY be performed on all DNS queries associated with the validation of domain authorization or control by Remote Network Perspectives used for Multi-Perspective Issuance Corroboration.
+
```
geegeea, line 754: Apologies I missed adding this in the DCV section in my earlier comment.

```suggestion

DNSSEC validation back to the IANA DNSSEC root trust anchor is considered outside the scope of self-audits performed to fulfill the requirements in [Section 8.7](#87-self-audits).
```

PR #501, opened 2024-04-18
Ballot SC-082:  Clarify CA Assisted DNS Validation under 3.2.2.4.7

slghtr-says:
Ballot SC-082: Clarify CA Assisted DNS Validation under 3.2.2.4.7

Previous Discussion Thread: 
https://github.com/slghtr-says/servercert/pull/1

### Background 

**CA Assisted DNS Validation** is the practice where Certification Authorities (CAs) instruct Applicants to create Canonical Name (CNAME) records specifically for the purpose of assisting the Applicant with Domain Control Verification (DCV) of their domain.

At F2F 59 (July 23’), the Validation Subcommittee of the Server Certificate WG presented the following conclusions on the practice of CA Assisted DNS Validation: 
* More clarity is needed around the practice
* Applicants generally delegate the performance of many aspects of operating a website.
* If done correctly, allowing Applicants to delegate the placement of the Random Value/ Request Token boosts agility and automation. 
* There are reasonable interpretations of the BRs that such delegation is already allowed today. 

A Tiger Team was formed to threat model CA Assisted DNS Validation and propose modifications to the BRs to add clarity and constraints around the practice. The results of the threat model exercise [1] were presented and discussed at F2F 60 [2] and F2F 61 [3].

### Purpose of Ballot 

The purpose of this ballot is to clarify the practice of CA Assisted DNS Validation and add constraints under Method 7 (3.2.4.4.7 DNS Change). Modification of other domain validation methods and the introduction of new domain validation methods are not in scope of this ballot but may be addressed in a future ballot. 

### Overview of Changes 
* New definition: Canonical Authorization Domain Name
* Addition of Canonical Authorization Domain Names into section 3.2.2.4.7 (DNS Change)
* Addition of constraints around the usage of Canonical Authorization Domain Names by CAs
  * Unique to an Applicant and not shared with multiple Applicants
  * DNS lookup results expire after 8 hours 
  * Restrictions on the type of DNS records located in zones for this purpose.

### Example

**Canonical Authorization Domain Name**: The domain name found within the RDATA value of a CNAME record located at an Authorization Domain Name that is prefixed with a Domain Label that begins with an underscore character.

Given an attempt to perform DCV for ``example.com`` using 3.2.2.4.7, 

The applicant inserts the following resource record in the public DNS zone of ``example.com``:

``_underscore-prefixed-domain-label.example.com. IN CNAME account-binding-id.cadomain.com``

where ``account-binding-id.cadomain.com`` is a resource record of the following form:

[...]

PR #608, opened 2025-08-06
SC-088: DNS TXT Record with Persistent Value DCV Method

slghtr-says:
Add 3.2.2.4.22 [SC-088] DNS TXT Record with Persistent Value DCV Method

## Background 

Previous Ballot (SC-082): https://github.com/cabforum/servercert/pull/501

## Purpose of Ballot 

This ballot introduces 3.2.2.4.22 DNS TXT Record with Persistent Value to the Baseline Requirements. DNS TXT Record with Persistent Value offers significant operational benefits while maintaining the existing security posture of the web PKI ecosystem. 

## Overview
Section 3.2.2.4.22 introduces a new domain control validation (DCV) method built around persistent DCV. Certificate Authorities may use this method to validate an applicant's control over a domain by verifying the presence of an applicant specific persistent DNS TXT record in the public DNS of a domain. 

### What's New
• **DCV Method Name**: "DNS TXT Record with Persistent Value"
• **DNS TXT Record Name **: _validation-persist.<domain-being-validated>
• **DNS TXT Record Value **: "authority.example; accounturi=https://authority.example/acct/123"

### DNS Record Requirements
1. Syntax: Must conform to RFC 8659 issue-value syntax
2. Issuer Domain: Must match CA's disclosed issuer domain name in their CP/CPS
3. Account URI: Must contain unique URI identifying the applicant's account 

### Validation Properties

* **persistUntil Attribute**: Optional. A unix timestamp after which the record can no longer be used for a DCV check 
* **Validation Reuse**: If DNS TXT record TTL < validation reuse period, reuse period = max(TTL, 8 hours)
• **Multi-Perspective Validation**: Required (same as other DNS methods)
• **Supports Wildcard Domains**: Yes - can validate wildcard certificates

### Example Validation Records

|Example record|DCV Will Succeed Until |Validation Reuse Period| 
| - | - | - |
|TXT "authority.example; accounturi=https://authority.example/acct/123" <br> TTL=300 | N/A | 8 hours |
|TXT "authority.example; accounturi=https://authority.example/acct/123" <br> TTL=63072000 | N/A | <maximum domain validation reuse period as specified in section 4.2.1> |
|TXT "authority.example; accounturi=https://authority.example/acct/123; persistUntil=1582428142" <br> TTL=300 | Sun Feb 23 2020 03:22:22 GMT+0000 | 8 hours |
|TXT "authority.example; accounturi=https://authority.example/acct/123; persistUntil=1782424856" <br> TTL=300 | Thu Jun 25 2026 22:00:56 GMT+0000 | 8 hours |
|TXT "authority.example; accounturi=https://authority.example/acct/123; persistUntil=1782424856" <br> TTL=604800 | Thu Jun 25 2026 22:00:56 GMT+0000 | 7 days |

### Optional Attribute: persistUntil
persistUntil(1)

### TTL Based Validation Reuse Period
TTLreusePeriod

### Attack Surface Assessment

Analysis: https://groups.google.com/a/groups.cabforum.org/g/servercert-wg/c/4tATbCpQRpM/m/8mH-SX87BAAJ?utm_medium=email&utm_source=footer

No Expansion of Attack Surface: Security analysis demonstrates that any adversary capable of compromising this method could already compromise existing methods:
• 3.2.2.4.14 (Email to DNS TXT Contact) 
• 3.2.2.4.15 (Phone Contact with Domain Contact)
• 3.2.2.4.7 (DNS Change) and ACME dns-01

### Security Benefits
1. Reduced DNS Credential Risk: Eliminates need for overly powerful DNS credentials on web infrastructure
2. No Third-Party Delegation: Avoids delegating _acme-challenge management to external services
3. Enhanced Transparency: Persistent records enable better monitoring and misuse detection
5. Enterprise LAN Support: Enables HTTPS for internal-only domains where HTTP/TLS-ALPN validation isn't possible

### Automation & Certificate Lifetimes
• Significantly simplifies certificate issuance process
• Supports push toward shorter certificate lifetimes
• Reduces barriers to automated certificate management

### Transparency Improvements
• Provides more public visibility into DCV strategy for domains
• Enables more third-party detection of certificate misuse

### Use Case Support
• **Shorter Lived Certificate Automation**: Reduces operational overhead for frequent certificate renewal
• **Enterprise LANs**: Domains behind firewalls that can't use HTTP validation
• **IoT Devices**: Simplified certificate management for constrained environments

## Implementation Notes
• Method is optional - CAs choose whether or not to support and domain owners choose whether to use it
• Compatible with existing CAA and validation infrastructure
• Follows established patterns from appendix A.2.3 for DNS TXT record placement

# References 
SC-082 Redux
*  https://groups.google.com/a/groups.cabforum.org/g/servercert-wg/c/4tATbCpQRpM/m/8mH-SX87BAAJ?utm_medium=email&utm_source=footer
* [Static DCV Security Proof](https://12458557179442167391.googlegroups.com/attach/5262e7c1f1c1/Proof%20of%20static%20DCV%20security.pdf?part=0.1&v
* https://datatracker.ietf.org/doc/draft-sheth-identifiers-dns/
SC-082 V1 
* SC-082 Ballot: https://github.com/cabforum/servercert/pull/501

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -368,6 +368,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **DNS TXT Record Phone Contact**: The phone number defined in [Appendix A.2.2](#a22-dns-txt-record-phone-contact).
 
+**DNS TXT Record Persistent DCV Domain Label**: The Domain Label defined in [Appendix A.2.3](#a23-dns-txt-record-persistent-dcv-domain-label).
+
 **Domain Contact**: The Domain Name Registrant, technical contact, or administrative contact (or the equivalent under a ccTLD) as listed in the WHOIS record of the Base Domain Name or in a DNS SOA record, or as obtained through direct contact with the Domain Name Registrar.
 
 **Domain Label**: From RFC 8499 (<https://tools.ietf.org/html/rfc8499>): "An ordered list of zero or more octets that makes up a portion of a domain name. Using graph theory, a label identifies one node in a portion of the graph of all possible domain names."
@@ -846,6 +848,8 @@ If a Random Value is used, the CA SHALL provide a Random Value unique to the Cer
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e. Random Value or Request Token) as the Primary Network Perspective.
 
+If the CA or an Affiliate of the CA operates a DNS zone to which Applicants can delegate (via CNAME) their underscore-prefixed Domain Label, the CA MUST ensure that each Applicant delegates to a unique FQDN within that zone. A CA or Affiliate of a CA SHOULD NOT operate such a service, and SHOULD direct any Applicants using such a service to use the method described in [Section 3.2.2.4.22](#322422-dns-change-with-persistent-value) instead.
+
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
 ##### 3.2.2.4.8 IP Address
@@ -1028,6 +1032,47 @@ The token (as defined in draft 00 of “Automated Certificate Management Environ
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same token as the Primary Network Perspective.
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+
+##### 3.2.2.4.22 DNS TXT Record with Persistent Value
+
+Confirming the Applicant's control over a FQDN by confirming the presence of a persistent record identifying the Applicant in a DNS TXT record for an Authorization Domain Name that is prefixed with one DNS TXT Record Persistent DCV Domain Label.
+
+The CA MUST confirm the presence of a TXT record whose RDATA value fulfills the following requirements:
+
+1. The RDATA value MUST conform to the `issue-value` syntax as defined in RFC 8659, section 4.2; and
+2. The `issuer-domain-name` value MUST be an Issuer Domain Name disclosed by the CA in Section 4.2 of the CA's Certificate Policy and/or Certification Practices Statement; and
+3. The `issue-value` MUST contain an `accounturi` parameter, where the parameter value is a unique URI (as described by RFC 8657, Section 3) identifying the account of the Applicant which requested validation for this FQDN; and
+4. The `issue-value` MAY contain a `persistUntil` parameter. If present, the parameter value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds).
+
+If the `persistUntil` parameter is present, the CA MUST evaluate its value. If the time of the check is after the time specified in the `persistUntil` parameter value, the CA MUST NOT use the record as evidence of the Applicant's control over the FQDN.
+For example, the DNS TXT record might look like:
+_validation-persist.example.com IN TXT "authority.example; accounturi=https://authority.example/acct/123; persistUntil=1782424856"
+
+CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information as the Primary Network Perspective.
+
+If the DNS TXT record has a TTL less than the validation data reuse period (see [Section 4.2.1](#421-performing-identification-and-authentication-functions)), then the CA MUST consider the validation data reuse period to be equal to the TTL or 8 hours, whichever is greater.
+
+The following table shows how the `persistUntil` parameter affects whether a DNS record can be used for validation at different points in time:
+
+Table: Examples of how the `persistUntil` parameter affects validation
+
+| __Date/time of validation__ | __persistUntil__ | __Usable for validation__ | __Explanation__ |
+|----------------------------|------------------|--------------------------|----------------|
+| 2025-06-15T12:00:00Z | 2026-01-01T00:00:00Z (1767225600) | Yes | Validation time is before persistUntil timestamp, so record is usable |
+| 2025-06-15T12:00:00Z | 2025-01-01T00:00:00Z (1735689600) | No | Validation time is after persistUntil timestamp, so record is not usable |
+| 2025-06-15T12:00:00Z | (not present) | Yes | No persistUntil parameter present, so no time restriction applies |
+
+The following table shows how the DNS TXT record TTL affects the validation data reuse period:
+
+Table: Examples of how the TTL affects the validation data reuse period
+
+| __Record TTL__ | __Maximum data reuse period ([Section 4.2.1](#421-performing-identification-and-authentication-functions))__ | __Effective validation data reuse period__ | __Explanation__ |
+|---------------|------------------------------------------------------------------------------------------------------------------------|------------------------------------------|----------------|
+| 14400 (4 hours) | 200 days | 28800 seconds (8 hours) | TTL is below 8-hour minimum, so minimum 8 hours applies |
+| 86400 (24 hours) | 200 days | 86400 seconds (24 hours) | TTL is greater than minimum but less than maximum, so TTL applies |
+| 20736000 (240 days) | 200 days | 17280000 seconds (200 days) | TTL exceeds 200-day maximum, so maximum 200 days applies |
+
+**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names. 
 
 #### 3.2.2.5 Authentication for an IP Address
 
@@ -3946,6 +3991,10 @@ The DNS TXT record MUST be placed on the "`_validation-contactemail`" subdomain 
 ### A.2.2. DNS TXT Record Phone Contact
 
 The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain of the domain being validated. The entire RDATA value of this TXT record MUST be a valid Global Number as defined in RFC 3966, Section 5.1.4, or it cannot be used.
+
+### A.2.3. DNS TXT Record Persistent DCV Domain Label
+
+The DNS TXT record MUST be placed at the "_validation-persist" label prepended to the Authorization Domain Name being validated.
 
 # APPENDIX B – Issuance of Certificates for Onion Domain Names
 


jesperkristensen:
Could this add a recommendation for CAs to support defenses against the confused deputy problem? With DNS CNAME delegation for example, an applicant who need such defense can choose to implement the CNAME delegation on their own side instead of allowing the CA to host it, if the CA does not support this defense. But with this new validation method it does not seem possible for the applicant to implement confused deputy defenses unless the CA supports it. An example could be a cloud provider using the same CA accounturi across tenants, but want to prevent certificates from being assigned to the wrong tenant.

docs/BR.md
```diff
@@ -1029,6 +1033,47 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
...
+##### 3.2.2.4.22 DNS TXT Record with Persistent Value
+
+Confirming the Applicant's control over a FQDN by confirming the presence of a persistent record identifying the Applicant in a DNS TXT record for an Authorization Domain Name that is prefixed with one DNS TXT Record Persistent DCV Domain Label.
+
+The CA MUST confirm the presence of a TXT record whose RDATA value fulfills the following requirements:
+
+1. The RDATA value MUST conform to the `issue-value` syntax as defined in RFC 8659, section 4.2; and
+2. The `issuer-domain-name` value MUST be an Issuer Domain Name disclosed by the CA in Section 4.2 of the CA's Certificate Policy and/or Certification Practices Statement; and
+3. The `issue-value` MUST contain an `accounturi` parameter, where the parameter value is a unique URI (as described by RFC 8657, Section 3) identifying the account of the Applicant which requested validation for this FQDN; and
+4. The `issue-value` MAY contain a `persistUntil` parameter. If present, the parameter value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds).
```
slghtr-says:
```suggestion
4. The `issue-value` MAY contain a `persistUntil` parameter. If present, the parameter value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds).
5. The `issue-value` MAY contain additional parameters. CAs MUST ignore any unknown parameter keys.
```
Rationale: Allows for forward compatibility with additional parameters and clarifies the expected behaviors for CAs when unknown keys are encountered.

slghtr-says:
> Could this add a recommendation for CAs to support defenses against the confused deputy problem? With DNS CNAME delegation for example, an applicant who need such defense can choose to implement the CNAME delegation on their own side instead of allowing the CA to host it, if the CA does not support this defense. But with this new validation method it does not seem possible for the applicant to implement confused deputy defenses unless the CA supports it. An example could be a cloud provider using the same CA accounturi across tenants, but want to prevent certificates from being assigned to the wrong tenant.

Thank you for the confused deputy feedback. In your cloud provider example, the cloud provider is the Applicant - the CA issues the certificate to them, not to their tenants.

CAs have no visibility into how Applicants structure their internal operations or tenant isolation. The cloud provider (as Applicant) is responsible for implementing appropriate confused deputy defenses in their own architecture. While this validation method may have different architectural constraints than CNAME delegation, the fundamental responsibility for tenant isolation remains with the Applicant.

Adding requirements on the CA wouldn't be effective since CAs cannot verify or audit how Applicants manage their internal account systems. That said, I do think recommendations on how Applicants and Cloud Providers should reason about confused deputy concerns would make sense in: https://github.com/ietf-wg-dnsop/draft-ietf-dnsop-domain-verification-techniques

docs/BR.md
```diff
@@ -3947,6 +3993,10 @@ The DNS TXT record MUST be placed on the "`_validation-contactemail`" subdomain
 
 The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain of the domain being validated. The entire RDATA value of this TXT record MUST be a valid Global Number as defined in RFC 3966, Section 5.1.4, or it cannot be used.
 
+### A.2.3. DNS TXT Record Persistent DCV Domain Label
```
ChristopherRC: Administrative comment and sorry for missing this earlier! Considering Appendix A is qualified with the description “CAA Contact Tag”, it’s not entirely clear why the "A.2.3. DNS TXT Record Persistent DCV Domain Label” content was added here beyond possibly following the pre-existing pattern used by the other A.2 entries (which also seems  kind of _awkward_ given the Appendix description). 

Should the naming of Appendix A change, or is there a more appropriate location for the A.2 text (including the newly proposed A.2.3. section)?

docs/BR.md
```diff
@@ -3947,6 +3993,10 @@ The DNS TXT record MUST be placed on the "`_validation-contactemail`" subdomain
...
+
+The DNS TXT record MUST be placed at the "_validation-persist" label prepended to the Authorization Domain Name being validated.
```
ChristopherRC:
```suggestion
The DNS TXT record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated.
```

nit for formatting consistency

docs/BR.md
```diff
@@ -1029,6 +1033,48 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
+##### 3.2.2.4.22 DNS TXT Record with Persistent Value
+
+Confirming the Applicant's control over a FQDN by confirming the presence of a persistent record identifying the Applicant in a DNS TXT record for an Authorization Domain Name that is prefixed with one DNS TXT Record Persistent DCV Domain Label.
```
slghtr-says:
```suggestion
Confirming the Applicant's control over a FQDN by confirming the presence of a persistent record identifying the Applicant in a DNS TXT record placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated.
```

@ChristopherRC - I want to be careful here to make sure the intent remains clear but what do you think about adding the label directly in the method and removing both the definition and appendix modifications?

docs/BR.md
slghtr-says: Great point. The line after ``APPENDIX A – CAA Contact Tag`` is ``These methods allow domain owners to publish **contact information** in DNS for the purpose of validating domain control.`` which scopes the section to contact based DCV methods. It would get tricky to try to expand the scope of this section. 

That said, I can't really think of a good reason to really maintain the definition and appendix entry either. I proposed an alternative approach in this comment: https://github.com/cabforum/servercert/pull/608/files#r2323549706

docs/BR.md
```diff
@@ -1029,6 +1033,48 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
...
+
+The CA MUST confirm the presence of a TXT record whose RDATA value fulfills the following requirements:
+
+1. The RDATA value MUST conform to the `issue-value` syntax as defined in RFC 8659, section 4.2; and
```
ryancdickson: nit (for self consistency in these new requirements)

```suggestion
1. The RDATA value MUST conform to the `issue-value` syntax as defined in RFC 8659, Section 4.2; and
```

docs/BR.md
ChristopherRC: Alternatively, what do you think about just pulling in the one line from the Appendix and allowing line 1038 to persist as is? Doing so might avoid unintentional changes in intent and could increase readability. 

```suggestion
Confirming the Applicant's control over a FQDN by confirming the presence of a persistent record identifying the Applicant in a DNS TXT record for an Authorization Domain Name that is prefixed with one DNS TXT Record Persistent DCV Domain Label.

The DNS TXT record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated.
```

docs/BR.md
```diff
@@ -1029,6 +1033,48 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
...
+2. The `issuer-domain-name` value MUST be an Issuer Domain Name disclosed by the CA in Section 4.2 of the CA's Certificate Policy and/or Certification Practices Statement; and
+3. The `issue-value` MUST contain an `accounturi` parameter, where the parameter value is a unique URI (as described by RFC 8657, Section 3) identifying the account of the Applicant which requested validation for this FQDN; and
+4. The `issue-value` MAY contain a `persistUntil` parameter. If present, the parameter value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds).
```
ChristopherRC:
```suggestion
4. The `issue-value` MAY contain a `persistUntil` parameter. If present, the parameter value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds); and
```

docs/BR.md
```diff
@@ -1029,6 +1033,48 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
...
+5. The `issue-value` MAY contain additional parameters. CAs MUST ignore any unknown parameter keys.
+
+If the `persistUntil` parameter is present, the CA MUST evaluate its value. If the time of the check is after the time specified in the `persistUntil` parameter value, the CA MUST NOT use the record as evidence of the Applicant's control over the FQDN.
+For example, the DNS TXT record might look like:
+_validation-persist.example.com IN TXT "authority.example; accounturi=https://authority.example/acct/123; persistUntil=1782424856"
+
+CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information as the Primary Network Perspective.
```
slghtr-says: https://github.com/slghtr-says/servercert/pull/3/files#diff-e0ac1bd190515a4f2ec09139d395ef6a8c7e9e5b612957c1f5a2dea80c6a6cfeR1173

@ChristopherRC / @ryancdickson  - Found a gap in 3.2.2.9 (MPIC) as well. 

Current: 
```
* a. the presence of the expected 1) Random Value, 2) Request Token, 3) IP Address, or 4) Contact Address, as required by the relied upon validation method specified in Sections 3.2.2.4 and 3.2.2.5; and
```

Suggested Change: 
```
* a. the presence of the expected challenge information, as required by the relied upon validation method specified in Sections 3.2.2.4 and 3.2.2.5; and
```

docs/BR.md
slghtr-says: While I like the direction, the challenge is what the do with the DNS TXT Record Persistent DCV Domain Label definition. Currently it is a defined term that is essentially a pointer back to the awkward Appendix section. So we'd either have to change the definition of the term or make it an undefined one.

docs/BR.md
ryancdickson: FWIW, my preference would be to leave this as it is currently written.

While the existing format is somewhat long-winded, I feel it most specifically describes what MPIC is evaluating. The change, as proposed, implies that an email address used to receive a challenge _is_ a challenge, which I disagree with. 

Can you share more on why you think this is a gap? Perhaps I'm misunderstanding the motivation.

slghtr-says: The persistent DNS record required for 3.2.2.4.22 is not one of: 1) Random Value, 2) Request Token, 3) IP Address, or 4) Contact Address

I hear you about it potentially being overly broad. The intent was to push the specific "thing" being verified to exist is delegated to the method itself. 

Curious about your thoughts

docs/BR.md
slghtr-says: Alternative suggestion: 

```suggestion
Confirming the Applicant's control over a FQDN by confirming the presence of a persistent record identifying the Applicant in a DNS TXT record for an Authorization Domain Name that is prepended with one Domain Label named "_validation-persist".
```

Rationale: 

- Removes the need for a new definition 
- Removes the reference to the Appendix (and the normative requirement that was buried there). 
- Better aligns with the structure of the original sentence.

docs/BR.md
ryancdickson: [I'll address this point in a soon to be posted comment!]

docs/BR.md
ryancdickson: @slghtr-says ... what if we...

- Removed the definition of "DNS TXT Record Persistent DCV Domain Label" in favor of a more generic "Persistent DCV TXT Record." Doing so allows us to avoid defining label requirements in the definition (defining requirements in definitions is not considered good practice). It also allows us to anchor on this phrase in the MPIC section (rather than a generic and possibly confusing approach, as we discuss in a lower comment).
- Remove the proposed appendix by clearly specifying requirements in the corresponding .22 requirements. This helps address the "this new sub-appendix doesn't make sense given the existing title of its parent"

This type of approach is imagined at https://github.com/cabforum/servercert/compare/main...ryancdickson:staging:static-dcv-tweak?expand=1

Could something like that work?

docs/BR.md
CBonnell: I like this approach. It consolidates three different sources of text into one, so it's a lot easier to read.

docs/BR.md
slghtr-says: I like it! Especially that this language clearly expresses the normative requirement for the format of the label in the method language itself. It resolves my concern from the MPIC section and improves overall readability of the method.

The definition is a tad circular

slghtr-says: Thank you! I think that approach does a some things really well.

1. I like that it adds a normative requirement for the Domain Label format directly in the method itself. 
2. It removes the awkward appendix. 
3. It shifts the approach from defining the Label to defining the Record which is helpful for language in other areas like the MPIC section. Without that additional definition I couldn't find a clean way to add Method 22 support without it getting too wordy or overly broad. 

I think the definition of the Record is a tad circular but I'm not convinced it makes the meaning less clear. 

I will go ahead and draft up a v2 ballot based all of the feedback provided.

ryancdickson: Other feedback / alternative suggestions also welcome!

docs/BR.md
```diff
@@ -435,6 +436,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **Pending Prohibition​​**: The use of a behavior described with this label is highly discouraged, as it is planned to be deprecated and will likely be designated as MUST NOT in the future.
 
+**Persistent DCV TXT Record:** A DNS TXT record identifying an Applicant in accordance with Section 3.2.2.4.22.
```
slghtr-says:
```suggestion
**Persistent DCV TXT Record:** A DNS TXT record identifying an Applicant in accordance with [Section 3.2.2.4.22](#322422-dns-txt-record-with-persistent-value).
```

docs/BR.md
```diff
@@ -1030,6 +1035,49 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
...
+4. The `issue-value` MAY contain a `persistUntil` parameter. If present, the parameter value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds); and
+5. The `issue-value` MAY contain additional parameters. CAs MUST ignore any unknown parameter keys.
+
+If the `persistUntil` parameter is present, the CA MUST evaluate its value. If the time of the check is after the time specified in the `persistUntil` parameter value, the CA MUST NOT use the record as evidence of the Applicant's control over the FQDN.
+
+For example, the Persistent DCV TXT Record might look like:
+`_validation-persist.example.com IN TXT "authority.example; accounturi=https://authority.example/acct/123; persistUntil=1782424856"`
+
+CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e., Persistent DCV TXT Record) as the Primary Network Perspective.
```
birgelee: MPIC implementations I have worked on (Let's Encrypt MPIC and Open MPIC) usually do not enforce record consistency between different perspectives. This is the language used in the other methods (e.g., the random token must be in the webpage but the two web pages retrieved by different perspectives do not need to be byte-for-byte identical).

Hypothetically, if perspective 1 sees:

Persistent DCV TXT Record: amazon.com; accounturi=abcd; persistUntil=<timestamp in the future>

and perspective 2 sees:

Persistent DCV TXT Record: amazon.com; accounturi=abcd

If both of these properly authorize issuance for the provided certificate request, I am of the opinion that issuance should proceed. The rationale for this is that if the victim controls both records, blocking issuance is a false positive. If one record is controlled by the adversary, the adversary could trivially put in the same value that is in the victim's true record because it still permits issuance. If both records are controlled by the adversary, the adversary can ensure they are consistent.

There is also a somewhat unfortunate failure mode that if a benign domain wants to add persistuUntil to their record, during the period of DNS propagation delay, they will not be able to get a cert even though both the old and new records permit issuance. This is a fairly counter intuitive behavior as it represents a system moving from one permitted state to another yet during the transition time, certificate issuance is blocked.

In MPIC implementations I have worked on we sort of have an "each perspective makes their own validation determination" approach. For challenges like this, the perspective individually decides if the provided record is capable of satisfying DCV and then their True False results are merged using the quorum policy. (there is some byte for byte identical language when say looking up a contact email, but that is a different situation).

On the MPIC language, one possibility is:

"To count as corroborating, a Network Perspective MUST observe challenge information (i.e., a Persistent DCV TXT Record) that permits issuance."

This is more similar to the case of CAA records where they don't need to be byte for byte identical.

ryancdickson: Thanks, Henry, this is a helpful clarification, and the operational rationale makes sense. The failure mode you described during a benign update (like adding persistUntil) due to DNS propagation delay is a compelling reason to avoid a strict byte-for-byte consistency requirement.

When I proposed the language that's in the current draft, my mental model was treating the entire 'Persistent DCV TXT Record' as a single value comparable to a random token in 3.2.2.4.7. 

To ensure we are explicitly capturing this intent (that authorization components must match, but optional metadata doesn't have to), what if we made your proposal slightly more specific? This might help avoid inadvertent misinterpretations since the rules here aren't as necessarily as crisp as they are for CAA (where circumstances that permit and block issuance are well-defined). 

**From:** "To count as corroborating, a Network Perspective MUST observe challenge information (i.e., a Persistent DCV TXT Record) that permits issuance."

**To something like:** "To count as corroborating, a Network Perspective MUST observe a Persistent DCV TXT Record that contains the same `issuer-domain-name` and `accounturi` parameters as the Primary Network Perspective."

Very much open to feedback on the above.

slghtr-says:
```suggestion
CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe a Persistent DCV TXT Record that permits issuance and contains the same issuer-domain-name and accounturi parameters as the Primary Network Perspective. 
```

What if we combined the suggested approaches? Require that 1) ``issuer-domain-name`` and ``accounturi`` parameters must match but also add 2) a clause that requires the entirety of the record as seen by the corroborating perspective (including the persistUntil) to be permissive of issuance. 

Thoughts?

romanf: I support Henry's position. I prefer the remote perspectives to be able to determine if what they see allows issuance. This is also how our MPIC implementation works. If we change this behaviour, our supplier would need to re-write the whole MPIC code. Which is of course a weak argument, but still an argument. ;)

birgelee: I like @slghtr-says 's suggestion. There is a minor consideration with the issuer-domain-name/accounturi match. While it does capture the intent of the requirement (the values seen at different perspectives match), there is still a potential false positive failure mode and it doesn't fully alleviate the engineering consideration Roman brought up.

False positive: Imagine a customer is moving from Entrust infrastructure to Sectigo given the acquisition of Entrust's public CA business by Sectigo (I mean no implications to either of these CAs, it was just an example of a recent CA operations acquisition of which there have been many). Sectigo in their CPS has a clause permitting issuance on "entrust.com" CAA records. This would presumably also be a valid issuer-domain-name for DNS persist validation. However, customers may want to move to the "sectigo.com" issuer-domain-name. Once again if a customer updates their DNS persist record, even though "sectigo.com" and "entrust.com" are valid issuer-domain-names for Sectigo issuance, there will be a false positive during DNS propagation delay.

Engineering: I do agree with Roman's point that every implementation I have seen determines perspective pass/fail in either an isolated concurrent subroutine or at remote perspectives themselves. The open MPIC implementation we would probably do to implement the language @slghtr-says proposed is to return the pass/fail boolean along with the issuer-domain-name and accounturi seen at each perspective and the coordinator would perform a consistency check on issuer-domain-name and accounturi. First, this is still better than a consistency check on all metadata (so I do like this wording), but it still means determining the validity of a record at perspective X requires knowledge of the record as it was seen by perspective Y. Neither CAA checking or the processing of other validation methods requires this.

The sort of 101-level explanation of MPIC we often give is as a consistency check (do remote perspectives see the same thing as the primary perspective). An alternative systems-level view of MPIC is to think of mini CAs spread across the Internet and seeing if those other CAs also think a given subscriber has passed DCV. The later view provides more robustness in the face of complex Internet infrastructure (e.g., complete consistency may require remote perspectives see the same DNS A records but this would incur a high false positive rate due to geo load balancing). CAB/F MPIC leans towards the latter approach (although there are some consistency check statements for some information lookup methods) and I think the more the independent determination approach is followed, the more robust the system is.

If there is a wording to not require any consistency and stipulate independent determination that each record seen passes the check requirements, I think this would be preferable. Regardless, @slghtr-says latest suggestion is still an improvement over the original wording.

slghtr-says: Thank you @birgelee and @romanf for the additional context and perspective.

### Issuer Domain Name Enforcement

Regarding the enforcement of the same `issuer-domain-name` between primary and corroborating perspectives, I'm not convinced this is worth the trade-off. 

Each CA has one or more defined values that are considered correct. As part of validation, the validation of the `issuer-domain-name` must be performed against that set and exact match one of them. 

**Implementation approach:** I would expect this to be similar to the set of [`caa_domains`](https://github.com/open-mpic/open-mpic-specification/blob/main/openapi.yaml#L173C9-L179) in OpenMPIC that is provided to each perspective as part of the CAA evaluation. I see no reason this approach has to be different.

Enforcement of this behavior would be covered by the overarching requirement that `"CA MUST observe a Persistent DCV TXT Record that [permits issuance/demonstrates the Applicant's control over the domain],"` so I see limited additional value in enforcing the sameness here.

### Account URI Enforcement

Regarding the enforcement of the same `accounturi` between primary and corroborating perspectives requirement, I think there is still value there.

In practice, I would expect there to be a single, specific `accounturi` that the validation is being performed for. Thus I'm not sure that perspective X requires knowledge of the record as it was seen by perspective Y for this to work.

**Implementation approach:** I would expect that value to be known prior to the start of the DCV action. Similar to Method 7 where the `token` and the `authorization-prefix` are provided to the corroborating perspectives, the `accounturi` could also be provided as part of the MPIC async action and independently verified by each perspective.

I fully admit I may be missing a solid reason for performing the "sameness" evaluation after the corroborating perspective evaluation.

### Recommended Language

Finally, given that this language is related to domain validation rather than issuance, I think the phrasing ``demonstrates the Applicant's control over the domain`` is clearer than ``permits issuance``. 

So my adjusted language recommendation would be: 
```suggestion
CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe a Persistent DCV TXT Record that demonstrates the Applicant's control over the domain and contains the same ``accounturi`` parameter as the Primary Network Perspective.
```

This language should alleviate the engineering concerns expressed by @romanf and @birgelee as the implementation of this method should be functionally similar to existing mechanisms in DCV and CAA lookups. This also adds more specificity about what needs to be the same.

birgelee: I like this language. It addresses the points I had raised. I also agree that it could be implemented using the existing mechanisms @slghtr-says discussed in open MPIC.

romanf: Perfect!

isnotnick:
I share the earlier concerns about the 'confused deputy' issue and I believe there's a missing piece in that many people may not realise - millions of certificates are issued through third parties that are *not* like 'cloud providers' and do not terminate TLS or handle private keys. They do not have domain control or delegation. They are simply passing a CSR through to a CA (sometimes through other third parties, multiple levels of 'reseller') to the CA, and returning the signed certificate.
The ultimate subscribers have no direct contact with the CA and as such no accountURI or other identifier.
These third-parties are certificate resellers, MSPs, MSSPs and more. They resell certificates - often offering preferential pricing or specific payment methods, support, tools, and ancillary services. They are middlemen, albeit adding value outside of the certificate itself.
There are resellers-of-resellers-of-resellers. There are SaaS/cloud 'certificate management' providers who again, do not terminate TLS or hold private keys, but could be given persistent ability to cause direct issuance over any subscriber that is or has been a customer.

There are also historic examples of where compromises of these 'resellers' (who, unrelatedly, may have been using poor practices with key generation) can impact tens of thousands of disparate subscribers.

The problem of course becomes one of account compromise where one reseller has direct-issuance capability for potentially millions of subscribers, many unaware of the risk. Account compromise shouldn't be considered limited to a single subscriber.

The difficulty of course is fixing this problem when the subscriber has no CA accountURI, and may be encouraged or required to use a 'shared' accountURI of a third-party, often unaware of the risks.

ACME generally requires a direct relationship and account with the CA, so I'd be in favour of Aaron's earlier suggestion of using some kind of ACME identifier in accountURI.
I don't know that CAs should be themselves responsible for self-indentifying 'resellers' or 'MSPs', so I think before we put this to ballot we need to ensure it's scoped to the correct use-cases, even if those are not applicable to all CAs.


(It could also be that the community and specifically the trust-store operators don't consider this much of a risk, and we should continue - but I do think that it's significant, and have brought this up a number of times before. I had hoped that the eventual move to 10 day DCV reuse period, effectively forcing many to do away with DCV reuse and do a challenge-per-issuance, would mitigate this - but this static challenge seems to exacerbate the problem).

docs/BR.md
```diff
@@ -1030,6 +1035,49 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
 
 **Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
+##### 3.2.2.4.22 DNS TXT Record with Persistent Value
+
+Confirming the Applicant's control over a FQDN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated (i.e., "`_validation-persist.[Authorization Domain Name]`").
```
slghtr-says: In response to: https://groups.google.com/a/groups.cabforum.org/g/servercert-wg/c/QWUbL8VRx2o/m/DdMVzgJ9AAAJ

There are actually two CNAME related considerations/scenarios at play here. 1) ADN determination and 2) following `_validation-persist` CNAME records during the DCV lookup.    

### Scenario 1: Authorization Domain Name Determination

```
certificate request for: foo.example.com
foo.example.com CNAME → bar.example.org
CA uses bar.example.org as Authorization Domain Name
CA looks for: _validation-persist.bar.example.org
```

**Suggestion:** 
Add ``CNAME records MUST NOT be followed when determining the Authorization Domain Name.``

```suggestion
Confirming the Applicant's control over a FQDN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "_validation-persist" label prepended to the Authorization Domain Name being validated(i.e., "`_validation-persist.[Authorization Domain Name]`"). CNAME records MUST NOT be followed when determining the Authorization Domain Name.
```

**Rationale:** 
1. Authorization Domain Name determination with CNAME following is complex and difficult to reason about for this method.
2. I can't think of a reason why _validation-persist.bar.example.org should represent authorization for foo.example.com simply because of a CNAME at domain level.

I ultimately do not believe that the complexity that adds is worth it for this method, but I am open to feedback and suggestions!

docs/BR.md
```diff
@@ -1,11 +1,11 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 
-subtitle: Version 2.1.7
+subtitle: Version 2.1.8
```
dzacharo:
```suggestion
subtitle: Version 2.1.9
```

docs/BR.md
```diff
@@ -1,11 +1,11 @@
...
 author:
   - CA/Browser Forum
 
-date: 25-August-2025
+date: DD-MONTH-2025
```
dzacharo:
```suggestion
date: 10-November-2025
```

docs/BR.md
```diff
@@ -151,6 +151,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.1.5       | SC81        | Introduce Schedule of Reducing Validity and Data Reuse Periods                         | 11-Apr-2025 | 16-May-2025                       |
 | 2.1.6       | SC85        | Require Validation of DNSSEC (when present) for CAA and DCV Lookups                    | 19-Jun-2025 | 21-Jul-2025                       |
 | 2.1.7       | SC089       | Mass Revocation Planning                                                               | 23-Jul-2025 | 25-Aug-2025                       |
+| 2.1.8       | SC088       | DNS TXT Record with Persistent Value DCV Method                                        | DD-MON-2025 | DD-MON-2025                       |
```
dzacharo:
```suggestion
| 2.1.9       | SC088       | DNS TXT Record with Persistent Value DCV Method                                        | 09-Oct-2025 | 10-Nov-2025                       |
```

docs/BR.md
```diff
@@ -151,6 +151,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.1.5       | SC81        | Introduce Schedule of Reducing Validity and Data Reuse Periods                         | 11-Apr-2025 | 16-May-2025                       |
 | 2.1.6       | SC85        | Require Validation of DNSSEC (when present) for CAA and DCV Lookups                    | 19-Jun-2025 | 21-Jul-2025                       |
 | 2.1.7       | SC089       | Mass Revocation Planning                                                               | 23-Jul-2025 | 25-Aug-2025                       |
+| 2.1.9       | SC088       | DNS TXT Record with Persistent Value DCV Method                                        | 09-Oct-2025 | 10-Nov-2025                       |
```
slghtr-says, line 155:
```suggestion
| 2.1.8       | SC092       | Sunset Precertificate Signing CAs                                                      | 03-Oct-2025 | 04-Nov-2025                       |
| 2.1.9       | SC088       | DNS TXT Record with Persistent Value DCV Method                                        | 09-Oct-2025 | 10-Nov-2025                       |
```

PR #619, opened 2025-09-29
SC-XXX: Cleanup for ADN CNAME use

richardwsmith:
Removes CNAME language from ADN definition and clarifies the use of FQDNs returned from DNS CNAME lookups to determine the ADN by method in section 3.2.2.4.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -750,7 +750,7 @@ The CA SHOULD implement a process to screen proxy servers in order to prevent re
 
 This section defines the permitted processes and procedures for validating the Applicant's ownership or control of the domain.
 
-The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qualified Domain Name (FQDN) listed in the Certificate as follows:
+The CA SHALL confirm that prior to issuance, the CA has validated each FQDN or Wildcard Domain Name listed in the Certificate as follows:
 
 1. When the FQDN is not an Onion Domain Name, the CA SHALL validate the FQDN using at least one of the methods listed below; and
 2. When the FQDN is an Onion Domain Name, the CA SHALL validate the FQDN in accordance with Appendix B.


docs/BR.md
```diff
@@ -323,7 +323,7 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
...
+**Authorization Domain Name**: The FQDN used to obtain authorization for a given FQDN to be included in a Certificate. If a Wildcard Domain Name is to be included in a Certificate, then the CA MUST remove "`*.`" from the left-most portion of the Wildcard Domain Name to yield the corresponding FQDN. The CA may prune zero or more Domain Labels of the FQDN from left to right until encountering a Base Domain Name and may use any one of the values that were yielded by pruning (including the Base Domain Name itself) for the purpose of domain validation.
```
aarongable: I think we should go a step further, in line with the idea that Definitions should merely be _definitions_, and not contain requirements. In that vein, I would propose:


```suggestion
**Authorization Domain Name**: The FQDN used to obtain authorization for a given FQDN to be included in a Certificate. See Section 3.2.2.4 and its subsections for descriptions of how an ADN may be derived from an applied-for FQDN.
```

See my comment below for the language I'd add to Section 3.2.2.4.

docs/BR.md
aarongable, line 757: In concert with my proposed definition change above, I would add something like the following language here:

```suggestion

The CA MAY apply any of the below transformations any number of times, in any order, to derive an Authorization Domain Name from the applied-for FQDN. Some validation methods restrict which transformations are valid to yield an ADN for use in that validation method.

1. Removing "`*.`" from the left-most portion of the FQDN. If the applied-for FQDN is a Wildcard Domain Name, then the CA MUST perform this transformation first.
2. Pruning the leftmost Domain Label of the FQDN. If the FQDN is a Base Domain Name, then the CA MUST NOT perform this transformation.
3. Replacing the FQDN with the FQDN returned from a DNS CNAME lookup of the first FQDN. If the FQDN has already had transformation #2 applied, then the CA MUST NOT perform this transformation.
```

Then the methods below which do allow CNAME delegation could be left alone, and the methods which disallow CNAME delegation could simply say "The CA MUST NOT use transformation #3 to derive the ADN used for this method."

docs/BR.md
jsha: I support this change in general. Wording tweak: "applied-for FQDN" is not general enough. It should be "applied-for FQDN or Wildcard Domain Name" (a Wildcard Domain Name is not an FQDN).

However, for the definition let's skip that complexity and just say:

> See Section 3.2.2.4 and its subsections for descriptions of how an ADN may be derived.

We can put details in 3.2.2.4.

docs/BR.md
jsha: This should say:


```suggestion
The CA SHALL confirm that prior to issuance, the CA has validated each FQDN or Wildcard Domain Name listed in the Certificate as follows:
```

docs/BR.md
aarongable: I like that suggestion.

Unfortunately the "is a Wildcard Domain Name an FQDN" question is already very muddy in the BRs. Per the Definitions, it can't be. But per language already in 3.2.2.4, the term "FQDN" is widely used to refer to Wildcard Domain Names already.

docs/BR.md
jsha: It's confusing that the first sentence says "any number of times, in any order," but (1) can be performed only once and must be first. And (3) cannot be performed after (2). How about an algorithm style:


The CA MUST follow this process to choose an Authorization Domain Name "_ADN_" for validation of each applied-for FQDN or Wildcard Domain Name:

 1. Initialize _ADN_ to the applied-for FQDN or Wildcard Domain Name.
 2. Remove "`*.`" from the left-most portion of _ADN_, if present.
 3. Optionally, if allowed by the validation method, replace _ADN_ with the result of a DNS CNAME lookup of _ADN_.
 4. Initialize _BDN_ to the Base Domain Name of _ADN_.
 5. Optionally, prune _N_ Domain Labels from _ADN_, from left to right. The CA MUST choose a positive value of _N_ such that _BDN_ is either a suffix of or equal to the resulting ADN.

jsha (changes requested):

docs/BR.md
aarongable: I like that rewrite a lot.

richardwsmith: How about replacing "FQDN or Wildcard Domain Name," with "dNSName," which includes both?

aarongable: That makes me just the slightest bit nervous, because the Common Name is not in fact a `dNSName`, it's a RelativeDistinguishedName with OID id-at-commonName. But because the CN MUST be duplicated in the SANs, this should work anyway. It's just making that requirement load-bearing in a way it wasn't previously.

jsha: Also, if we did that, for consistency we would want to also update "3.2.2.5 Authentication for an IP Address" to refer to iPAddress. As you mention on the mailing list, it would be easy to wind up refactoring more than we set out to do. "FQDN or Wildcard Domain Name" is clumsy but it's the current state of the art in this document, so I think it's fine.

richardwsmith: I like the direction, but I think it may be adding a more strict ruleset than is there currently, which I'm not totally opposed to, but I also think it may be more strict than needed and unnecessarily burden Subscribers.

jsha: What's an example of a use case that this would prohibit that is currently allowed? Pruning labels first, and then following a CNAME and setting ADN to the result?

defacto64:
Thankyou @aarongable and @jsha for your comments, which lead me to believe that my question was not obvious and that some cleanup is really worth doing. Right now, based on current BR language, I am inclined to think that my deduction is correct at least in the non-ACME case, so if this is not in the "spirit" of the BRs it 'd better be clarified.

docs/BR.md
richardwsmith: Yes, that in particular is the one I was looking at.  There are definitely cases, particularly in the enterprise sphere where I can see a Subscriber wishing to get their entire domain verified in one go, because they know they'll be requesting additional certificates for other sub-domains and therefore setting the CNAME at the Base Domain level of the requested FQDN, and I don't think there a reason to not allow that.  What we need to remember as we discuss CNAME in relation to DCV is that the CNAME put in place may or may not be set for the wider purpose permanent DNS traffic.  The CNAME record we're following may well be in place purely for the purpose of following the rules in place for DCV, and the CNAME itself may be removed 5 minutes after the DCV has been completed, so looking at it from the perspective of how it might be used for permanent DNS traffic is the wrong outlook, IMO.

jsha: > setting the CNAME at the Base Domain level of the requested FQDN

This doesn't work, because the Base Domain needs NS records (it is always a zone cut), and NS records cannot coexist with a CNAME. Or as more commonly put, "no CNAMEs at the apex of a domain."

https://www.rfc-editor.org/rfc/rfc2181#section-10.1

> An alias name (label of a CNAME record) may, if DNSSEC is in use, have SIG, NXT, and KEY RRs, but may have **no other data**.

(emphasis mine)

jsha: By the way, the use case you describe (verifying an entire domain in one go) is easily satisfied by:

1. CA chooses ADN = Base Domain (no CNAME chasing)
2. Subscriber puts a Random Value or Request Token in a TXT record for the ADN.
3. CA performs "3.2.2.4.7 DNS Change" for the ADN.
4. For any subsequent requests, e.g. `some.subdomain.example.com`, the CA again chooses ADN = Base Domain, and sees that there is a previous validation it can rely on.

dzacharo: > > setting the CNAME at the Base Domain level of the requested FQDN
> 
> This doesn't work, because the Base Domain needs NS records (it is always a zone cut), and NS records cannot coexist with a CNAME. Or as more commonly put, "no CNAMEs at the apex of a domain."
> 
> https://www.rfc-editor.org/rfc/rfc2181#section-10.1
> 
> > An alias name (label of a CNAME record) may, if DNSSEC is in use, have SIG, NXT, and KEY RRs, but may have **no other data**.
> 
> (emphasis mine)

CNAMEs at the BDN are not allowed as you correctly explained but what about the case of [3.2.2.4.7](https://github.com/cabforum/servercert/blob/main/docs/BR.md#32247-dns-change) which allows the ADN to have a label starting with an underscore?

> 2. an Authorization Domain Name that is prefixed with a Domain Label that begins with an underscore character

If the BDN is `example.com` the domain owner could create a CNAME entry of `_foo.example.com` and point to `delegated.myhostingprovider.com`. Validating `delegated.myhostingprovider.com` with 3.2.2.4.7 would effectively validate the entire Domain Namespace of `example.com`.

Is this correct?

aarongable: Ugh, that's a mess.

How I think this *should* work is that `_foo.example.com` is not an ADN. It is "an ADN prefixed with a domain label that begins with an underscore"; i.e. the ADN is still `example.com`. Following a CNAME from `_foo.example.com` would be part of the validation process, just like following HTTP 3XX redirects is part of the Agreed-Upon Change To Website validation process, but it would not be part of deriving an ADN.

dzacharo: Right. The way 3.2.2.4.7 is written, indicates that the "ADN" and "ADN prefixed with...." are treated equally. I'm not sure if this was intended to be a "solution" to the restriction described by @jsha in https://github.com/cabforum/servercert/pull/619#discussion_r2474187540. 

My point is that a Domain Owner can redirect control of a BDN (`example.com`) to another Domain Owner using CNAME. The CA will follow the validation process starting with either and "ADN" or an "ADN prefixed with..." and follow aliases (CNAMEs) as needed.

jsha:
I went ahead and drafted a version with the changes I was suggesting here, and some more on top that seemed to be required for consistency: https://github.com/cabforum/servercert/pull/627

PR #624, opened 2025-10-16
Draft SC-XX: Add MLDSA-87

CBonnell:
This preliminary proposal outlines the changes needed to the TLS BR to facilitate the TLS PQC pilot program.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -2824,7 +2824,7 @@ Table: Permitted `policyQualifiers`
 
 ##### 7.1.2.7.11 Subscriber Certificate Key Usage
 
-The acceptable Key Usage values vary based on whether the Certificate's `subjectPublicKeyInfo` identifies an RSA public key, an ECC public key, or ML-DSA public key. CAs MUST ensure the Key Usage is appropriate for the Certificate Public Key.
+The acceptable Key Usage values vary based on whether the Certificate's `subjectPublicKeyInfo` identifies an RSA public key, an ECC public key, or an ML-DSA public key. CAs MUST ensure the Key Usage is appropriate for the Certificate Public Key.
 
 Table: Key Usage for RSA Public Keys
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -2012,6 +2012,11 @@ For ECDSA key pairs, the CA SHALL:
 
 * Ensure that the key represents a valid point on the NIST P-256, NIST P-384 or NIST P-521 elliptic curve.
 
+For ML-DSA key pairs, the CA SHALL:
+
+* Ensure the Key uses one of the following parameter sets:
+  * ML-DSA-87 (OID: 2.16.840.1.101.3.4.3.19).
+
 No other algorithms or key sizes are permitted.
 
 ### 6.1.6 Public key parameters generation and quality checking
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -2012,6 +2012,11 @@ For ECDSA key pairs, the CA SHALL:
 
 * Ensure that the key represents a valid point on the NIST P-256, NIST P-384 or NIST P-521 elliptic curve.
 
+For ML-DSA key pairs, the CA SHALL:
+
+* Ensure the Key uses one of the following parameter sets:
+  * ML-DSA-87 (OID: 2.16.840.1.101.3.4.3.19).
```
robstradling, line 2018: Microsoft's announcement at the CABForum F2F yesterday said "Mandatory support for ML-DSA-87".

ML-DSA-44 and ML-DSA-65 weren't mentioned at all.  I don't know if that implies that the PQC Pilot Program will (1) disallow these parameter sets or (2) permit (but not require) them to be supported.

If 2, then this PR should probably allow them too.

CBonnell: Thanks for the call-out on this. I discussed this with the MSFT reps a bit ago and they confirmed that the pilot is limited to MLDSA-87 only at this point. So, I believe this language can stay as-is.

docs/BR.md
```diff
@@ -3344,6 +3365,19 @@ When encoded, the `AlgorithmIdentifier` for ECDSA keys MUST be byte-for-byte ide
 * For P-384 keys, `301006072a8648ce3d020106052b81040022`.
 * For P-521 keys, `301006072a8648ce3d020106052b81040023`.
 
+
+##### 7.1.3.2.4 ML-DSA
```
robstradling: s/7.1.3.2.4/7.1.3.1.3/

docs/BR.md
```diff
@@ -3442,6 +3476,13 @@ If the signing key is P-384, the signature MUST use ECDSA with SHA-384. When enc
 
 If the signing key is P-521, the signature MUST use ECDSA with SHA-512. When encoded, the `AlgorithmIdentifier` MUST be byte-for-byte identical with the following hex-encoded bytes: `300a06082a8648ce3d040304`.
 
+##### 7.1.3.2.4 ML-DSA
+
+The CA SHALL use the appropriate signature algorithm and encoding based upon the signing key used.
+
+If the signing key is ML-DSA-87, the signature algorithm SHALL be id-ml-dsa-87 (OID: 2.16.840.1.101.3.4.3.19). When encoded, the `AlgorithmIdentifier` SHALL be byte-for-byte identical with the following hex-encoded bytes: `300b0609608648016503040313`.
+
```
robstradling, line 3484: Since we're expecting only one relying party platform to support ML-DSA-87 in serverAuth certificates in the short term, I suggest adding text (perhaps with a note about it being a temporary requirement/advice) to require or recommend that ML-DSA-87 private keys must/should only sign ML-DSA-87 certificates.

This would help to ensure no interop issues with existing publicly-trusted RSA and ECC WebPKI hierarchies, and (I'm speculating) might help to make this ballot more palatable to some Root Store voting members.

timfromdigicert: Yeah I like that idea. One algorithm keeps the pilot simple, and pure chains with no mixture with RSA and ECC makes it even simpler.

CBonnell: Thanks, this is a good idea. I updated the ballot with a SHALL NOT to prevent mixed algorithm hierarchies.

CBonnell: Argh, proposed language doesn't prevent RSA/ECC CAs from signing certs that certify ML-DSA keys. So we need a restriction in the SPKI algs section for MLDSA too.

CBonnell: Fix for the above pushed.

docs/BR.md
```diff
@@ -2817,7 +2824,7 @@ Table: Permitted `policyQualifiers`
 
 ##### 7.1.2.7.11 Subscriber Certificate Key Usage
 
-The acceptable Key Usage values vary based on whether the Certificate's `subjectPublicKeyInfo` identifies an RSA public key or an ECC public key. CAs MUST ensure the Key Usage is appropriate for the Certificate Public Key.
+The acceptable Key Usage values vary based on whether the Certificate's `subjectPublicKeyInfo` identifies an RSA public key, an ECC public key, or ML-DSA public key. CAs MUST ensure the Key Usage is appropriate for the Certificate Public Key.
```
robstradling: Nit: s/or ML-DSA/or an ML-DSA/
Alternatively, remove the "an" before "ECC public key".

CBonnell: Thanks, fixed.

docs/BR.md
aarongable: Leaving a comment here as welll, since it's not yet clear whether this PR or https://github.com/cabforum/servercert/pull/662 will be the one that moves forward.

Now that we're making this change due to general preparedness, and not just the Microsoft pilot program, we should allow ML-DSA-44 and ML-DSA-65.

PR #627, opened 2025-10-30
SC-101v2: Clarify Authorization Domain Names

jsha:
Simplify the definition of ADN and move the algorithm in 3.2.2.4. Choosing an ADN is a precursor to doing validation. Every validation operates on an ADN.

Simplify the language in each method around pruning labels and wildcard issuance. Add corresponding language about following CNAMEs when choosing the ADN, and around onion issuance.

Require record keeping when choosing the ADN.

Replace retired methods with language saying they are retired.

Define Base Domain Name as a function of any given FQDN.

Move the definition of Domain Contact into the only remaining method that uses it (3.2.2.4.12 Validating Applicant as a Domain Contact). Note that this method does not allow CNAME lookups when choosing the ADN, because it previously relied on Base Domain Name, which previously had the language "applied-for FQDN".

Use "ADN" consistently in method definitions instead of "FQDN" (which is insufficiently precise as to whether it's the applied-for FQDN or the ADN).

Clean up the language in the onion section around ADNs.

Here's a table of the methods and what ADN operations can happen on them according to this PR (plus onion issuance):

| Method                                         | Wildcard | Prune | CNAME | Onion |
| --------                                                    | - | - | - | - |
| 3.2.2.4.4 Constructed Email to Domain Contact               | ✔️ | ✔️ | ✔️ | - |
| 3.2.2.4.7 DNS Change                                        | ✔️ | ✔️ | ✔️ | - |
| 3.2.2.4.8 IP Address                                        | - | - | - | - |
| 3.2.2.4.12 Validating Applicant as a Domain Contact         | ✔️ | ✔️ | - | - |
| 3.2.2.4.13 Email to DNS CAA Contact                         | ✔️ | ✔️ | ✔️ | - |
| 3.2.2.4.14 Email to DNS TXT Contact                         | ✔️ | ✔️ | ✔️ | - |
| 3.2.2.4.16 Phone Contact with DNS TXT Record Phone Contact  | ✔️ | ✔️ | ✔️ | - |
| 3.2.2.4.17 Phone Contact with DNS CAA Phone Contact         | ✔️ | ✔️ | ✔️ | - |
| 3.2.2.4.18 Agreed-Upon Change to Website v2                 | - | - | - | ✔️ |
| 3.2.2.4.19 Agreed-Upon Change to Website - ACME             | - | - | - | ✔️ |
| 3.2.2.4.20 TLS Using ALPN                                   | - | - | - | ✔️ |
| 3.2.2.4.21 DNS Labeled with Account ID - ACME               | ✔️ | ✔️ | - | - |
| 3.2.2.4.22 DNS TXT Record with Persistent Value             | ✔️ | ✔️ | - | - |
| Appendix B.2.b                                              | ✔️ | ✔️ | - | ✔️ |

This is a more in-depth replacement for #619.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -754,7 +754,7 @@ The CA MUST follow this process when choosing the Authorization Domain Name "ADN
 2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method that allows wildcard issuance. If `A` is an Onion Domain Name, the CA MUST choose a validation method that allows Onion Domain Name issuance.
 3. If `A` is an FQDN, choose at most one:
   a. If the validation method allows CNAME lookups when choosing the ADN, replace `A` with the result of a DNS CNAME lookup of `A`.
-  c. If the validation method allows pruning domain labels when choosing the ADN, prune zero or more Domain Labels of `A` from left to right until encountering the Base Domain Name of `A` or the CA chooses to stop pruning, whichever comes first.
+  b. If the validation method allows pruning domain labels when choosing the ADN, prune zero or more Domain Labels of `A` from left to right until encountering the Base Domain Name of `A` or the CA chooses to stop pruning, whichever comes first.
 4. If `A` is a Wildcard Domain Name:
   a. Remove "\*." from the left-most portion of `A`.
   b. If the validation method allows pruning domain labels when choosing the ADN, prune zero or more Domain Labels of `A` from left to right until encountering the Base Domain Name of `A` or the CA chooses to stop pruning, whichever comes first.
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -851,8 +851,6 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
 
 If the CA or an Affiliate of the CA operates a DNS zone to which Applicants can delegate (via CNAME) their underscore-prefixed Domain Label, the CA MUST ensure that each Applicant delegates to a unique FQDN within that zone. A CA or Affiliate of a CA SHOULD NOT operate such a service, and SHOULD direct any Applicants using such a service to use the method described in [Section 3.2.2.4.22](#322422-dns-txt-record-with-persistent-value) instead.
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
-
 ##### 3.2.2.4.8 IP Address
 
 Confirming the Applicant's control over the ADN by confirming that the Applicant controls an IP address returned from a DNS lookup for A or AAAA records for the ADN in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
@@ -997,9 +995,7 @@ CAs performing validations using this method MUST implement Multi-Perspective Is
 
 ##### 3.2.2.4.22 DNS TXT Record with Persistent Value
 
-Confirming the Applicant's control over a FQDN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated (i.e., "`_validation-persist.[Authorization Domain Name]`"). For this method, the CA MUST NOT use the FQDN returned from a DNS CNAME lookup as the FQDN for the purposes of domain validation. This prohibition overrides the Authorization Domain Name definition. CNAME records MAY be followed when resolving the Persistent DCV TXT Record.
-
-The CA MUST confirm the Persistent DCV TXT Record’s RDATA value fulfills the following requirements:
+Confirming the Applicant's control over the ADN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated (i.e., "`_validation-persist.[Authorization Domain Name]`"). This prohibition overrides the Authorization Domain Name definition.
 
 1. The RDATA value MUST conform to the `issue-value` syntax as defined in RFC 8659, Section 4.2; and
 2. The `issuer-domain-name` value MUST be an Issuer Domain Name disclosed by the CA in Section 4.2 of the CA's Certificate Policy and/or Certification Practices Statement; and
@@ -1025,8 +1021,6 @@ Table: Examples of how the `persistUntil` parameter affects validation
 | 2025-06-15T12:00:00Z | (not present) | Yes | No persistUntil parameter present, so no time restriction applies |
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe a Persistent DCV TXT Record that demonstrates the Applicant's control over the domain and contains the same `accounturi` parameter as the Primary Network Perspective.
-
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
 
 #### 3.2.2.5 Authentication for an IP Address
 
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -757,7 +757,7 @@ The CA MUST follow this process when choosing the Authorization Domain Name "ADN
 1. Initialize `A` to the applied-for FQDN or Wildcard Domain Name.
 2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method with a check in the Wildcard column below. If `A` is an Onion Domain Name, the CA MUST choose a validation method with a check in the Onion column below.
 3. If `A` is an FQDN:
-    1. Optionally, if the validation method has a check in the CNAME column below, replace `A` with the result of a DNS CNAME lookup of `A`.
+    1. Optionally, if the validation method has a check in the CNAME column below, replace `A` with the result of a DNS CNAME lookup of `A`. This step may be repeated.
     2. If the validation method has a check in the Prune column below, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
 4. If `A` is a Wildcard Domain Name:
     1. Remove "\*." from the left-most portion of `A`.
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -751,13 +751,30 @@ This section defines the permitted processes and procedures for validating the A
 The CA MUST follow this process when choosing the Authorization Domain Name "ADN" for validation of each applied-for FQDN or Wildcard Domain Name:
 
 1. Initialize `A` to the applied-for FQDN or Wildcard Domain Name.
-2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method that allows wildcard issuance. If `A` is an Onion Domain Name, the CA MUST choose a validation method that allows Onion Domain Name issuance.
+2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method with a check in the Wildcard column below. If `A` is an Onion Domain Name, the CA MUST choose a validation method with a check in the Onion column below.
 3. If `A` is an FQDN:
-    1. Optionally, if the validation method allows CNAME lookups when choosing the ADN, replace `A` with the result of a DNS CNAME lookup of `A`.
-    2. If the validation method allows pruning domain labels when choosing the ADN, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
+    1. Optionally, if the validation method has a check in the CNAME column below, replace `A` with the result of a DNS CNAME lookup of `A`.
+    2. If the validation method has a check in the Prune column below, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
 4. If `A` is a Wildcard Domain Name:
     1. Remove "\*." from the left-most portion of `A`.
-    2. If the validation method allows pruning domain labels when choosing the ADN, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
+    2. If the validation method has a check in the Prune column below, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
+5. Use `A` as the ADN.
+
+| Method                                         | Wildcard | Prune | CNAME | Onion |
+| --------                                                    | - | - | - | - |
+| 3.2.2.4.4 Constructed Email to Domain Contact               | ✔️ | ✔️ | ✔️ | - |
+| 3.2.2.4.7 DNS Change                                        | ✔️ | ✔️ | ✔️ | - |
+| 3.2.2.4.8 IP Address                                        | - | - | - | - |
+| 3.2.2.4.12 Validating Applicant as a Domain Contact         | ✔️ | ✔️ | - | - |
+| 3.2.2.4.13 Email to DNS CAA Contact                         | ✔️ | ✔️ | ✔️ | - |
+| 3.2.2.4.14 Email to DNS TXT Contact                         | ✔️ | ✔️ | ✔️ | - |
+| 3.2.2.4.16 Phone Contact with DNS TXT Record Phone Contact  | ✔️ | ✔️ | ✔️ | - |
+| 3.2.2.4.17 Phone Contact with DNS CAA Phone Contact         | ✔️ | ✔️ | ✔️ | - |
+| 3.2.2.4.18 Agreed-Upon Change to Website v2                 | - | - | - | ✔️ |
+| 3.2.2.4.19 Agreed-Upon Change to Website - ACME             | - | - | - | ✔️ |
+| 3.2.2.4.20 TLS Using ALPN                                   | - | - | - | ✔️ |
+| 3.2.2.4.21 DNS Labeled with Account ID - ACME               | ✔️ | ✔️ | - | - |
+| Appendix B.2.b                                              | ✔️ | ✔️ | - | ✔️ |
 
 When the ADN is an Onion Domain Name, the CA SHALL validate it in accordance with Appendix B.
 
@@ -806,12 +823,6 @@ The email MAY be re-sent in its entirety, including the re-use of the Random Val
 
 The Random Value SHALL remain valid for use in a confirming response for no more than 30 days from its creation. The CPS MAY specify a shorter validity period for Random Values.
 
-This method allows wildcard issuance.
-
-This method allows pruning domain labels when choosing the ADN.
-
-This method allows CNAME lookups when choosing the ADN.
-
 ##### 3.2.2.4.5 Domain Authorization Document
 
 This method has been retired and MUST NOT be used. Prior validations using this method and validation data gathered according to this method SHALL NOT be used to issue certificates.
@@ -833,23 +844,11 @@ If a Random Value is used, the CA SHALL provide a Random Value unique to the Cer
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same challenge information (i.e. Random Value or Request Token) as the Primary Network Perspective.
 
-This method allows wildcard issuance.
-
-This method allows pruning domain labels when choosing the ADN.
-
-This method does not allow CNAME lookups when choosing the ADN. Note: Domain name resolution includes processing of CNAMEs, so validation under this method naturally includes processing CNAMES, even when the DNS query type is TXT or CAA.
-
 ##### 3.2.2.4.8 IP Address
 
 Confirming the Applicant's control over the ADN by confirming that the Applicant controls an IP address returned from a DNS lookup for A or AAAA records for the ADN in accordance with [Section 3.2.2.5](#3225-authentication-for-an-ip-address).
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same IP address as the Primary Network Perspective.
-
-This method MUST NOT be used for wildcard issuance.
-
-This method does not allow pruning domain labels when choosing the ADN.
-
-This method does not allow CNAME lookups when choosing the ADN.
 
 ##### 3.2.2.4.9 Test Certificate
 
@@ -869,12 +868,6 @@ Confirming the Applicant's control over the ADN by validating the Applicant is t
 
 The Domain Contact for an ADN is: The registrant, technical contact, or administrative contact (or the equivalent under a ccTLD) as listed in the WHOIS record of the ADN or as obtained through direct contact with the Domain Name Registrar, or the holder of the email address in an SOA record for the ADN.
 
-This method allows wildcard issuance.
-
-This method allows pruning domain labels when choosing the ADN.
-
-This method does not allow CNAME lookups when choosing the ADN.
-
 Effective January 15, 2025:
 - When issuing Subscriber Certificates, the CA MUST NOT rely on Domain Contact information obtained using an HTTPS website, regardless of whether previously obtained information is within the allowed reuse period.
 - When obtaining Domain Contact information for the ADN the CA:
@@ -892,12 +885,6 @@ The Random Value SHALL be unique in each email. The email MAY be re-sent in its 
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same selected contact address used for domain validation as the Primary Network Perspective.
 
-This method allows wildcard issuance.
-
-This method allows pruning domain labels when choosing the ADN.
-
-This method allows CNAME lookups when choosing the ADN.
-
 ##### 3.2.2.4.14 Email to DNS TXT Contact
 
 Confirming the Applicant's control over the ADN by sending a Random Value via email and then receiving a confirming response utilizing the Random Value. The Random Value MUST be sent to a DNS TXT Record Email Contact for the Authorization Domain Name.
@@ -907,12 +894,6 @@ The same email MAY be sent to multiple recipients as long as all recipients are 
 The Random Value SHALL be unique in each email. The email MAY be re-sent in its entirety, including the re-use of the Random Value, provided that its entire contents and recipient(s) SHALL remain unchanged. The Random Value SHALL remain valid for use in a confirming response for no more than 30 days from its creation. The CPS MAY specify a shorter validity period for Random Values.
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same selected contact address used for domain validation as the Primary Network Perspective.
-
-This method allows wildcard issuance.
-
-This method allows pruning domain labels when choosing the ADN.
-
-This method allows CNAME lookups when choosing the ADN.
 
 ##### 3.2.2.4.15 Phone Contact with Domain Contact
 
@@ -930,12 +911,6 @@ The Random Value SHALL remain valid for use in a confirming response for no more
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same selected contact address used for domain validation as the Primary Network Perspective.
 
-This method allows wildcard issuance.
-
-This method allows pruning domain labels when choosing the ADN.
-
-This method allows CNAME lookups when choosing the ADN.
-
 ##### 3.2.2.4.17 Phone Contact with DNS CAA Phone Contact
 
 Confirm the Applicant's control over the ADN by calling the DNS CAA Phone Contact’s phone number and obtain a confirming response to validate the ADN. Each phone call MAY confirm control of multiple ADNs provided that the same DNS CAA Phone Contact phone number is listed for each ADN being verified and the recipient of the phone call provides a confirming response for each ADN. The relevant CAA Resource Record Set MUST be found using the search algorithm defined in RFC 8659 Section 3.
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -871,67 +863,62 @@ This method has been retired and MUST NOT be used.
...
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
+This method allows pruning domain labels when choosing the ADN.
+
+This method does not allow CNAME lookups when choosing the ADN.
 
 Effective January 15, 2025:
 - When issuing Subscriber Certificates, the CA MUST NOT rely on Domain Contact information obtained using an HTTPS website, regardless of whether previously obtained information is within the allowed reuse period.
-- When obtaining Domain Contact information for a requested Domain Name the CA:
+- When obtaining Domain Contact information for a requested ADN the CA:
```
aarongable: nit:
```suggestion
- When obtaining Domain Contact information for the ADN the CA:
```

docs/BR.md
```diff
@@ -871,67 +863,62 @@ This method has been retired and MUST NOT be used.
...
 
 ##### 3.2.2.4.13 Email to DNS CAA Contact
 
-Confirming the Applicant's control over the FQDN by sending a Random Value via email and then receiving a confirming response utilizing the Random Value. The Random Value MUST be sent to a DNS CAA Email Contact. The relevant CAA Resource Record Set MUST be found using the search algorithm defined in RFC 8659, Section 3.
+Confirming the Applicant's control over the ADN by sending a Random Value via email and then receiving a confirming response utilizing the Random Value. The Random Value MUST be sent to a DNS CAA Email Contact. The relevant CAA Resource Record Set MUST be found using the search algorithm defined in RFC 8659, Section 3.
 
-Each email MAY confirm control of multiple FQDNs, provided that each email address is a DNS CAA Email Contact for each Authorization Domain Name being validated. The same email MAY be sent to multiple recipients as long as all recipients are DNS CAA Email Contacts for each Authorization Domain Name being validated.
+The same email MAY be sent to multiple recipients as long as all recipients are DNS CAA Email Contacts for the Authorization Domain Name being validated.
```
aarongable: I'm not sure that the removal of

> Each email MAY confirm control of multiple FQDNs, provided that each email address is a DNS CAA Email Contact for each Authorization Domain Name being validated

is quite correct, if the goal is for this part of the diff to be a no-op. I think that sentence is saying that, if `blog.example.org` and `shop.example.com` have the exact same DNS CAA Email Contact, then you can validate both ADNs with a single email and single random value. So we probably want to instead leave the sentence in place, but edit it to say something like:

> Each email MAY confirm control of multiple ADNs, provided that each email address is a DNS CAA Email Contact for all ADNs being validated.

jsha: Ah, good point! I had been thinking of this as covering a case that is also covered by "choose the same ADN for those FQDNs" but of course that might not be possible; they might just both have the same email address in CAA.

docs/BR.md
```diff
@@ -750,14 +748,22 @@ The CA SHOULD implement a process to screen proxy servers in order to prevent re
...
 This section defines the permitted processes and procedures for validating the Applicant's ownership or control of the domain.
 
-The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qualified Domain Name (FQDN) listed in the Certificate as follows:
+The CA MUST follow this process when choosing the Authorization Domain Name "ADN" for validation of each applied-for FQDN or Wildcard Domain Name:
+
+1. Initialize `A` to the applied-for FQDN or Wildcard Domain Name.
+2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method that allows wildcard issuance. If `A` is an Onion Domain Name, the CA MUST choose a validation method that allows Onion Domain Name issuance.
+3. If `A` is an FQDN, choose at most one:
+  a. If the validation method allows CNAME lookups when choosing the ADN, replace `A` with the result of a DNS CNAME lookup of `A`.
+  c. If the validation method allows pruning domain labels when choosing the ADN, prune zero or more Domain Labels of `A` from left to right until encountering the Base Domain Name of `A` or the CA chooses to stop pruning, whichever comes first.
```
srdavidson: Is there a missing subsection 3(b)?

jsha: Nope, this was a drafting error. Fixed, and thanks for noticing.

docs/BR.md
```diff
@@ -323,11 +323,11 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
...
+**Authorization Domain Name**: The FQDN used to perform validation of domain authorization or control for a given FQDN or Wildcard Domain Name.
 
 **Authorized Ports**: One of the following ports: 80 (http), 443 (https), 25 (smtp), 22 (ssh).
 
-**Base Domain Name**: The portion of an applied-for FQDN that is the first Domain Name node left of a registry-controlled or public suffix plus the registry-controlled or public suffix (e.g. "example.co.uk" or "example.com"). For FQDNs where the right-most Domain Name node is a gTLD having ICANN Specification 13 in its registry agreement, the gTLD itself may be used as the Base Domain Name.
+**Base Domain Name**: The portion of a given FQDN that is the first Domain Name node left of a registry-controlled or public suffix plus the registry-controlled or public suffix (e.g. "example.co.uk" or "example.com"). For FQDNs the right-most Domain Name node is a gTLD having ICANN Specification 13 in its registry agreement, the gTLD itself may be used as the Base Domain Name.
```
cardonator: was the "where" removal here intentional?

jsha:
> First, it seems like the intersection of allowances with the algorithm for choosing an ADN is complex and leaves room for ambiguity. For example, perhaps under 4 it would be worth explicitly stating that following CNAMEs is not allowed for selecting ADNs on Wildcard Domains regardless of if the method allows it or not.

Good point. I'll have a think about the clearest way to express this. Ideally the validation methods would say something like "this has property `foo`", and the ADN selection algorithm would specify things that are allowable for validations with property `foo`; but the way I currently express that ("This method allows pruning domain labels when choosing the ADN.") could certainly be interpreted to convey a permission independent of what the algorithm says.

> Second, further down in the text, there are references to ADNs that aren't clear whether they should be referring to an ADN that was selected from following a CNAME or not. An example of what I mean follows.

My goal in writing this is that, once the ADN is selected, the validation method shouldn't have any dependency on how it was selected (i.e. whether a CNAME was followed).

> Under 3.2.2.4.7, we see this heading:
> Under the ADN selection algorithm provided, however, this ADN could be my domain, a base domain of my domain, or a CNAME that one of those domains pointed to. Is this meant to allow the following condition?

```
testing.payments.mydomain.com (no value)
payments.mydomain.com (pruned) CNAME customernumber.vendordomain.com
customernumber.vendordomain.com (no value)
_dns.customernumber.vendordomain.com TXT myrandomvalue
```

This example uses prune-then-CNAME, which IMO should not be allowed because it doesn't follow the actual semantics of control in DNS. The operator of `vendordomain.com` does not actually control `testing.payments.mydomain.com`, because the CNAME at the parent domain `payments.mydomain.com` does not actually impact name resolutions for `testing.payments.mydomain.com`.

Perhaps you meant to ask about a situation like this?

```
payments.example.com CNAME customernumber.vendordomain.com
customernumber.vendordomain.com (no value)
_dns.customernumber.vendordomain.com TXT myrandomvalue
```

In this example, the CA would be selecting `customernumber.vendordomain.com` as the ADN, and then prefixing the ADN with a label (`_dns`). Since PR specifically says that "3.2.2.4.7 DNS Change" does not allow CNAME lookups when choosing the ADN, that wouldn't be allowed.

The idea being that it's better for CNAME following to happen automatically as part of the normal DNS operations that occur during validation. So in this example, there could be `customernumber.vendordomain.com TXT myrandomvalue`, and a DNS lookup of `payments.example.com` with QTYPE TXT would find that record (because DNS resolution follows CNAMEs).

> Similarly, under 3.2.2.4.4 the ADN is referenced multiple times as the target of the constructed email domains, suggesting in the above example that the emails would/should be send to the target of the CNAME domain (but in this case I assume no pruning would be allowed, therefore the emails would necessarily be to addresses like admin@customernumber.vendordomain.com).

This is why CNAME-then-prune is convenient for subscribers. In my example above, if the validation method is "3.2.2.4.4 Constructed Email to Domain Contact," the CA would be allowed to first follow the single CNAME, resulting in `customernumber.vendordomain.com`, and subsequently prune a label to choose `vendordomain.com` as the ADN. The CA would then email, e.g. `admin@vendordomain.com`.

sd9067016-cmd:
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','GTM-NGX3TD7N');</script>

docs/BR.md
```diff
@@ -1013,29 +979,22 @@ If the CA follows redirects, the following apply:
...
 
 CAs performing validations using this method MUST implement Multi-Perspective Issuance Corroboration as specified in [Section 3.2.2.9](#3229-multi-perspective-issuance-corroboration). To count as corroborating, a Network Perspective MUST observe the same token as the Primary Network Perspective.
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
-
 ##### 3.2.2.4.22 DNS TXT Record with Persistent Value
 
 Confirming the Applicant's control over a FQDN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated (i.e., "`_validation-persist.[Authorization Domain Name]`"). For this method, the CA MUST NOT use the FQDN returned from a DNS CNAME lookup as the FQDN for the purposes of domain validation. This prohibition overrides the Authorization Domain Name definition. CNAME records MAY be followed when resolving the Persistent DCV TXT Record.
```
slghtr-says:
```suggestion
Confirming the Applicant's control over the ADN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the ADN being validated (i.e., "`_validation-persist.[Authorization Domain Name]`"). CNAME records MAY be followed when resolving the Persistent DCV TXT Record.
```

These suggested changes should bring method 22 in line with the updated ADN definition in this draft ballot. Essentially this is removing the references to the previous definition of ADN around following CNAMEs as part of ADN determination. That practice is now prohibited for this method per section 3.2.2.4 so the language here no longer makes sense. I decided to leave the clause around CNAME following during DNS resolution mainly to minimize changes but would be open to removing it as well.

docs/BR.md
```diff
@@ -754,14 +752,40 @@ The CA SHOULD implement a process to screen proxy servers in order to prevent re
...
-The CA SHALL confirm that prior to issuance, the CA has validated each Fully-Qualified Domain Name (FQDN) listed in the Certificate as follows:
-
-1. When the FQDN is not an Onion Domain Name, the CA SHALL validate the FQDN using at least one of the methods listed below; and
-2. When the FQDN is an Onion Domain Name, the CA SHALL validate the FQDN in accordance with Appendix B.
+The CA MUST follow this process when choosing the Authorization Domain Name "ADN" for validation of each applied-for FQDN or Wildcard Domain Name:
+
+1. Initialize `A` to the applied-for FQDN or Wildcard Domain Name.
+2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method with a check in the Wildcard column below. If `A` is an Onion Domain Name, the CA MUST choose a validation method with a check in the Onion column below.
+3. If `A` is an FQDN:
+    1. Optionally, if the validation method has a check in the CNAME column below, replace `A` with the result of a DNS CNAME lookup of `A`.
```
aarongable: From discussion in the 2026-01-08 call:

The Validation WG believes that we should rephrase this such that it explicitly allows following multiple CNAME delegations in a chain.

On the one hand, allowing CNAME chaining at this top level is very unusual: it requires the CA to implement a whole layer on top of DNS, since no DNS resolver (stub or recursive) will do this CNAME chaining for you. And it seems like it should be unnecessary, since the methods which allow following CNAMEs during ADN derivation (e.g. dns change, email to caa or txt contact) then also follow CNAMEs during actual validation.

On the other hand, disallowing CNAME chaining at this level may be a surprising change from the status quo, as some ecosystem participants believe that the current ADN definition is recursive and therefore allows CNAME chaining. More importantly, it disallows one specific scenario:
- A CDN wants to get a cert for `example.org`, which is CNAMEd to `example.com`, which is CNAMEd to `example.com.content.cdn.com`.
- They want to use a single central solver for ACME DNS-01 challenges, so they CNAME `_acme-challenge.example.com.content.cdn.com` to `acme.cdn.com`

If CNAME chaining in ADN construction were disallowed in this scenario, the CDN would either need their customer to CNAME `example.org` to the CDN directly (so that there's not a chain of CNAMEs prior to reaching the CDN), or they would need their customer to place a CNAME record on `_acme-challenge.example.com`. Both solutions place higher overhead on the customer/cdn relationship, and neither enforces desirable properties.

docs/BR.md
richardwsmith: Please add a compliance date for this ballot.  Given the change of target of "other FQDNs that end with all the Domain Labels of the validated FQDN," from the requested FQDN to the ADN, I propose a compliance date of 2026-09-15 to allow time for both full code review, and to allow affected customers to adapt.

docs/BR.md
```diff
@@ -783,53 +808,26 @@ This method has been retired and MUST NOT be used. Prior validations using this
...
 
-Each email MAY confirm control of multiple FQDNs, provided the Authorization Domain Name used in the email is an Authorization Domain Name for each FQDN being confirmed
-
 The Random Value SHALL be unique in each email.
 
 The email MAY be re-sent in its entirety, including the re-use of the Random Value, provided that its entire contents and recipient SHALL remain unchanged.
 
 The Random Value SHALL remain valid for use in a confirming response for no more than 30 days from its creation. The CPS MAY specify a shorter validity period for Random Values.
 
-**Note**: Once the FQDN has been validated using this method, the CA MAY also issue Certificates for other FQDNs that end with all the Domain Labels of the validated FQDN. This method is suitable for validating Wildcard Domain Names.
```
richardwsmith, line 801: Removal of this note and these "Note" statements in other places throughout this section potentially makes  very large changes to the section, particularly with respect to, "...other FQDNs that end with all the Domain Labels of the validated FQDN," as the additional text added above in 3.2.2.4 has only explicitly addressed the Wildcard case.  Additionally these "Note" sections have been inconsistently treated as many, but not all of them have been deleted.

docs/BR.md
```diff
@@ -754,14 +752,40 @@ The CA SHOULD implement a process to screen proxy servers in order to prevent re
...
+2. Choose a validation method. If `A` is a Wildcard Domain Name, the CA MUST choose a validation method with a check in the Wildcard column below. If `A` is an Onion Domain Name, the CA MUST choose a validation method with a check in the Onion column below.
+3. If `A` is an FQDN:
+    1. Optionally, if the validation method has a check in the CNAME column below, replace `A` with the result of a DNS CNAME lookup of `A`.
+    2. If the validation method has a check in the Prune column below, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
+4. If `A` is a Wildcard Domain Name:
+    1. Remove "\*." from the left-most portion of `A`.
+    2. If the validation method has a check in the Prune column below, prune zero or more Domain Labels of `A` from left to right until `A` is equal to the Base Domain Name of `A`, or the CA chooses to stop pruning, whichever comes first.
+5. Use `A` as the ADN.
+
+| Method                                         | Wildcard | Prune | CNAME | Onion |
```
richardwsmith: Text and table specifically states what to do with Wildcards and where they are or are not allowed, however "...other FQDNs that end with all the Domain Labels of the validated FQDN," have not been specifically addressed.  I suggest that either 'other FQDNs' be explicitly lumped in with Wildcard in the table, or other specific text and an additional column in the table added to specify when they are allowed.  See additional note on this subject below.

richardwsmith (commented):
Added some suggestions for adding a compliance date, and clarifying "...other FQDNs that end with all the Domain Labels of the validated FQDN."

docs/BR.md
jsha: The "prune" column is equivalent to "other FQDNs." The old BR language specified the behavior twice: once in ADN selection and again in these Note: sections. I've clarified that the operating principle is ADN selection.

Say a subscriber wishes to issue for `www.example.com`, and the CA has performed a validation for `example.com` via `3.2.2.4.7 DNS Change`, and that validation is still valid. During ADN selection, the CA prunes one label (`www`) and uses `example.com` as the ADN, relying on the existing validation.

docs/BR.md
```diff
@@ -1020,34 +1020,25 @@ If the CA follows redirects, the following apply:
...
-
 ##### 3.2.2.4.22 DNS TXT Record with Persistent Value
 
-Confirming the Applicant's control over a FQDN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated (i.e., "`_validation-persist.[Authorization Domain Name]`"). For this method, the CA MUST NOT use the FQDN returned from a DNS CNAME lookup as the FQDN for the purposes of domain validation. This prohibition overrides the Authorization Domain Name definition. CNAME records MAY be followed when resolving the Persistent DCV TXT Record.
-
-The CA MUST confirm the Persistent DCV TXT Record’s RDATA value fulfills the following requirements:
+Confirming the Applicant's control over the ADN by verifying the presence of a Persistent DCV TXT Record identifying the Applicant. The record MUST be placed at the "`_validation-persist`" label prepended to the Authorization Domain Name being validated (i.e., "`_validation-persist.[Authorization Domain Name]`"). This prohibition overrides the Authorization Domain Name definition.
```
aarongable: The sentence "This prohibition..." doesn't flow right now that the preceding MUST NOT has been removed.

hablutzel1:
Could someone be so kind as to share with me some references that provide the detailed rationale that introduced "The CA may use the FQDN returned from a DNS CNAME lookup as the FQDN for the purposes of domain validation." in Ballot 169?

I'm especially interested in the exact use cases discussed back then that motivated this language.

docs/BR.md
aarongable: Thanks, I've incorporated this change, but also removed the comment about following CNAMEs during TXT resolution, because that's Just How DNS Works (tm).

docs/BR.md
aarongable: I've double-checked to ensure that all of these "Note" sections have been removed. There are notes in other parts of the BRs, but none remain in Section 3.2.2.4's subsections.

Addressing "...other FQDNs that end with all the Domain Labels of the validated FQDN" is already handled by the "Prune" column of the table in 3.2.2.4.

docs/BR.md
aarongable: The "Prune" column of the table covers "other FQDNs that end with all the Domain Labels of the validated FQDN". If a validation method allows pruning during ADN construction, then a validation for a given ADN can be used for other FQDNs that also prune to that same ADN. These are the same thing.

docs/BR.md
aarongable: I've added an effective date of 2026-09-15. This should be plenty of time for CAs to comply, since the intent of the ballot is to codify and clarify existing acceptable behavior, not massively change allowed behavior. We can discuss how the effective date is implemented (I've used the same approach as the NCSSRs did, since this change touches so many subsections) in the Validation call tomorrow morning.

docs/BR.md
```diff
@@ -731,16 +730,44 @@ The CA SHOULD implement a process to screen proxy servers in order to prevent re
 
 #### 3.2.2.4 Validation of Domain Authorization or Control
 
-This section defines the permitted processes and procedures for validating the Applicant's ownership or control of the domain.
+Prior to 2026-09-15, the CA SHALL adhere to Section 3.2.2.4 (and its subsections) of these requirements or Section 3.2.2.4 of Version 2.2.7 of the Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates. Effective 2026-09-15, the CA SHALL adhere to Section 3.2.2.4 of these requirements.
```
CBonnell:
```suggestion
Prior to 2026-09-15, the CA SHALL adhere to Section 3.2.2.4 (and its subsections) of these Requirements or Section 3.2.2.4 of Version 2.2.7 of the Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates. Effective 2026-09-15, the CA SHALL adhere to Section 3.2.2.4 of these Requirements.
```

docs/BR.md
```diff
@@ -1083,7 +1075,7 @@ Effective March 15, 2027:
 
 ##### 3.2.2.5.3 Reverse Address Lookup
 
-Confirming the Applicant's control over the IP Address by obtaining a Domain Name associated with the IP Address through a reverse-IP lookup on the IP Address and then verifying control over the FQDN using a method permitted under [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control).
+Confirming the Applicant's control over the IP Address by obtaining an FQDN associated with the IP Address through a reverse-IP lookup on the IP Address and then using that FQDN as an ADN and performing validation using a method permitted under [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). The ADN for this method MUST be exactly the FQDN returned from the reverse-IP lookup. The ADN selection algorithm in section 3.2.2.4 does not apply.
```
e3n0: > The ADN for this method MUST be exactly the FQDN returned from the reverse-IP lookup.

This could use some clarification. Is this an additional check to be applied to the reverse IP response? If so, how exactly should a CA perform this check?

aarongable: It's not a check: the ADN is something that the CA _chooses_, and this is constraining the CA's choice of ADN.

docs/BR.md
```diff
@@ -775,22 +799,18 @@ This method has been retired and MUST NOT be used. Prior validations using this
 
 ##### 3.2.2.4.4 Email to a Constructed Address
 
-Confirm the Applicant's control over the FQDN by:
+Confirm the Applicant's control over the ADN by:
 
-1. Sending an email to one or more addresses created by using 'admin', 'administrator', 'webmaster', 'hostmaster', or 'postmaster' as the local part, followed by the at-sign ("@"), followed by an Authorization Domain Name; and
+1. Sending an email to one or more addresses created by using 'admin', 'administrator', 'webmaster', 'hostmaster', or 'postmaster' as the local part, followed by the at-sign ("@"), followed by the Authorization Domain Name; and
```
ChristopherRC: Nit: unless there is a reason why we're not using the acronym? 

```suggestion
1. Sending an email to one or more addresses created by using 'admin', 'administrator', 'webmaster', 'hostmaster', or 'postmaster' as the local part, followed by the at-sign ("@"), followed by the ADN; and
```

docs/BR.md
```diff
@@ -807,10 +827,10 @@ This method has been retired and MUST NOT be used. Prior validations using this
 
 ##### 3.2.2.4.7 DNS Change
 
-Confirming the Applicant's control over the FQDN by confirming the presence of a Random Value or Request Token in a DNS CNAME, TXT or CAA record for either:
+Confirming the Applicant's control over the ADN by confirming the presence of a Random Value or Request Token in a DNS CNAME, TXT or CAA record returned in a query for either:
 
-1. an Authorization Domain Name; or
-2. an Authorization Domain Name that is prefixed with a Domain Label that begins with an underscore character.
+1. the Authorization Domain Name; or
```
ChristopherRC: Nit

```suggestion
1. the ADN; or
```

docs/BR.md
```diff
@@ -807,10 +827,10 @@ This method has been retired and MUST NOT be used. Prior validations using this
...
+2. the Authorization Domain Name prefixed with a Domain Label that begins with an underscore character.
```
ChristopherRC: Nit

```suggestion
2. the ADN prefixed with a Domain Label that begins with an underscore character.
```

docs/BR.md
```diff
@@ -880,16 +888,14 @@ Effective March 15, 2028:
 
 ##### 3.2.2.4.14 Email to DNS TXT Contact
 
-Confirming the Applicant's control over the FQDN by sending a Random Value via email and then receiving a confirming response utilizing the Random Value. The Random Value MUST be sent to a DNS TXT Record Email Contact for the Authorization Domain Name selected to validate the FQDN.
+Confirming the Applicant's control over the ADN by sending a Random Value via email and then receiving a confirming response utilizing the Random Value. The Random Value MUST be sent to a DNS TXT Record Email Contact for the Authorization Domain Name.
```
ChristopherRC: Nit

```suggestion
Confirming the Applicant's control over the ADN by sending a Random Value via email and then receiving a confirming response utilizing the Random Value. The Random Value MUST be sent to a DNS TXT Record Email Contact for the ADN.
```

docs/BR.md
```diff
@@ -4039,21 +4031,17 @@ The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain
 
 This appendix defines permissible verification procedures for including one or more Onion Domain Names in a Certificate.
 
-1. The Domain Name MUST contain at least two Domain Labels, where the rightmost Domain Label is "onion", and the Domain Label immediately preceding the rightmost "onion" Domain Label is a valid Version 3 Onion Address, as defined in Section 6 of the Tor Rendezvous Specification - Version 3 located at <https://spec.torproject.org/rend-spec-v3>.
-
-2. The CA MUST verify the Applicant's control over the Onion Domain Name using at least one of the methods listed below:
+1. The Authorization Domain Name MUST contain at least two Domain Labels, where the rightmost Domain Label is "onion", and the Domain Label immediately preceding the rightmost "onion" Domain Label is a valid Version 3 Onion Address, as defined in Section 6 of the Tor Rendezvous Specification - Version 3 located at <https://spec.torproject.org/rend-spec-v3>.
```
ChristopherRC: Nit

```suggestion
1. The ADN MUST contain at least two Domain Labels, where the rightmost Domain Label is "onion", and the Domain Label immediately preceding the rightmost "onion" Domain Label is a valid Version 3 Onion Address, as defined in Section 6 of the Tor Rendezvous Specification - Version 3 located at <https://spec.torproject.org/rend-spec-v3>.
```

docs/BR.md
```diff
@@ -4039,21 +4031,17 @@ The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain
...
 
-   a. The CA MAY verify the Applicant's control over the .onion service by using one of the following methods from [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control):
+2. The CA MUST verify the Applicant's control over the Authorization Domain Name using at least one of the methods listed below:
```
ChristopherRC: Nit

```suggestion
2. The CA MUST verify the Applicant's control over the ADN using at least one of the methods listed below:
```

docs/BR.md
```diff
@@ -4039,21 +4031,17 @@ The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain
...
 
-      i. [Section 3.2.2.4.18](#322418-agreed-upon-change-to-website-v2) - Agreed-Upon Change to Website v2
-      ii. [Section 3.2.2.4.19](#322419-agreed-upon-change-to-website---acme) - Agreed-Upon Change to Website - ACME
-      iii. [Section 3.2.2.4.20](#322420-tls-using-alpn) - TLS Using ALPN
+   a. The CA MAY verify the Applicant's control over the ADN by using any method from [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control) that says "This method allows Onion Domain Name issuance", with this modification:
 
-      When these methods are used to verify the Applicant's control over the .onion service, the CA MUST use Tor protocol to establish a connection to the .onion hidden service. The CA MUST NOT delegate or rely on a third-party to establish the connection, such as by using Tor2Web.
+      When these methods are used to verify the Applicant's control over an Onion Domain Name, the CA MUST use Tor protocol to establish a connection to the Authorization Domain Name. The CA MUST NOT delegate or rely on a third-party to establish the connection, such as by using Tor2Web.
```
ChristopherRC: Nit

```suggestion
      When these methods are used to verify the Applicant's control over an Onion Domain Name, the CA MUST use Tor protocol to establish a connection to the ADN. The CA MUST NOT delegate or rely on a third-party to establish the connection, such as by using Tor2Web.
```

docs/BR.md
```diff
@@ -4039,21 +4031,17 @@ The DNS TXT record MUST be placed on the "`_validation-contactphone`" subdomain
...
 
-      **Note**: This section does not override or supersede any provisions specified within the respective methods. The CA MUST only use a method if it is still permitted within that section and MUST NOT issue Wildcard Certificates or use it as an Authorization Domain Name, except as specified by that method.
+      **Note**: This section does not override or supersede any provisions specified within the respective methods. The CA MUST only use a method if it is still permitted within that section.
 
-   b. The CA MAY verify the Applicant's control over the .onion service by having the Applicant provide a Certificate Request signed using the .onion service's private key if the Attributes section of the certificationRequestInfo contains:
+   b. The CA MAY verify the Applicant's control over the .onion service corresponding to the Authorization Domain Name by having the Applicant provide a Certificate Request signed using the .onion service's private key if the Attributes section of the certificationRequestInfo contains:
```
ChristopherRC: Nit

```suggestion
   b. The CA MAY verify the Applicant's control over the .onion service corresponding to the ADN by having the Applicant provide a Certificate Request signed using the .onion service's private key if the Attributes section of the certificationRequestInfo contains:
```

ChristopherRC (commented):
Several nits, unless there is a reason why we're not using the ADN acronym in those places?

docs/BR.md
```diff
@@ -1081,7 +1073,7 @@ Effective March 15, 2027:
 
 ##### 3.2.2.5.3 Reverse Address Lookup
 
-Confirming the Applicant's control over the IP Address by obtaining a Domain Name associated with the IP Address through a reverse-IP lookup on the IP Address and then verifying control over the FQDN using a method permitted under [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control).
+Confirming the Applicant's control over the IP Address by obtaining an FQDN associated with the IP Address through a reverse-IP lookup on the IP Address and then using that FQDN as an ADN and performing validation using a method permitted under [Section 3.2.2.4](#3224-validation-of-domain-authorization-or-control). The ADN for this method MUST be exactly the FQDN returned from the reverse-IP lookup. The ADN selection algorithm in section 3.2.2.4 does not apply.
```
e3n0: Is it intentional that the ballot effective date only applies to section 3.2.2.4? Since this change requires action by CAs to comply, section 3.2.2.5 should also have an effective date - ideally November 2026 at earliest since migrations will be required.

Since this validation method is being banned in March 2027 anyway, it would be helpful to continue to allow trimming labels until the method is banned; if we require CAs to make the change in the meantime, it could distract from the migration to other methods, and would only have effect for ~4 months with a November 2026 effective date.

e3n0: This change to Reverse Address Lookup specifically isn't justified in the PR description or the preamble. The table in the PR description doesn't contain methods from section 3.2.2.5.

aarongable: This is a rephrasing and clarification, not a change.

The current definition of Authorization Domain Name says that it is "The FQDN used to obtain authorization **for a given FQDN to be included in a Certificate**" (emphasis added). There is no concept of an ADN for IP addresses. If a CA is doing a reverse address lookup on an IP address to get an FQDN, then converting that FQDN into an ADN by pruning labels, they are already violating the BRs.

Therefore there is no need for an effective date on this piece of the ballot.

aarongable: This method is "do a reverse address lookup, then do a method from 3.2.2.4 on the result". There's no need for the table to include any information about this method, because it already includes all of the relevant information about the methods to which this method could delegate.

See the comment above for why this isn't actually a change.

e3n0: While it is true that the literal definition of an ADN specifies FQDNs 'to be included in a Certificate', Method 3.2.2.5.3 explicitly directs CAs to “verify control over the FQDN using a method permitted under Section 3.2.2.4.” and methods like 3.2.2.4.7 require an ADN to operate. Interpreting the current BRs to forbid the concept of ADN on the FQDN obtained via reverse IP lookup would break the intersection of these two rules, rendering 3.2.2.4.7 unusable with 3.2.2.5.3.


Method 3.2.2.4.7 says:

> Confirming the Applicant's control over the **FQDN** by confirming the presence of a Random Value or Request Token in a DNS CNAME, TXT or CAA record for either:
> 
> 1. **an Authorization Domain Name**; or
> 2. **an Authorization Domain Name** that is prefixed with a Domain Label that begins with an underscore character.

Not 

> Confirming the Applicant's control over the **FQDN** by confirming the presence of a Random Value or Request Token in a DNS CNAME, TXT or CAA record for either:
> 
> 1. **The FQDN**; or
> 2. **The FQDN** that is prefixed with a Domain Label that begins with an underscore character.

dzacharo (approved):
Ballot 101v2 passed and cleared IPR Review period

PR #656, opened 2026-03-10
Ballot SC-099: Improve Recording of Validation Methods

aarongable:
The current BRs contain the following text in Sections 3.2.2.4 and 3.2.2.5:

> CAs SHALL maintain a record of which [domain/IP] validation method, including relevant BR version number, they used to validate every [domain/IP Address].

This text is problematic for four reasons:
- it only applies to domains and IP addresses, not other things being validated like organization information;
- the requirement to "maintain a record" is unclear (in audit logs? in a database? in a document?);
- no one is sure what the "relevant BR version" is (the effective version when the validation happened? the version that the validation process believes it implements?); and
- the text is not in Section 5.4.1, which is where the BRs concentrate requirements about recording information.

To resolve these issues, we need to start from first principles. The goal, as evidenced by discussion when this requirement was introduced and recollections of CA/BF members who were participating at the time, is to ensure that CAs and auditors are able to definitively identify the validation process the with which the CA was required to comply for any given validation.

To determine what rules governed any given validation, we need two pieces of information:
1. When they performed that validation. The current requirements in Section 5.4.1 already require the CA to audit log the time of every verification procedure.
2. What validation method the CA used. This ballot augments Section 5.4.1 to also require that the CA audit log the validation method used, as well as other useful information.

Because we can accomplish the goal with a small addition to Section 5.4.1, this ballot removes the current text from Sections 3.2.2.4 and 3.2.2.5.

Note that this ballot removes the requirement to "record" the "relevant BR version number". This is not considered a loss, for several reasons:
- No CA or auditor should be relying on a logged BR version number to determine which version of the BRs governed the validation. The logged version number is not enforceable; the only information which can be relied upon to map a validation to a BRs version is the time at which the validation was performed.
- If a new version of the BRs adds requirements to a validation method, but gates those new requirements behind a future effective date, the logged version number may indicate that the CA complies with that new version long before the CA has actually implemented the soon-to-be-required behavior.
- If a new version of the BRs adds requirements to a validation method immediately upon the new version becoming effective, then a CA must update their processes to comply before the new version of the BRs is even published, and therefore the logged version number will be out-of-date compared to the implemented behavior.
- Part of the original motivation behind these statements was to encourage CAs to actually read new versions of the BRs and ensure their processes comply with them. We now have other mechanisms to encourage this behavior, such as CCADB self-assessments.

Therefore we conclude that recording the relevant BRs version number is neither useful nor well-specified, and therefore should not be included in the BRs.

This issue was discussed [on Mozilla dev-security-policy@](https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/x4zRcboSDwU) as well as at the CA/BF Face-to-Face Meeting 67 in Houston on March 10, 2026. The conclusion of those discussions was that we should create this ballot.

This ballot is written by Aaron Gable (ISRG / Let's Encrypt) and endorsed by Gurleen Grewal (Google Trust Services), Trev Ponds-White (Amazon Trust Services), and Roman Fischer (SwissSign).

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -230,6 +230,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2026-03-15     | 4.2.1                     | Domain Name and IP Address validation maximum data reuse period is 200 days.     |
 | 2026-03-15     | 6.3.2                     | Maximum validity period of Subscriber Certificates is 200 days.   |
 | 2026-03-15     | 7.1.2.4                   | CAs MUST NOT use Precertificate Signing CAs to issue Precertificates. CAs MUST NOT issue certificates using the Technically Constrained Precertificate Signing CA Certificate Profile specified in Section 7.1.2.4.    |
+| 2026-07-15     | 5.4.1                     | Audit logged validation records MUST include specific information.    |
 | 2026-09-15     | 7.1.3.2.1                 | Sunset all remaining use of SHA-1 in Certificates and CRLs.   |
 | 2027-03-15     | 3.2.2.4 and 3.2.2.5       | CAs MUST NOT rely on Methods 3.2.2.4.16, 3.2.2.4.17, 3.2.2.5.2, and 3.2.2.5.5 to issue Subscriber Certificates.    |
 | 2027-03-15     | 3.2.2.5.3                 | CAs MUST NOT rely on Method 3.2.2.5.3 to issue Subscriber Certificates.    |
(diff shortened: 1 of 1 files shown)


docs/BR.md
```diff
@@ -1864,7 +1860,10 @@ The CA SHALL record at least the following events:
 
 2. Subscriber Certificate lifecycle management events, including:
    1. Certificate requests, renewal, and re-key requests, and revocation;
-   2. All verification activities stipulated in these Requirements and the CA's Certification Practice Statement;
+   2. All verification activities stipulated in these Requirements and the CA's Certification Practice Statement, minimally recording the following information:
```
aww-aww:
```suggestion
   2. All verification activities stipulated in these Requirements and the CA's Certification Practice Statement. Records must include at a minimum:
```

docs/BR.md
```diff
@@ -1864,7 +1860,10 @@ The CA SHALL record at least the following events:
...
+      1. the information being validated (e.g., the applied-for FQDN or the organization name);
```
aww-aww, line 1835: Nit: capitalize The at the start of each of these.

aarongable: I considered it, but decided to match the style of the sub-bullets for MPIC, just a few lines below this diff.

docs/BR.md
aarongable: This phrasing matches what was used for MPIC below. I'm happy to do something else if there's an argument for it being truly better, but otherwise I'd like to match local style.

aarongable: Update: Trev pointed out that this should have an effective date, so I've adopted your suggested phrasing in order to elegantly incorporate that.

dzacharo (approved):

PR #662, opened 2026-04-22
SC-XXX: Permit ML-DSA public keys and signatures in certificates

e3n0:
Currently the BRs require the use of RSA or ECDSA keys and signatures. This ballot allows ML-DSA-{44,65,87} as well.

Based on [SMC013](https://github.com/cabforum/smime/pull/284) by @srdavidson.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -3412,6 +3412,8 @@ The CA SHALL indicate an ML-DSA key using one of the following algorithm identif
 - ML-DSA-65 (OID: 2.16.840.1.101.3.4.3.18), or
 - ML-DSA-87 (OID: 2.16.840.1.101.3.4.3.19).
 
+As described in [Section 6.1.5](#615-key-sizes), ML-DSA CA Key Pairs MUST use the ML-DSA-87 parameter set.
+
 The parameters for ML-DSA keys SHALL be absent. The CA MUST NOT use HashML-DSA; only "pure" ML-DSA is permitted.
 
 When encoded, the AlgorithmIdentifier for ML-DSA keys SHALL be byte-for-byte identical with the following hex-encoded bytes:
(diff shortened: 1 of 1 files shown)


dzacharo:
@e3n0 do you represent a CABF Member?

dzacharo:
> @e3n0 do you represent a CABF Member?

Clarified that this is a proposal from Google Trust Services.

docs/BR.md
```diff
@@ -1,11 +1,11 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 
-subtitle: Version 2.2.6
+subtitle: Version 2.2.x
 author:
   - CA/Browser Forum
 
-date: 31-March-2026
+date: XX-XX-2026
```
aarongable: note: Don't include these updates in the draft PR, because they'll cause difficulties with the merge if any other ballots land before this one. The chairs will handle this part if/when the ballot passes.

docs/BR.md
```diff
@@ -161,6 +161,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.2.4 | SC096 | Carve-out for DNSSEC verification logging requirements                                  | 2026-01-14 | 2026-02-17 |
 | 2.2.5 | SC097 | Sunset all remaining use of SHA-1 signatures in Certificates and CRLs                   | 2026-02-24 | 2026-02-25 |
 | 2.2.6 | SC095 | Clean-up 2025                                                                           | 2026-02-27 | 2026-03-31 |
+| 2.2.X | SCXXX | Permit ML-DSA public keys and signatures in Certificates                                | 2026-XX-XX | 2026-XX-XX |
```
aarongable: Same here -- the chairs will handle adding this line.

docs/BR.md
```diff
@@ -2061,6 +2062,14 @@ For ECDSA key pairs, the CA SHALL:
 
 - Ensure that the key represents a valid point on the NIST P-256, NIST P-384 or NIST P-521 elliptic curve.
 
+For ML-DSA key pairs, the CA SHALL:
+
+- If the key pair is a CA Key Pair, ensure that the key uses the ML-DSA-87 (OID: 2.16.840.1.101.3.4.3.19) parameter set.
```
aarongable: Why only allow ML-DSA-87 for CA keys?

For RSA CAs we allow 2048-bit keys, which NIST considers to have fewer bits of security than ML-DSA-44. For ECDSA CAs we allow P-256 keys, which NIST similarly considers to have fewer bits of security than ML-DSA-44.

Requiring ML-DSA-87 when we could be using ML-DSA-44 doubles both pubkey and signature sizes, a large cost we should not force upon the webpki for security theater.

(Yes, CNSA 2.0 requires ML-DSA-87. We don't have to and shouldn't follow their lead; they're oversizing everything for the sake of simplicity. CAs that want to comply with those requirements can limit themselves in their CPS; no need for the BRs to do it too.)

e3n0: Updated to permit the ML-DSA-44 and ML-DSA-65 parameter sets for CA Keys as well.

ghen2:
That last commit says "only allow pure PQ chains", but it only prohibits "ML-DSA-to-other", not "other-to-ML-DSA", correct?

e3n0:
This PR is being superseded. The current review copy for the new canonical ballot can be found at https://github.com/slghtr-says/servercert/pull/8.

PR #661, opened 2026-04-16
SC-102: EVG Domain Ownership Validation Reuse

dustinhollenback-apple:
The EV Guidelines currently require CAs to verify WHOIS records when revalidating domain names for existing subscribers (Section 3.2.2.14.1). With the TLS BRs sunsetting WHOIS-based validation methods (SC-080) and the broader industry move away from reliance on WHOIS data, this requirement needs to be updated. This still provides a higher bar for data re-use than is present in the Baseline Requirements.

Additionally, the EVGs hardcode specific values for certificate validity periods (Section 6.3.2) and domain name validation data reuse periods (Section 3.2.2.14.3). With the TLS Baseline Requirements now containing a schedule of reducing validity periods and data reuse periods (introduced by SC-081), these hardcoded values risk becoming stale or giving the misleading impression that EV certificates are subject to less restrictive limits than the TLS BRs.

This ballot makes three changes:

1. Replaces the WHOIS-based domain revalidation requirement in Section 3.2.2.14.1 with a prioritized set of methods: first attempting verification via an authenticated channel with the domain registrar or registry, then falling back to a recent domain control validation, and finally performing a new domain control validation.

2. Replaces the hardcoded "398 days" domain name reuse period in Section 3.2.2.14.3 with a reference to Section 4.2.1 of the Baseline Requirements, so the EVG automatically follows the planned reductions without requiring further EVG amendments.

3. Removes the EV-specific validity period language in Section 6.3.2, allowing the TLS BR requirements to govern directly.

PR #669, opened 2026-06-11
SC-102 alternative: align EV domain validation reuse and validity with the Baseline Requirements

dustinhollenback-apple:
## Summary

This is an alternative to #661. It lets the Baseline Requirements govern EV domain re-validation rather than adding the EV-specific same-entity check that #661 proposes. It also removes hardcoded numbers in the EVG that are now wrong.

The EV Guidelines today:

- require CAs to re-check WHOIS or RDAP registration data when revalidating domain names for existing subscribers (Section 3.2.2.14.1);
- set the Domain Name data reuse period at "398 days" (Section 3.2.2.14.3); and
- cap EV certificate validity at 398 days and recommend a twelve-month maximum (Section 6.3.2).

WHOIS-based validation is being retired (SC-080), and SC-081 put certificate validity and data reuse periods on a schedule that shrinks them over time. The two "398 day" figures are already wrong: the Baseline Requirements cap both validity and Domain Name reuse at 200 days today, and lower on the published schedule. An EV certificate is a TLS Subscriber Certificate, so it is already bound by those limits, and the separate EVG numbers only invite confusion.

## Changes

This ballot makes four changes:

1. Section 3.2.2.14.1(6): remove the WHOIS/RDAP same-registrant test. The Applicant's right to use the Domain Name is re-verified under Section 3.2.2.7 (which uses BR Section 3.2.2.4), within the data reuse period set in Section 3.2.2.14.3(1)(F).
2. Section 3.2.2.14.3(1)(F): replace the hardcoded "398 days" Domain Name reuse period with a reference to Section 4.2.1 of the Baseline Requirements, so the EVG follows the scheduled reductions without further amendment.
3. Section 3.2.2.14.3(2): correct the "398-day period" sentence, which stops being accurate once item (F) points to the Baseline Requirements.
4. Section 6.3.2: replace the EV validity text with a reference to Section 6.3.2 of the Baseline Requirements.

Items (A) through (E) and (G) in Section 3.2.2.14.3(1) stay at 398 days, which matches the Baseline Requirements Subject Identity Information reuse period.

### DIFF
--- a/docs/EVG.md
+++ b/docs/EVG.md
@@ -935,7 +935,7 @@ If an Applicant has a currently valid EV Certificate issued by the CA, a CA MAY 
 3. The Applicant's Verified Method of Communication required by [Section 3.2.2.5](#3225-verified-method-of-communication) but still MUST perform the verification required by [Section 3.2.2.5.2](#32252-acceptable-methods-of-verification) (B);
 4. The Applicant's Operational Existence under [Section 3.2.2.6](#3226-verification-of-applicants-operational-existence);
 5. The Name, Title, Agency and Authority of the Contract Signer, and Certificate Approver, under [Section 3.2.2.8](#3228-verification-of-name-title-and-authority-of-contract-signer-and-certificate-approver); and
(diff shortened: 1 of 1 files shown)


romanf:
Very clean! Like it a lot.

srdavidson:
@dustinhollenback-apple is this still required now that the alt ballot passed?

dustinhollenback-apple:
No longer needed. Closed.

dustinhollenback-apple:
Re-opening to allow merge.

dzacharo (approved):

PR #678, opened 2026-08-19
SC105: Remove High Risk language from TLS BR and EVG 

srdavidson:


docs/EVG.md
```diff
@@ -880,8 +880,6 @@ A Qualified Government Tax Information Source is a Qualified Government Informat
 
 ##### 3.2.2.12.1 High Risk Status
 
-The High Risk Certificate requirements of Section 4.2.1 of the Baseline Requirements apply equally to EV Certificates.
```
romanf, line 883: I propose to remove this chapter and re-number the following chapters.
There seems to be no reference to this or the following chapters in any other CA/B document, so re-numbering would not have side-effects.

I if removal is not desirable, then I would put "No stipulation." as it's done in the TLS BR (e.g. 1.4.2).

PR #680, opened 2026-09-03
SC-1XX EVG Cleanup and BR Sync

richardwsmith:
Redline for ballot SC-1XX (EVG Cleanup and BR Sync). The Guidelines for the Issuance and Management of Extended Validation Certificates (EV Guidelines) incorporate the Baseline Requirements by reference, so this ballot removes EV Guidelines text that duplicates, contradicts, or has fallen behind the Baseline Requirements, and migrates the Extended Validation (EV) certificate profile from EV Guidelines Section 7 into Baseline Requirements Section 7, including the subject attribute rules, the cabfOrganizationIdentifier extension, and Appendices G and H.  The EV Guidelines that remain are a validation-methods document.

### DIFF
--- a/docs/BR.md
+++ b/docs/BR.md
@@ -1,11 +1,11 @@
 ---
 title: Baseline Requirements for the Issuance and Management of Publicly-Trusted TLS Server Certificates
 
-subtitle: Version 2.2.9
+subtitle: Version 2.2.X
 author:
   - CA/Browser Forum
 
-date: 6-Aug-2026
+date: TBD
 
 copyright: |
   Copyright 2026 CA/Browser Forum
@@ -164,6 +164,7 @@ The following Certificate Policy identifiers are reserved for use by CAs to asse
 | 2.2.7 | SC099 | Improve Recording of Validation Method                                                  | 2026-04-18 | 2026-05-19 |
 | 2.2.8 | SC098 | Process RFC 8657 CAA Parameters                                                         | 2026-05-13 | 2026-06-16 |
 | 2.2.9 | SC101 | Clarify Authorization Domain Names                                                      | 2026-07-02 | 2026-08-06 |
+| 2.2.X | SC1XX | EVG Cleanup and BR Sync                                                                 | TBD        | TBD        |
 
 \* Effective Date and Additionally Relevant Compliance Date(s)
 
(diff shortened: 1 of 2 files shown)


docs/EVG.md
```diff
@@ -324,30 +291,23 @@ Abbreviations and Acronyms are defined in the Baseline Requirements except as ot
...
-| IFAC | International Federation of Accountants |
 | IRS | Internal Revenue Service |
 | ISP | Internet Service Provider |
 | QGIS | Qualified Government Information Source |
 | QTIS | Qualified Government Tax Information Source |
 | QIIS | Qualified Independent Information Source |
 | SEC | (US Government) Securities and Exchange Commission |
-| UTC(k) | National realization of Coordinated Universal Time |
 
 ### 1.6.3 References
```
romanf, line 310: I suggest to add "No stipulation." instead of leaving the section blank.

richardwsmith: Thanks, Roman. I don't disagree but it was decided years ago, for reasons I no longer recall, that CA/B Forum documents would simply leave blank sections blank. If we decide we want to change that I have no objection but would prefer that it be done by a singular "Add 'No stipulation'," ballot.

romanf: Sorry, I wasn't aware of that decision. I just read the TLS BR 1.1 where it says: "However, rather than beginning with a “no stipulation” comment in all empty sections, the CA/Browser Forum is leaving such sections initially blank until a decision of “no stipulation” is made." and thought that this would be applicable here since we explicitly removed the refernces... But I'm in no way insisting. 😉

docs/BR.md
```diff
@@ -311,6 +313,8 @@ The Definitions found in the CA/Browser Forum's Network and Certificate System S
 
 **CA Key Pair**: A Key Pair where the Public Key appears as the Subject Public Key Info in one or more Root CA Certificate(s) and/or Subordinate CA Certificate(s).
 
+**Canonical Date Representation**: A date that is formatted as YYYY-MM-DD, where "YYYY" is the four-digit year, "MM" is the two-digit month, and "DD" is the two-digit day of the month. Each element of the date is separated with a single hyphen-minus "-" (0x2D (ASCII), U+002D (UTF-8)). Each element is padded with leading zeroes as needed to ensure that year values consist of four digits and month and day of the month values consist of two digits. Example dates in this representation: "0748-04-02", "2024-10-14".
```
aww-aww, line 316: Is this date format being used instead of ISO 8601 for simplicity? 8601 is pretty universal, but it has some subtle complexity that may not be warranted here.

